PR #37, opened 2023-08-02
Fix the swap logic

Adamantios:
Fixes a comparison that was causing the depositing for wrapped xDAI to fail.
Moreover, adds some useful logging info.


packages/valory/skills/decision_maker_abci/behaviours/bet_placement.py
```diff
@@ -341,7 +363,7 @@ def async_act(self) -> Generator:
             tx_submitter = betting_tx_hex = None
 
             can_exchange = (
-                self.collateral_token == WXDAI
+                self.is_wxdai
```
Adamantios, line 366: The addresses provided by the subgraph are not checksummed. As a result, `can_exchange` was always `False`.

angrybayblade (approved):

jmoreira-valory (approved):

PR #45, opened 2023-08-08
Implement a script to get trades

Adamantios:
Implement a script to get trades from the OMEN subgraph for a specific trade creator.

Usage: `python trades.py address`

Where the `address` can be either checksummed or not. As a matter of fact, the `address` is always converted to lower as this is how the `OMEN` subgraph expects it.

### DIFF
--- a/scripts/trades.py
+++ b/scripts/trades.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2022-2023 Valory AG
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


scripts/trades.py
```diff
@@ -0,0 +1,85 @@
...
+def parse_response() -> str:
+    """Parse the trades from the response."""
+    serialized_res = res.json()
+    trades = serialized_res.get("data", {}).get("fpmmTrades", [])
+    return json.dumps(trades, indent=4)
+
+
+if __name__ == "__main__":
+    creator = parse_arg()
+    res = query_subgraph()
```
dvilelaf, line 83: I think you need to pass creator here

scripts/trades.py
```diff
@@ -0,0 +1,85 @@
...
+    parsed = parse_response()
```
dvilelaf, line 84: And res here

scripts/trades.py
Adamantios: The `creator` is globally defined so no need to pass it around.

dvilelaf: Oh, I see

dvilelaf (approved):

0xArdi (approved):

PR #46, opened 2023-08-09
Stop previous containers

Adamantios:
Check if `docker` is installed and stop containers from previous runs, if still running.

### DIFF
--- a/scripts/run_service.sh
+++ b/scripts/run_service.sh
@@ -13,6 +13,13 @@ command -v poetry >/dev/null 2>&1 ||
   exit 1
 }
 
+command -v docker >/dev/null 2>&1 ||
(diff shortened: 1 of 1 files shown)


dagacha:
@Adamantios This is the ouptut that I see

```
./run_service.sh 
abci0
node0
Error response from daemon: No such container: trader_abci_0
Error response from daemon: No such container: trader_tm_0
Enter agent address:
```
I think we should avoid displaying an error message before the user has even had a chance to enter their inputs. Alternatives:
- Displaying `stopping abci0, node0`messages
- Displaying nothing

dagacha:
Changes LGTM @Adamantios (even if I'm listed as reviewer I cannot see the reviewing links open for me)

dagacha (approved):
LGTM

PR #52, opened 2023-08-16
Fix FSM rewiring RefillRequiredRound

jmoreira-valory:
Fix FSM rewiring RefillRequiredRound

### DIFF
--- a/packages/valory/skills/trader_abci/composition.py
+++ b/packages/valory/skills/trader_abci/composition.py
@@ -33,6 +33,7 @@ from packages.valory.skills.decision_maker_abci.states.decision_receive import (
 from packages.valory.skills.decision_maker_abci.states.final_states import (
     FinishedDecisionMakerRound,
     FinishedWithoutDecisionRound,
+    RefillRequiredRound,
 )
 from packages.valory.skills.decision_maker_abci.states.sampling import SamplingRound
 from packages.valory.skills.market_manager_abci.rounds import (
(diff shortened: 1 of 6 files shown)


packages/valory/skills/trader_abci/composition.py
```diff
@@ -76,6 +77,7 @@
     FinishedMarketManagerRound: SamplingRound,
     FailedMarketManagerRound: ResetAndPauseRound,
     FinishedDecisionMakerRound: RandomnessTransactionSubmissionRound,
+    RefillRequiredRound: ResetAndPauseRound,
```
angrybayblade, line 80: Before this change where were we transitioning to after the `RefillRequiredRound` round?

0xArdi: nowhere, it was missing

0xArdi (approved):

angrybayblade (approved):

DavidMinarsch (approved):

DavidMinarsch (commented):
Ok so I don't know if this will work. RefillRequiredRound(DegenerateRound) - not sure the framework allows moving through these rounds

DavidMinarsch (commented):
Ah nm - this is just chaining and therefore replaced

0xArdi:
@DavidMinarsch its only an issue if not wired properly

Adamantios (commented):
Why did we not delete the `RefillRequiredRound`? It has no purpose now.

PR #55, opened 2023-08-17
Rewired FSM to wait until Mech Deliver received

jmoreira-valory:
Rewired FSM to wait until Mech Deliver received

### DIFF
--- a/packages/valory/skills/decision_maker_abci/rounds.py
+++ b/packages/valory/skills/decision_maker_abci/rounds.py
@@ -76,7 +76,7 @@ class DecisionMakerAbciApp(AbciApp[Event]):
             - none: 8.
         2. DecisionReceiveRound
             - done: 4.
-            - mech response error: 3.
+            - mech response error: 2.
             - no majority: 2.
             - tie: 3.
             - unprofitable: 3.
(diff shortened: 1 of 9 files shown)


angrybayblade (approved):

DavidMinarsch (approved):

packages/valory/skills/decision_maker_abci/rounds.py
```diff
@@ -127,7 +127,7 @@ class DecisionMakerAbciApp(AbciApp[Event]):
         },
         DecisionReceiveRound: {
             Event.DONE: BetPlacementRound,
-            Event.MECH_RESPONSE_ERROR: BlacklistingRound,
+            Event.MECH_RESPONSE_ERROR: DecisionReceiveRound,  # loop on the same state until Mech deliver is received
```
Adamantios, line 130: What about the round timeout event?

jmoreira-valory: To my understanding, round timeout can be caused due to several other issues not only Mech not responding (for example, network latency, slow agent, etc.). What do you suggest to do in this cae?

DavidMinarsch: Isn't re-entering `DecisionReceiveRound` exactly what we are doing here?

Adamantios: Yes, what I suggest is that we do the same in case of a round timeout event.

PR #72, opened 2023-09-06
Make the strategy more sensible

Adamantios:
Makes the strategy more sensible and potentially more profitable.


packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -221,20 +223,81 @@ def _get_decision(
...
+                f"A non-positive bet threshold was given ({bet_threshold}). The threshold will be disabled, "
+                f"which means that any non-negative potential profit will be considered profitable!"
             )
-            return True
+            bet_threshold = 0
 
-        bet_amount = self.params.get_bet_amount(confidence)
-        fee = self.synchronized_data.sampled_bet.fee
-        return bet_amount - fee >= bet_threshold
+        potential_net_profit = num_shares - net_bet_amount - mech_price - bet_threshold
```
Adamantios, line 288: I believe that the mech's price should be removed from the profitability calculation as this has already been paid. 
If `0 < num_shares - net_bet_amount - bet_threshold < mech_price`, then the service would at least have a chance to cover a part of the mech's costs plus the threshold's amount.

0xArdi (approved):

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -221,20 +223,81 @@ def _get_decision(
 
         return self.mech_response.result.vote, self.mech_response.result.confidence
 
-    def _is_profitable(self, confidence: float) -> bool:
+    def _calc_binary_shares(self, net_bet_amount: int, vote: int) -> Tuple[int, int]:
+        """Calculate the claimed shares. This calculation only works for binary markets."""
```
Adamantios, line 227: The output of this method was tested with multiple values and the returned number of shares is accurate.

jmoreira-valory (approved):

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
DavidMinarsch: Ok! We need to add the mech cost to the outer decision making problem later

Adamantios: Removed on 30bf870d342dd3bb85727569179194dd5d3e2813 ans opened https://github.com/valory-xyz/trader/issues/75.

PR #73, opened 2023-09-06
Fix exception handling

Adamantios:
Fixes the exception handling for the RPC timeouts.


packages/valory/skills/decision_maker_abci/behaviours/reedem.py
```diff
@@ -267,6 +267,9 @@ def _conditional_tokens_interact(
 
     def _check_already_redeemed(self) -> WaitableConditionType:
         """Check whether we have already redeemed for this bet."""
+        if len(self.trades) == 0:
+            return True
```
Adamantios, line 271: Not related to the exception handling, but was causing unnecessary checks for redeeming on-chain, with very wide filtering, i.e., from the earliest to the latest block, which results in a guaranteed RPC timeout on Gnosis.

dvilelaf (approved):

PR #94, opened 2023-09-20
Implement tool selection strategy

Adamantios:


packages/valory/agents/trader/aea-config.yaml
```diff
@@ -187,6 +187,11 @@ models:
       redeeming_batch_size: ${int:5}
       slippage: ${float:0.01}
       redeem_margin_days: ${int:15}
+      epsilon: ${float:0.1}
+      irrelevant_tools: ${list:["openai-text-davinci-002", "openai-text-davinci-003",
```
DavidMinarsch, line 191: should the list not be empty to start with?

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -56,6 +57,7 @@
 # which is what we want in most cases
 # more info here: https://safe-docs.dev.gnosisdev.com/safe/docs/contracts_tx_execution/
 SAFE_GAS = 0
+CID_PREFIX = "f01701220"
```
DavidMinarsch, line 60: afaik you can import this from autonomy somewhere as well

packages/valory/agents/trader/aea-config.yaml
Adamantios: I just added these here because we definitely know they are not useful to the trader agent. We can keep the list empty, but this would mean that some txs which will cost mech fees will result in invalid responses as they do not return the required format (`p_yes`, `p_no`, etc.) until the policy stops choosing them.

packages/valory/skills/decision_maker_abci/behaviours/base.py
Adamantios: Checked it out. It is in `autonomy.chain.metadata` and would require adding `open-autonomy` as a dependency to the skill :thinking:

packages/valory/contracts/agent_registry/contract.yaml
jmoreira-valory, line 1: Why are agent registry and mech contracts on this repository ?

Adamantios: Can you explain a bit more?

jmoreira-valory: from the point of view of source code organization, these packages don't seem to belong to this repository, rather to the mech/open-autonomy repository, and then be reused/referenced here.

packages/valory/services/trader/service.yaml
```diff
@@ -104,6 +104,11 @@ type: skill
         redeeming_batch_size: ${REDEEMING_BATCH_SIZE:int:1}
         slippage: ${SLIPPAGE:float:0.01}
         redeem_margin_days: ${REDEEM_MARGIN_DAYS:int:15}
+        epsilon: ${EPSILON:float:0.1}
```
jmoreira-valory, line 107: What is this?

Adamantios: For the epsilon greedy policy.

packages/valory/contracts/agent_registry/contract.yaml
Adamantios: Yes, you are right, I will re-use the agent registry as a third party!
The mech contract should be unified at some point with the one from the mech repo but right now it would delay this PR a lot.

packages/valory/services/trader/service.yaml
jmoreira-valory: please use a more descriptive name if possible and don't require significant refactoring.

packages/valory/skills/decision_maker_abci/policy.py
```diff
@@ -0,0 +1,110 @@
...
+    n_tools: int
+    counts: List[int]
+    rewards: List[float]
+
+    @classmethod
+    def initial_state(cls, eps: float, n_tools: int) -> "EGreedyPolicy":
+        """Return an instance on its initial state."""
+        if n_tools == 0:
+            error = f"Cannot initialize an e Greedy Policy with {n_tools=}"
+            raise ValueError(error)
```
jmoreira-valory, line 57:
```suggestion
        if n_tools <= 0 || eps > 1 || eps < 0:
            error = f"Cannot initialize an e Greedy Policy with {eps=} and {n_tools=}"
            raise ValueError(error)
```

packages/valory/skills/decision_maker_abci/policy.py
```diff
@@ -0,0 +1,110 @@
...
+                error = "Attempted to remove tools using incorrect indexes!"
+                raise ValueError(error) from exc
+
+    def select_tool(self) -> Optional[int]:
+        """Select a Mech tool and return its index."""
+        if self.n_tools == 0:
+            return None
+
+        if sum(self.counts) == 0 or random.random() < self.eps:  # nosec
+            return self.random_tool
```
jmoreira-valory, line 99:
```suggestion
            return self.random_tool()
```

packages/valory/skills/decision_maker_abci/redeem_info.py
```diff
@@ -124,14 +125,17 @@ def __hash__(self) -> int:
...
     def is_winning(self) -> bool:
         """Return whether the current position is winning."""
         our_answer = self.outcomeIndex
         correct_answer = self.fpmm.current_answer_index
         return our_answer == correct_answer
+
+    @property
+    def claimable_amount(self) -> int:
+        """Get the claimable amount of the current market."""
+        amount = int(self.outcomeTokenMarginalPrice * self.outcomeTokensTraded)
```
jmoreira-valory, line 138: Is this correct? I am considering "outcomeTokensTraded" in this script https://github.com/valory-xyz/trader-quickstart/blob/04b185e0db4d65449bc86f89b194842eb765b78b/trades.py#L441-L442

packages/valory/skills/decision_maker_abci/redeem_info.py
```diff
@@ -124,14 +125,17 @@ def __hash__(self) -> int:
...
+        if self.is_winning:
+            return amount
+        return -amount
```
jmoreira-valory, line 141: Also why the claimable amount is -amount here? shouldn't it be 0?
What about markets declared invalid ?

packages/valory/contracts/agent_registry/contract.yaml
Adamantios: Done in ac2c027eb488be8b03b35849366760e4f05d33f6, reverted in 8451f3a50dc59b4f520f273aacbce699283a8913 because it would not work. 

The utilized contract method does not exist in the contract deployed on gnosis.

packages/valory/services/trader/service.yaml
Adamantios: Renamed on fb597753a6657296a20048dd42e831b4e6d9b781.

packages/valory/skills/decision_maker_abci/policy.py
Adamantios: 20bc58ad6b6b54119f9f0dabff44e767ef7554f7

packages/valory/skills/decision_maker_abci/redeem_info.py
Adamantios: I believe, you need to multiply the price by the number of conditional tokens to get the result.

packages/valory/skills/decision_maker_abci/redeem_info.py
Adamantios: The amount is negative because the policy's reward should be negative as well:

https://github.com/valory-xyz/trader/blob/e075ac6ecbcf532a4fabbbc84d4879fd7652a470/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L117

https://github.com/valory-xyz/trader/blob/e075ac6ecbcf532a4fabbbc84d4879fd7652a470/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L123

angrybayblade (approved):

dvilelaf (approved):

packages/valory/skills/decision_maker_abci/redeem_info.py
0xArdi: I agree with @Adamantios here

Adamantios: Updated in #105.

PR #105, opened 2023-10-06
Fix redeem amount

Adamantios:
https://github.com/valory-xyz/trader/pull/94#discussion_r1333075432

### DIFF
--- a/packages/valory/skills/decision_maker_abci/redeem_info.py
+++ b/packages/valory/skills/decision_maker_abci/redeem_info.py
@@ -134,7 +134,7 @@ class Trade:
     @property
     def claimable_amount(self) -> int:
         """Get the claimable amount of the current market."""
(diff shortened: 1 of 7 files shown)


0xArdi (approved):

packages/valory/skills/decision_maker_abci/redeem_info.py
```diff
@@ -134,7 +134,7 @@ def is_winning(self) -> bool:
     @property
     def claimable_amount(self) -> int:
         """Get the claimable amount of the current market."""
-        amount = int(self.outcomeTokenMarginalPrice * self.outcomeTokensTraded)
+        amount = self.outcomeTokensTraded
```
Adamantios, line 137: The previous calculation was resulting in the number of conditional tokens instead of the amount in WEI.

PR #115, opened 2023-10-26
Handle invalid responses

Adamantios:


packages/valory/skills/decision_maker_abci/behaviours/blacklisting.py
```diff
@@ -53,7 +53,7 @@ def _blacklist(self) -> Optional[str]:
         if self.synchronized_data.is_mech_price_set:
             # impose a penalty equivalent to the mech's price on the tool responsible for blacklisting the market
             tool_idx = self.synchronized_data.mech_tool_idx
-            penalty = -self.synchronized_data.mech_price
+            penalty = -self.wei_to_native(self.synchronized_data.mech_price)
```
Adamantios, line 56: The penalty should be in xDAI as the rewards are also given in this scale.

PR #121, opened 2023-10-31
Store redeeming progress

Adamantios:
The implementation allows for the redeeming round to continue from where it left off after timing out.


DavidMinarsch (approved):
I can confirm it works for me on my agent. Haven't done a code review though.

0xArdi (approved):

PR #135, opened 2023-11-15
Implement checkpoint call for staking

Adamantios:


packages/valory/skills/trader_abci/composition.py
```diff
@@ -88,9 +95,12 @@
     FailedTransactionSubmissionRound: HandleFailedTxRound,
     FinishedDecisionRequestTxRound: DecisionReceiveRound,
     FinishedBetPlacementTxRound: RedeemRound,
-    FinishedRedeemingTxRound: ResetAndPauseRound,
+    FinishedRedeemingTxRound: CallCheckpointRound,
```
DavidMinarsch, line 98: This means all services, even non-staking ones need to transition here...

packages/valory/skills/staking_abci/behaviours.py
```diff
@@ -0,0 +1,323 @@
...
+        """Set the safe_tx_hash."""
+        length = len(safe_hash)
+        if length != TX_HASH_LENGTH:
+            raise ValueError(
+                f"Incorrect length {length} != {TX_HASH_LENGTH} detected "
+                f"when trying to assign a safe transaction hash: {safe_hash}"
+            )
+        self._safe_tx_hash = safe_hash[2:]
+
+    def wait_for_condition_with_sleep(
```
DavidMinarsch, line 140: Why do we need this?

packages/valory/skills/staking_abci/behaviours.py
```diff
@@ -0,0 +1,323 @@
...
+            ETH_PRICE,
+            SAFE_GAS,
+            self.params.staking_contract_address,
+            self.checkpoint_data,
+        )
+
+    def async_act(self) -> Generator:
+        """Do the action."""
+        with self.context.benchmark_tool.measure(self.behaviour_id).local():
+            yield from self.wait_for_condition_with_sleep(self._check_service_staked)
```
DavidMinarsch, line 296: This doesn't make sense for non-staking services.

I would have preferred non-staking services never reach this FSM. But given this design, you need at least some sort of boolean to track if this service even needs to do anything here.

DavidMinarsch: I mean now that it is implemented this way, it will work..

packages/valory/skills/trader_abci/composition.py
Adamantios: Correct. Conceptually, it would not make much sense to add a check to any of the existing rounds on whether we should transition here or not. Therefore, `CallCheckpointRound` is always entered and immediately exits if the service is not staked.

packages/valory/skills/staking_abci/behaviours.py
Adamantios: In order to perform the contract calls in a serialized manner, and only continue to the next step if the current call succeeded. Otherwise, the behaviour sleeps and retries the same call.

packages/valory/skills/staking_abci/behaviours.py
Adamantios: It does have a boolean in the synced data and checks on-chain only during the first period:

https://github.com/valory-xyz/trader/blob/908bc84b6f9832d795d6700a4115c1cda402b4fa/packages/valory/skills/staking_abci/behaviours.py#L230-L232

If not staked, it does nothing.

As mentioned above, it would not make much sense conceptually to check whether the service is staked in any of the other abcis.
Moreover, this implementation makes this component as much independent as possible, so that future services can easily make use of it.

dvilelaf (approved):

DavidMinarsch (approved):

PR #136, opened 2023-11-15
Check max bet amount

### LINKED ISSUE #117: Max bet amount should be required to make progress
The max bet amount + the mech's cost should be available in order to make progress. Otherwise, the service should request funds.

Adamantios:
Checks if the max bet amount is available before sending a request to the mech, if not using the kelly criterion.

Closes #117.


dvilelaf (commented):
LGTM, there are some linter failures

0xArdi (approved):

PR #139, opened 2023-11-16
feat: add staking contract utils

0xArdi:
This PR adds a couple utils functions on the staking contract.


packages/valory/contracts/service_staking_token/contract.py
```diff
@@ -112,3 +112,26 @@ def get_next_checkpoint_ts(
...
+    @classmethod
+    def get_service_info(
+        cls,
+        ledger_api: LedgerApi,
+        contract_address: str,
+        service_id: int,
+    ) -> JSONLike:
+        """Retrieve the service info for a service."""
+        contract = cls.get_instance(ledger_api, contract_address)
+        info = contract.functions.mapServiceInfo(service_id).call()
```
jmoreira-valory, line 136: you are calling the interface method get_service_info, but you are calling "mapServiceInfo" instead of "getServiceInfo" in the contract. Is this correct?

jmoreira-valory: I would say, either use getServiceInfo, or rename to map_service_info, to be consistent with the contract interface.

0xArdi: https://github.com/valory-xyz/autonolas-registries/blob/main/contracts/staking/ServiceStakingBase.sol#L594
Its the same

jmoreira-valory (approved):
See commen

PR #148, opened 2023-11-18
fix: add missing local var

DavidMinarsch:


packages/valory/skills/decision_maker_abci/behaviours/reedem.py
```diff
@@ -769,6 +769,7 @@ def async_act(self) -> Generator:
                 self.context.logger.info(msg)
                 self._load_progress()
 
+            success = False
```
DavidMinarsch, line 772: Was missing initialisation

Issue #146, opened 2023-11-18
[FIX] Dynamic batching window for redeeming does not change


dagacha:
Release 0.9.4

`[2023-11-18 15:35:17,500] [ERROR] [agent] The RPC timed out! This usually happens if the filtering is too wide. The service tried to filter from block 30237353 to 30242353.If this issue persists, please try lowering the `EVENT_FILTERING_BATCH_SIZE`!
[2023-11-18 15:35:17,500] [WARNING] [agent] Repeating this call with a decreased batch size of 2109.
[2023-11-18 15:36:09,129] [ERROR] [agent] The RPC timed out! This usually happens if the filtering is too wide. The service tried to filter from block 30239462 to 30244462.If this issue persists, please try lowering the `EVENT_FILTERING_BATCH_SIZE`!
[2023-11-18 15:36:09,129] [WARNING] [agent] Repeating this call with a decreased batch size of 1582.
[2023-11-18 15:37:00,864] [ERROR] [agent] The RPC timed out! This usually happens if the filtering is too wide. The service tried to filter from block 30241044 to 30246044.If this issue persists, please try lowering the `EVENT_FILTERING_BATCH_SIZE`!`

Error message states the window batch size is getting smaller but the distance between the blocks reported stays constant to 5000.

DavidMinarsch:
The `batch_size` is currently (after #149) not persisted across rounds. This should be added imo to avoid going back to large badges on each round entry.

PR #149, opened 2023-11-19
fix: properly handle batch size reduction

DavidMinarsch:
Fixes https://github.com/valory-xyz/trader/issues/146

### DIFF
--- a/packages/valory/skills/decision_maker_abci/behaviours/reedem.py
+++ b/packages/valory/skills/decision_maker_abci/behaviours/reedem.py
@@ -435,10 +435,9 @@ class RedeemBehaviour(RedeemInfoBehaviour):
             if not result:
                 n_retries += 1
                 keep_fraction = (1 - self.params.reduce_factor) ** n_retries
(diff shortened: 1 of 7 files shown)


packages/valory/skills/decision_maker_abci/behaviours/reedem.py
```diff
@@ -435,10 +435,9 @@ def _check_already_redeemed(self) -> WaitableConditionType:
             if not result:
                 n_retries += 1
                 keep_fraction = (1 - self.params.reduce_factor) ** n_retries
-                step = int(batch_size * keep_fraction)
-                msg = f"Repeating this call with a decreased batch size of {step}."
+                batch_size = int(self.params.event_filtering_batch_size * keep_fraction)
+                msg = f"Repeating this call with a decreased batch size of {batch_size}."
                 self.context.logger.warning(msg)
-                from_block += step
```
DavidMinarsch, line 441: This would modify the starting point of the failed query and keep the batch size unchanged.

PR #155, opened 2023-11-21
Persist the status of the clean operation for the redeeming round

Adamantios:


DavidMinarsch (approved):

DavidMinarsch:
What about persisting the batch size as well?

Adamantios:
> What about persisting the batch size as well?

Opening a PR for that soon.

angrybayblade (approved):

PR #119, opened 2023-10-30
feat/add-kelly-strategy

jhehemann:
This pull request adds a new trading strategy "kelly_criterion" to the trader and adds the parameter TRADING_STRATEGY that contains the trading strategy in use. The favoured type of strategy can be set as an enviconment variable. The current trading strategy "bet_amount_per_conf_threshold" is still the default when starting the trader. The new trading strategy is the Kelly Criterion and it is explained here. 

Note: For utilizing the strategy in an effective way, the parameter SAMPLE_BETS_CLOSING_DAYS has also been added and can be set as an environment variable. It determines the amount of days until markets' closing times and makes the trader only bet on those markets that close wihtin SAMPLE_BETS_CLOSING_DAYS amount of days. Small values for this parameter ensure that wins can be redeemed within short periods of time and can also be reinvested earlier. As the Kelly Criterion places bets based on the amount of tokens that are available to bet, this leads to a steeper profit curve. It is also useful to gather trading data within short timeframes and make adjustments to the trader. 

# Introduction
The Kelly Criterion is the calculation rule for the Kelly fraction $f^*$, which is the portion of the gambling capital to be wagered in order to maximize the profit considering the payoff odds and the probability of winning. The calculation rule is given by

$$
f^* = \frac{bp - (1 - p)}{b},
$$

where $f^*$ is the Kelly fraction of the bankroll, which is the maximum available gambling capital, $b$ is the net odds received on the wager (bet amount excluded), $p$ is the probability of winning, and $(1-p)$ is the probability of losing ([Beygelzimer et. al., 2012](#references)).

In traditional betting frameworks, the odds for a bet are usually fixed when placing a bet and stay the same regardless of the betting amount. The Kelly Criterion was originally designed for static odds and faces some challenges when being used in scenarios with dynamic odds. In prediction markets run by Constant Product Market Makers (CPMMs) the odds depend on the bet amount. As the Kelly Criterion calculates the optimal bet amount based on the odds, there exist an interdependence of bet amount and odds. The following sections provide a way to calculate the optimal absolute Kelly bet amount depending on the Bankroll and number of outcome tokens for each outcome before placing the bet. The here derived function only applies for prediction markets run by CPMMs with binary outcomes.

# Kelly equation with dynamic odds
## General Equation
The Kelly formula can be converted into 

$$
f^* = \frac{Q \cdot p - 1}{Q - 1},
$$

[...]

PR #157, opened 2023-11-26
chore: idea for a refactor/extraction

DavidMinarsch:
To extract the contributions for minting of https://github.com/valory-xyz/trader/pull/119/files


DavidMinarsch (commented):
Whilst this is a useful direction, it's not quite where I want it. I'll explain on a call..

0xArdi (approved):

PR #166, opened 2023-12-01
Fix/redeeming

0xArdi:
This PR makes 2 changes to redeeming:
1. Introduces a lower limit for the redeeming batch size.
2. Don't retry redeeming once it times out.


packages/valory/skills/decision_maker_abci/rounds.py
```diff
@@ -184,7 +184,9 @@ class DecisionMakerAbciApp(AbciApp[Event]):
             Event.DONE: FinishedDecisionMakerRound,
             Event.NO_REDEEMING: FinishedWithoutRedeemingRound,
             Event.NO_MAJORITY: RedeemRound,
-            Event.ROUND_TIMEOUT: RedeemRound,
+            # in case of a round timeout, there likely is something wrong with redeeming
+            # it could be the RPC, or some other issue. We don't want to be stuck trying to redeem.
+            Event.ROUND_TIMEOUT: FinishedWithoutRedeemingRound,
```
0xArdi, line 189: Open to discussion here. I don't think it practically make sense to continue retrying on timeout.

Adamantios: This requires a different solution. At the moment the round times out and picks up from where it was left. Usually, it takes more than one round to redeem.

I think a better solution would be to allow for a configurable number of timeouts before we stop trying or removing the progress tracking completely and adding a separate timeout for the redeeming.

Adamantios: There's no meaning to keep the progress logic since we will start from the beginning next time:

https://github.com/valory-xyz/trader/blob/cf6eeb0dd2395db672a0d94e7f960b635d54f676/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L776

Otherwise, we update the logic to keep it across periods.

Adamantios: Because it is not period-aware as it is currently implemented. So we should either remove it or improve it to keep the progress across periods, i.e., not reset it here:

https://github.com/valory-xyz/trader/blob/cf6eeb0dd2395db672a0d94e7f960b635d54f676/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L776

0xArdi: This is never called in case of a timeout:
https://github.com/valory-xyz/trader/blob/cf6eeb0dd2395db672a0d94e7f960b635d54f676/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L776 

Also I see that `redeeming_progress` is stored on the `shared_state`, which is not period dependent. 

https://github.com/valory-xyz/trader/blob/cf6eeb0dd2395db672a0d94e7f960b635d54f676/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L193-L201

Am I missing something?

Adamantios: Imagine if the redeeming tx fails. Then we start again instead of using the previous info. The reset can be performed in the tx settlement multiplexer instead.

Adamantios (approved):

PR #167, opened 2023-12-01
[WIP]: Dynamically load strategies

Adamantios:


packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -85,6 +89,25 @@ def __init__(self, **kwargs: Any) -> None:
...
+        if STRATEGY_RUN_METHOD in globals():
+            del globals()[STRATEGY_RUN_METHOD]
+        exec(self.strategy_exec, globals())  # pylint: disable=W0122  # nosec
+        method = globals().get(STRATEGY_RUN_METHOD, None)
+        if method is None:
+            self.context.logger.error(
+                f"No {STRATEGY_RUN_METHOD!r} method was found in {self.params.trading_strategy} strategy's executable."
+            )
+            return {BET_AMOUNT_FIELD: 0}
+        return method(*args, **kwargs)
```
0xArdi, line 110: So basically the assumption here is that the strategy doesnt block. I guess its fine for now, although maybe it makes sense to run this on a separate thread/process to not block the agent.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -217,48 +240,106 @@ def check_balance(self) -> WaitableConditionType:
...
+        """
+        if self._inflight_strategy_req is not None:
+            # there already is a req in flight
+            return
+        if len(self.shared_state.strategy_to_filehash) == 0:
+            # no strategies pending to be fetched
+            return
+        for strategy, file_hash in self.shared_state.strategy_to_filehash.items():
+            self.context.logger.info(f"Fetching {strategy} strategy...")
+            ipfs_msg, message = self._build_ipfs_get_file_req(file_hash)
```
0xArdi, line 293: why do this? this class has access to the IPFS helper methods, you can use generators for it and avoid having callbacks?

packages/valory/skills/decision_maker_abci/handlers.py
```diff
@@ -48,4 +50,34 @@
 LedgerApiHandler = BaseLedgerApiHandler
 ContractApiHandler = BaseContractApiHandler
 TendermintHandler = BaseTendermintHandler
-IpfsHandler = BaseIpfsHandler
+
+
+class IpfsHandler(AbstractResponseHandler):
```
0xArdi, line 55: you could just use the default IPFS handler, no?

packages/valory/skills/decision_maker_abci/behaviours/base.py
Adamantios: Opened #169.

packages/valory/skills/decision_maker_abci/behaviours/base.py
Adamantios: Yeah this is an overkill, not sure why I chose this path. 
I opened #170 as this works for now.

packages/valory/skills/decision_maker_abci/handlers.py
Adamantios: Correct. Will be addressed in #170.

packages/valory/skills/decision_maker_abci/tests/__init__.py
```diff
@@ -1,25 +0,0 @@
-# -*- coding: utf-8 -*-
-# ------------------------------------------------------------------------------
-#
```
DavidMinarsch, line 3: Would be good to keep that and add at least one test so we can later build on it. Rather than deleting it now and adding it back in later.

0xArdi (approved):

packages/valory/skills/decision_maker_abci/tests/__init__.py
Adamantios: Done in 73567cee9c1078618f896e95ae43e00f6941af87.

PR #168, opened 2023-12-04
Fix/redeeming

0xArdi:
This PR adds support for enforcing a timeout on `check_redeemed` contract calls. 
This was added since we noticed that sometimes RPCs would not respond.


packages/valory/contracts/conditional_tokens/contract.py
```diff
@@ -30,11 +30,39 @@
...
+                # Handle the case where the execution times out
+                err = f"The RPC didn't respond in {timeout}."
+                return None, err
+
+            # Check if an error occurred
+            if isinstance(data, str):
+                # Handle the case where the execution failed
+                return None, data
+
+            return data, None
```
0xArdi, line 64: This is duplicated on purpose to avoid introducing deps between the contracts. I think it should ultimately be something offered by the ledger connection, as it already is executing these ledger interactions in a ThreadPoolExecutor. 
I think we should extend the current contract_api and ledger_api protocol to include an optional timeout field.

Adamantios (approved):

PR #178, opened 2023-12-12
Feat/strategy pacakges

0xArdi:
This PR adds support for having strategies as custom packages.


angrybayblade (approved):

strategies/kelly_criterion/component.yaml
```diff
@@ -0,0 +1,14 @@
+name: kelly_criterion
+author: valory
```
DavidMinarsch, line 2: Needs fixing

Issue #177, opened 2023-12-12
Agent crashes with missing persistent keys


jmoreira-valory:
Agent crashes randomly with this error:

```
[2023-12-12 12:20:39,708] [ERROR] [agent] <An error occurred during handle of handler valory/trader_abci:0.1.0/ABCIRoundHandler:
Traceback (most recent call last):

  File "/root/agent/vendor/valory/skills/abstract_round_abci/base.py", line 636, in create
    raise ABCIAppInternalError(

packages.valory.skills.abstract_round_abci.base.ABCIAppInternalError: internal error: Cross period persisted key `is_service_staked` was not found in the db but was required for the next period.
> raised during `<bound method AEA.handle_envelope of <aea.aea.AEA object at 0x7f44a8463980>>`
Traceback (most recent call last):
```

service version v0.10.1

PR #188, opened 2023-12-22
Ensure the cross-period keys are always set

Adamantios:
Before this change, the cross-period keys were not being set in case of a round timeout.


0xArdi (approved):

DavidMinarsch:
Fixes https://github.com/valory-xyz/trader/issues/177

PR #190, opened 2023-12-26
Mints

DavidMinarsch:


strategies/bet_amount_per_threshold/component.yaml
```diff
@@ -1,7 +1,7 @@
 name: bet_amount_per_threshold
 author: valory
 version: 0.1.0
-type: custom_component
```
DavidMinarsch, line 4: This was too verbose. It's implied that it is a component

0xArdi (approved):
Looks good. Hashes show up to be the same for me.

PR #221, opened 2024-02-15
Trader app template for olas operate app

angrybayblade:


packages/valory/services/trader_omen_gnosis/service.yaml
```diff
@@ -0,0 +1,131 @@
+name: trader_omen_gnosis
```
angrybayblade, line 1: This defines the operate app compatible template for omen-gnosis-trader agent

packages/valory/services/trader_omen_gnosis/service.yaml
```diff
@@ -0,0 +1,131 @@
...
+    nft: bafybeig64atqaladigoc3ds4arltdu63wkdrk3gesjfvnfdmz35amv7faq
+    agent_id: 14
+    threshold: 1
+    cost_of_bond: 10000000000000000
+    funds:
+      agent: 0.1
+      safe: 0.5
+  local:
+    volumes:
+      data: /data
```
angrybayblade, line 23: Right now the deployment parameter does not have constrains so we can customise it until we find a right schema and back port the changes on the open-autonomy framework

packages/valory/services/trader_omen_gnosis/service.yaml
```diff
@@ -0,0 +1,131 @@
...
+  benchmark_tool:
+    args:
+      log_dir: /benchmarks
+---
+public_id: valory/ledger:0.19.0
+type: connection
+config:
+  ledger_apis:
+    ethereum:
+      address: http://host.docker.internal:8545
```
angrybayblade, line 112: Will be updated by the operate at the runtime

dvilelaf (approved):

packages/valory/services/trader_omen_gnosis/service.yaml
jmoreira-valory: `deployment` name is confusing here. I would choose something different lik `on-chain-deployment`. LMK what you think.

Adamantios (changes requested):
Why duplicate the service config? It would be better to add this to the existing one.

packages/valory/services/trader_omen_gnosis/service.yaml
DavidMinarsch: really what this is is the "minting" specification. I think this deserves its own config file. Will explain in person.

angrybayblade:
> Why duplicate the service config? It would be better to add this to the existing one.

This spec is configured to run on the olas app with specific settings. If you check the overrides it does not have any configurable overrides it only has default values

0xArdi:
Merging to avoid having to track many branches.

PR #241, opened 2024-03-28
Add component whitelist

angrybayblade:
Needs to be minted
- contract/valory/conditional_tokens/0.1.0
- contract/valory/realitio/0.1.0
- contract/valory/realitio_proxy/0.1.0
- contract/valory/service_staking_token/0.1.0
- contract/valory/transfer_nft_condition/0.1.0
- skill/valory/staking_abci/0.1.0

### DIFF
--- a/mints/whitelist.txt
+++ b/mints/whitelist.txt
@@ -1,3 +1,2 @@
-contract/valory/agent_registry/0.1.0
 custom/valory/mike_strat/0.1.0
 custom/w1kke/always_blue/0.1.0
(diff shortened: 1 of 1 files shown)


mints/whitelist.txt
```diff
@@ -0,0 +1,3 @@
+contract/valory/agent_registry/0.1.0
```
DavidMinarsch, line 1: Why are we not minting this?

angrybayblade: I basically wrote a script which skipped the registry contracts from the service repos since they already exists on the core repo, but this is a different type of registry. I've fixed it in the latest commit

DavidMinarsch (approved):

PR #242, opened 2024-04-04
[WIP] Check for stop trading conditions

jmoreira-valory:
Check for stop trading conditions.

The trader will stop trading and transit to RedeemRound if any of these conditions occur:

* The parameter/env variable `STOP_TRADING` is set to `true`.
* The KPI fot the staking contract has been met.


jmoreira-valory:
Implemented in https://github.com/valory-xyz/trader/pull/245

PR #244, opened 2024-04-05
Update mech interact abci app to use version with unwrap tokens.

jmoreira-valory:
Update mech interact abci app to use version with unwrap tokens.


dagacha (approved):
Tested ✅

PR #245, opened 2024-04-10
Feat: stop trading conditions

jmoreira-valory:
Check for stop trading conditions.

The trader will stop trading and transit to RedeemRound if any of these conditions occur:

* The parameter/env variable `DISABLE_TRADING` is set to `true`.
* The KPI fot the staking contract has been met.

This PR requires to reactor the staking_abci Skill to extract common functionality to be used in the check_stop_trading_abci.
This PR contains a small hack to solve an issue with non-initialized data. Namely, it always requires that the trader don't skip going to the "trading path" on the first period. See detailed comment.

### DIFF
--- a/packages/valory/skills/check_stop_trading_abci/behaviours.py
+++ b/packages/valory/skills/check_stop_trading_abci/behaviours.py
@@ -86,71 +86,73 @@ class CheckStopTradingBehaviour(StakingInteractBaseBehaviour):
     def is_staking_kpi_met(self) -> WaitableConditionType:
         """Return whether the staking KPI has been met (only for staked services)."""
         yield from self.wait_for_condition_with_sleep(self._check_service_staked)
-        self.context.logger.info(f"{self.service_staking_state=}")
+        self.context.logger.debug(f"{self.service_staking_state=}")
         if self.service_staking_state != StakingState.STAKED:
             return False
 
         yield from self.wait_for_condition_with_sleep(self._get_mech_request_count)
         mech_request_count = self.mech_request_count
-        self.context.logger.info(f"{self.mech_request_count=}")
+        self.context.logger.debug(f"{self.mech_request_count=}")
 
         yield from self.wait_for_condition_with_sleep(self._get_service_info)
         mech_request_count_on_last_checkpoint = self.service_info[2][1]
-        self.context.logger.info(f"{mech_request_count_on_last_checkpoint=}")
+        self.context.logger.debug(f"{mech_request_count_on_last_checkpoint=}")
 
         yield from self.wait_for_condition_with_sleep(self._get_ts_checkpoint)
         last_ts_checkpoint = self.ts_checkpoint
-        self.context.logger.info(f"{last_ts_checkpoint=}")
+        self.context.logger.debug(f"{last_ts_checkpoint=}")
 
         yield from self.wait_for_condition_with_sleep(self._get_liveness_period)
         liveness_period = self.liveness_period
-        self.context.logger.info(f"{liveness_period=}")
+        self.context.logger.debug(f"{liveness_period=}")
 
         yield from self.wait_for_condition_with_sleep(self._get_liveness_ratio)
         liveness_ratio = self.liveness_ratio
-        self.context.logger.info(f"{liveness_ratio=}")
+        self.context.logger.debug(f"{liveness_ratio=}")
 
         mech_requests_since_last_cp = mech_request_count - mech_request_count_on_last_checkpoint
-        self.context.logger.info(f"{mech_requests_since_last_cp=}")
+        self.context.logger.debug(f"{mech_requests_since_last_cp=}")
 
         current_timestamp = self.synced_timestamp
-        self.context.logger.info(f"{current_timestamp=}")
+        self.context.logger.debug(f"{current_timestamp=}")
 
         required_mech_requests = math.ceil(max(
             (current_timestamp - last_ts_checkpoint) * liveness_ratio / LIVENESS_RATIO_SCALE_FACTOR,
             (liveness_period) * liveness_ratio / LIVENESS_RATIO_SCALE_FACTOR
         )) + REQUIRED_MECH_REQUESTS_SAFETY_MARGIN
-        self.context.logger.info(f"{required_mech_requests=}")
+        self.context.logger.debug(f"{required_mech_requests=}")
 
         if mech_requests_since_last_cp >= required_mech_requests:
             return True
         return False
 
+    def _compute_stop_trading(self) -> Generator:
+        # This is a "hacky" way of getting required data initialized on
+        # the Trader: On first period, the FSM needs to initialize some
+        # data on the trading branch so that it is available in the
+        # cross-period persistent keys.
+        if self.is_first_period:
+            self.context.logger.debug(f"{self.is_first_period=}")
+            return False
+
+        stop_trading_conditions = []
+
+        self.context.logger.debug(f"{self.params.disable_trading=}")
+        stop_trading_conditions.append(self.params.disable_trading)
+
+        self.context.logger.debug(f"{self.params.stop_trading_if_staking_kpi_met=}")
+        if self.params.stop_trading_if_staking_kpi_met:
+            staking_kpi_met = yield from self.is_staking_kpi_met()
+            self.context.logger.debug(f"{staking_kpi_met=}")
+            stop_trading_conditions.append(staking_kpi_met)
+
+        return any(stop_trading_conditions)
+
     def async_act(self) -> Generator:
         """Do the action."""
         with self.context.benchmark_tool.measure(self.behaviour_id).local():
-
-            # This is a "hacky" way of getting required data initialized on
-            # the Trader: On first period, the FSM needs to initialize some
-            # data on the trading branch so that it is available in the
-            # cross-period persistent keys.
-            if self.is_first_period:
-                stop_trading = False
-            else:
-                stop_trading_conditions = []
-
-                disable_trading = self.params.disable_trading
-                self.context.logger.info(f"{disable_trading=}")
-                stop_trading_conditions.append(disable_trading)
-
-                if self.params.stop_trading_if_staking_kpi_met:
-                    staking_kpi_met = yield from self.is_staking_kpi_met()
-                    self.context.logger.info(f"{staking_kpi_met=}")
-                    stop_trading_conditions.append(staking_kpi_met)
-
-                stop_trading = any(stop_trading_conditions)
-
-            self.context.logger.info(f"{stop_trading=}")
+            stop_trading = yield from self._compute_stop_trading()
+            self.context.logger.info(f"Computed {stop_trading=}")
             payload = CheckStopTradingPayload(
                 self.context.agent_address, stop_trading
             )
--- a/packages/valory/skills/check_stop_trading_abci/behaviours.py
+++ b/packages/valory/skills/check_stop_trading_abci/behaviours.py
@@ -1,0 +1,85 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2024 Valory AG
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""This module contains the behaviours for the check stop trading skill."""
+
+from datetime import datetime, timedelta
+from typing import Any, Callable, Generator, Optional, Set, Type, cast
+
+from aea.configurations.data_types import PublicId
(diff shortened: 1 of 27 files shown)


packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -0,0 +1,156 @@
...
+    def async_act(self) -> Generator:
+        """Do the action."""
+        with self.context.benchmark_tool.measure(self.behaviour_id).local():
+
+            # This is a "hacky" way of getting required data initialized on
+            # the Trader: On first period, the FSM needs to initialize some
+            # data on the trading branch so that it is available in the
+            # cross-period persistent keys.
+            if self.is_first_period:
+                stop_trading = False
```
jmoreira-valory, line 130: This forces the Trader to always go to the full "bet placement" cycle at least once, when restarted. This ensures that the values in the persitent data are initialized. Of course this is is a "hacky" way. However, this was the simplest way to address this issue. An alternative to fix this will be prepared in a different PR, in order to avoid too much refactoring of existing components in this PR.

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -0,0 +1,156 @@
...
+        mech_requests_since_last_cp = mech_request_count - mech_request_count_on_last_checkpoint
+        self.context.logger.info(f"{mech_requests_since_last_cp=}")
+
+        current_timestamp = self.synced_timestamp
+        self.context.logger.info(f"{current_timestamp=}")
+
+        required_mech_requests = math.ceil(max(
+            (current_timestamp - last_ts_checkpoint) * liveness_ratio / 10**18,
+            (liveness_period) * liveness_ratio / 10**18
+        )) + 1
```
jmoreira-valory, line 114: "+1" is added to avoid rounding situations on the smart contract between the moment this check is performed, and the moment the checkpoint is called (e.g., if it's called at 24h 01m, it migth cause a rounding problem). See the detailed internal document.

packages/valory/skills/check_stop_trading_abci/models.py
```diff
@@ -0,0 +1,64 @@
...
+
+Requests = BaseRequests
+BenchmarkTool = BaseBenchmarkTool
+
+
+class CheckStopTradingParams(BaseParams):
+    """CheckStopTrading parameters."""
+
+    staking_contract_address: str
+    mech_contract_address: str
```
jmoreira-valory, line 44: These are required because they are defined in another skill's params

packages/valory/skills/staking_abci/behaviours.py
```diff
@@ -59,34 +60,30 @@
 SAFE_GAS = 0
 
 
-class CallCheckpointBehaviour(BaseBehaviour):
-    """Behaviour that calls the checkpoint contract function if the service is staked and if it is necessary."""
-
-    matching_round = CallCheckpointRound
+class StakingInteractBaseBehaviour(BaseBehaviour, ABC):
```
jmoreira-valory, line 63: This class "abstracts" common **read** functionalities for the contract

packages/valory/skills/check_stop_trading_abci/models.py
Adamantios: Can you explain why not override the params of the staking skill instead?

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -0,0 +1,154 @@
...
+
+class CheckStopTradingRound(CollectSameUntilThresholdRound):
+    """A round for checking stop trading conditions."""
+
+    payload_class = CheckStopTradingPayload
+    synchronized_data_class = SynchronizedData
+    done_event = Event.DONE
+    none_event = Event.NONE
+    no_majority_event = Event.NO_MAJORITY
+    selection_key = (get_name(SynchronizedData.stop_trading),)
```
Adamantios, line 82:
```suggestion
    selection_key = get_name(SynchronizedData.stop_trading)
```

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -0,0 +1,154 @@
...
+
+        return synced_data, event
+
+
+class FinishedCheckStopTradingRound(DegenerateRound, ABC):
+    """A round that represents check stop trading has finished."""
+
+
+class FinishedCheckStopTradingWithSkipTradingRound(DegenerateRound, ABC):
+    """A round that represents check stop trading has finished with skip trading."""
```
Adamantios, line 105: It is not necessary to prefix with `FinishedCheckStopTrading`. Could you please change the names to something more intuitive?

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -0,0 +1,162 @@
...
+    def is_first_period(self) -> bool:
+        """Return whether it is the first period of the service."""
+        return self.synchronized_data.period_count == 0
+
+    @property
+    def params(self) -> CheckStopTradingParams:
+        """Return the params."""
+        return cast(CheckStopTradingParams, self.context.params)
+
+    def is_staking_kpi_met(self) -> WaitableConditionType:
```
Adamantios, line 79: Why does this return `WaitableConditionType` since you do not wait for `True` from the caller method?

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -0,0 +1,162 @@
...
+        mech_requests_since_last_cp = mech_request_count - mech_request_count_on_last_checkpoint
+        self.context.logger.info(f"{mech_requests_since_last_cp=}")
+
+        current_timestamp = self.synced_timestamp
+        self.context.logger.info(f"{current_timestamp=}")
+
+        required_mech_requests = math.ceil(max(
+            (current_timestamp - last_ts_checkpoint) * liveness_ratio / 10**18,
+            (liveness_period) * liveness_ratio / 10**18
+        )) + 1
```
Adamantios, line 115: Please specify constants for `10**18` and `1`.

packages/valory/skills/check_stop_trading_abci/behaviours.py
Adamantios: Let's open an issue for that and not release until this is done.

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -0,0 +1,162 @@
...
+                disable_trading = self.params.disable_trading
+                self.context.logger.info(f"{disable_trading=}")
+                stop_trading_conditions.append(disable_trading)
+
+                if self.params.stop_trading_if_staking_kpi_met:
+                    staking_kpi_met = yield from self.is_staking_kpi_met()
+                    self.context.logger.info(f"{staking_kpi_met=}")
+                    stop_trading_conditions.append(staking_kpi_met)
+
+                stop_trading = any(stop_trading_conditions)
```
Adamantios, line 144: Please extract to a method.

packages/valory/skills/check_stop_trading_abci/behaviours.py
Adamantios, line 1: The implemented behaviour is logging some information at the `INFO` level, which would be more suitable for the `DEBUG` level.

Adamantios (changes requested):

packages/valory/skills/check_stop_trading_abci/behaviours.py
jmoreira-valory: Fixed.

packages/valory/skills/check_stop_trading_abci/behaviours.py
jmoreira-valory: Opened an issue https://github.com/valory-xyz/trader/issues/246

packages/valory/skills/check_stop_trading_abci/behaviours.py
jmoreira-valory: This call is mimmicking similar calls from staking_abci:

https://github.com/valory-xyz/trader/blob/69af494465a802fe80f4dd5ce6d53f0b1f14e010/packages/valory/skills/staking_abci/behaviours.py#L247

https://github.com/valory-xyz/trader/blob/69af494465a802fe80f4dd5ce6d53f0b1f14e010/packages/valory/skills/staking_abci/behaviours.py#L255

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @property
+    def stop_trading(self) -> bool:
+        """Get if the service must stop trading."""
+        return bool(self.db.get("stop_trading", False))
+
+    @property
+    def participant_to_selection(self) -> DeserializedCollection:
+        """Get the participants to selection round."""
+        return self._get_deserialized("participant_to_selection")
```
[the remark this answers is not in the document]
jmoreira-valory: Fixed.

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -0,0 +1,154 @@
...
+        """Process the end of the block."""
+        res = super().end_block()
+        if res is None:
+            return None
+
+        synced_data, event = cast(Tuple[SynchronizedData, Enum], res)
+        stop_trading_payload = self.most_voted_payload
+
+        if event == Event.DONE and stop_trading_payload == True:
+            return synced_data, Event.SKIP_TRADING
```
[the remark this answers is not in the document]
jmoreira-valory: Can you elaborate on that?

packages/valory/skills/check_stop_trading_abci/rounds.py
jmoreira-valory: Fixed.

packages/valory/skills/check_stop_trading_abci/behaviours.py
jmoreira-valory: Fixed.

packages/valory/skills/check_stop_trading_abci/models.py
jmoreira-valory: Fixed. The reasoning was that I thought it was a bit uglier how the Params need to be loader on the composed app now:

```
class TraderParams(
    DecisionMakerParams,  # It also contains MechInteractParams
    TerminationParams,
    TxSettlementMultiplexerParams,
    CheckStopTradingParams,  # It also contains StakingParams
):
```

... but I noticed it already happens for `MechInteractParams`

packages/valory/skills/check_stop_trading_abci/behaviours.py
jmoreira-valory: I don't think we should block the release for this issue. Of course it's not the ideal behaviour (it always require one mech call at restart) but at least it's better than the current implementation (it requires indefinite amount of mech calls).

packages/valory/skills/check_stop_trading_abci/behaviours.py
jmoreira-valory: Fixed.

packages/valory/skills/check_stop_trading_abci/behaviours.py
Adamantios: Ok, but this needs to be clear in the release announcement.

packages/valory/skills/check_stop_trading_abci/rounds.py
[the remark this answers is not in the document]
Adamantios: I mean this is a voting round. The payload only includes a single boolean value. Why hack around with collect same instead of using a voting round?

packages/valory/skills/check_stop_trading_abci/behaviours.py
Adamantios: Yes, but these are consumed by the `wait_for_condition_with_sleep` method.

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -135,20 +130,20 @@ class CheckStopTradingAbciApp(AbciApp[Event]):  # pylint: disable=too-few-public
...
+        FinishedWithSkipTradingRound,
     }
     event_to_timeout: Dict[Event, float] = {
         Event.ROUND_TIMEOUT: 30.0,
     }
     db_pre_conditions: Dict[AppState, Set[str]] = {CheckStopTradingRound: set()}
     db_post_conditions: Dict[AppState, Set[str]] = {
         FinishedCheckStopTradingRound: set(),
-        FinishedCheckStopTradingWithSkipTradingRound: set(),
+        FinishedWithSkipTradingRound: set(),
```
Adamantios: How about `ResumeTradingRound` and `StopTradingRound`? Easier to read the composition.

jmoreira-valory: Because this is computed on each iteration: on each iteration it's decided whether to go to the "trading branch" and whether skip the trading branch. "Resume" is not an approppriate name here, IMO (it can simply always trade, so it loses the meaning of "resume").

packages/valory/skills/check_stop_trading_abci/behaviours.py
jmoreira-valory: Ok, it will be included in the release announcement.

packages/valory/skills/check_stop_trading_abci/rounds.py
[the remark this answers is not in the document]
jmoreira-valory: Fixed. Used a VotingRound instead.

Adamantios (approved):

PR #261, opened 2024-05-27
Parameterize the sleep time in-between unsuccessful RPC calls

Adamantios:


Adamantios:
This is a WIP as it is also pending a bump of the mech interact package in https://github.com/valory-xyz/IEKit/pull/171 in order to be complete.

dvilelaf (approved):

PR #270, opened 2024-06-10
Do not blacklist if the tool selection has not been run

Adamantios:
This fixes an edge case caused when the tx settlement round fails and no tool has been selected. This can happen in two cases:
1. After the nevermined subscription round
2. After the KPI has been met and the tool selection is skipped intentionally. 

If this happens, the `HandleFailedTxRound` is called, and then the `BlacklistingRound`. During the blacklisting, no tool has been selected for the current round, and therefore we get:

```
ValueError: 'mech_tool_idx' field is not set for this period [x] and no default value was provided.
```

In order to fix this, we simply check whether the tool selection has been run, and only then proceed to the blacklisting operation.


0xArdi (approved):

dagacha:
Tested - LGTM

PR #275, opened 2024-06-12
Make all the api specs configurable

Adamantios:


dvilelaf (approved):

packages/valory/agents/trader/aea-config.yaml
```diff
@@ -249,6 +249,77 @@ models:
...
       results_filename: ${str:benchmarking_results.csv}
+  network_subgraph:
+    args:
+      headers:
+        Content-Type: ${str:application/json}
+      method: ${str:POST}
+      response_key: ${str:data:blocks}
+      response_index: ${int:0}
+      response_type: ${str:dict}
+      retries: ${int:5}
```
jmoreira-valory, line 260: Why you need all this here? isn't enough with the URL ?

Same applies below.

jmoreira-valory (approved):

PR #277, opened 2024-06-13
Fix file not found error

Adamantios:
Handle the situation where the utilized tools file is missing from local storage.


dagacha (approved):
Tested - fixed the issue

PR #281, opened 2024-06-19
Feat/liquidity info

cyberosa:
Feature to include liquidity information at the benchmarking results output file including the initial and final liquidity values of the two options in the pool: option 0 and option 1.


packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -618,6 +624,13 @@ def _write_benchmark_results(
             )
             return
 
+        if old_amounts is not None:
+            l0_start, l0_end = old_amounts[0], new_amounts[0]
+            l1_start, l1_end = old_amounts[1], new_amounts[1]
```
Adamantios, line 629: Let's convert this to a structure, for example, a dataclass called `Liquidity` that has the l0s and l1s.

Adamantios (commented):
Looks good! Need to address the CI failures though.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -277,7 +277,11 @@ def _get_mocked_bet(self) -> Bet:
         liquidity_amounts = self.shared_state.liquidity_amounts
         liquidity_prices = self.shared_state.liquidity_prices
         markets = list(liquidity_amounts.keys())
-        question_id = self.shared_state.mock_data.id
+        if self.shared_state.mock_data is not None:
+            question_id = self.shared_state.mock_data.id
+        else:
+            raise ValueError("No mocked data information")
```
Adamantios, line 283: This is duplicated below, you may create a property for safely accessing the `self.shared_state.mock_data`.

Adamantios (commented):
Not exactly what I had in mind. Old amounts and new amounts lists should be completely replaced by a single object of the liquidity info. That way we do not deal with indexes etc at all, we enhance readability and protect against index errors.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -609,6 +614,7 @@ def _write_benchmark_results(
         p_no: Optional[float] = None,
         confidence: Optional[float] = None,
         bet_amount: Optional[float] = None,
+        liquidity_info: Optional[LiquidityInfo] = None,
```
Adamantios, line 617:
```suggestion
        liquidity_info: LiquidityInfo,
```

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -618,6 +624,9 @@ def _write_benchmark_results(
             )
             return
 
+        if liquidity_info is None:
+            self.context.logger.info("No market liquidity information.")
```
Adamantios, line 628: Here you can instead check if all the attributes are `None`. You can create a method in the dataclass for that.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -647,6 +660,10 @@ def _write_benchmark_results(
                 p_no,
                 confidence,
                 bet_amount,
+                liquidity_info.l0_start if liquidity_info is not None else None,
+                liquidity_info.l1_start if liquidity_info is not None else None,
+                liquidity_info.l0_end if liquidity_info is not None else None,
+                liquidity_info.l1_end if liquidity_info is not None else None,
```
Adamantios, line 666:
```suggestion
                liquidity_info.l0_start,
                liquidity_info.l1_start,
                liquidity_info.l0_end,
                liquidity_info.l1_end,
```

packages/valory/skills/decision_maker_abci/models.py
```diff
@@ -88,6 +92,25 @@ class PromptTemplate(Template):
...
+    l1_start: int
+    # Liquidity of tokens for option 0, after placing the bet
+    l0_end: int
+    # Liquidity of tokens for option 1, after placing the bet
+    l1_end: int
+
+    def __init__(self, l0_start, l1_start, l0_end, l1_end):
+        """Function to update the liquidity amounts"""
+        self.l0_start, self.l0_end = l0_start, l0_end
+        self.l1_start, self.l1_end = l1_start, l1_end
```
Adamantios, line 111:
```suggestion
    # Liquidity of tokens for option 0, before placing the bet
    l0_start: Optional[int]
    # Liquidity of tokens for option 1, before placing the bet
    l1_start: Optional[int]
    # Liquidity of tokens for option 0, after placing the bet
    l0_end: Optional[int]
    # Liquidity of tokens for option 1, after placing the bet
    l1_end: Optional[int]
```

packages/valory/skills/decision_maker_abci/models.py
```diff
@@ -88,6 +92,30 @@ class PromptTemplate(Template):
...
+    )
+    l1_end: Optional[int] = (
+        None  # Liquidity of tokens for option 1, after placing the bet
+    )
+
+    def update_liquidity_amounts(self, old_amounts: List[int], new_amounts: List[int]):
+        """Function to update the liquidity amounts"""
+        self.l0_start, self.l0_end = old_amounts[0], new_amounts[0]
+        self.l1_start, self.l1_end = old_amounts[1], new_amounts[1]
+        return
```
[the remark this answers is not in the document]
cyberosa: corrected

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
cyberosa: Without the BenchmarkingMockData there is no benchmarking mode so I guess this use-case is already covered in the code even before. No extra checks needed here.

packages/valory/skills/decision_maker_abci/behaviours/base.py
cyberosa: We cannot declare it like this without a default value, because the previous parameters have default values.

packages/valory/skills/decision_maker_abci/behaviours/base.py
cyberosa: done

packages/valory/skills/decision_maker_abci/models.py
cyberosa: done

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -618,6 +624,9 @@ def _write_benchmark_results(
             )
             return
 
+        if liquidity_info.l0_start is None:
+            self.context.logger.info("No market liquidity information.")
```
dvilelaf, line 628: I see you are using liquidity_info on line 663-666. If this is None, shouldnt we handle this case differently?

cyberosa: Corrected

Adamantios (approved):

PR #282, opened 2024-06-19
Fix/healthcheck

dvilelaf:
* Adds local ip support to the healthcheck


jmoreira-valory (approved):

Adamantios (approved):
Maybe we could make the regex configurable in future implementations.

PR #290, opened 2024-07-12
Fix/test check stop trading

annasambrook:
Payload test for check stop trading added.


pyproject.toml
```diff
@@ -32,7 +32,7 @@ web3 = "<7,>=6.0.0"
 ipfshttpclient = "==0.8.0a2"
 open-aea-cli-ipfs = "==1.52.0"
 aiohttp = "<4.0.0,>=3.8.5"
-certifi = "*"
+certifi = "==2021.10.8"
```
annasambrook, line 35: Any idea why this has changed?

Adamantios (commented):
Looks good, just lock the deps.

pyproject.toml
Adamantios: I think you ran `tox -e check-dependencies`.

Adamantios (approved):

PR #291, opened 2024-07-12
Fix benchmarking policy

Adamantios:


packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
```diff
@@ -54,7 +54,8 @@ def async_act(self) -> Generator:
             selected_tool = yield from self._select_tool()
             if selected_tool is not None:
                 # the period will increment when the benchmarking finishes
-                if self.benchmarking_mode.enabled and self.is_first_period:
+                benchmarking_running = self.synchronized_data.period_count == 0
+                if self.benchmarking_mode.enabled and benchmarking_running:
```
Adamantios, line 58: The property does not do the same:

https://github.com/valory-xyz/trader/blob/684f96491fb9f37f934e449e788ab1553e276371/packages/valory/skills/decision_maker_abci/behaviours/base.py#L256-L263

cyberosa (approved):

PR #294, opened 2024-07-16
Fix the randomness while in benchmarking mode

Adamantios:
The randomness should only be set once when benchmarking.

### DIFF
--- a/packages/valory/skills/decision_maker_abci/policy.py
+++ b/packages/valory/skills/decision_maker_abci/policy.py
@@ -137,12 +137,14 @@ class EGreedyPolicy:
             for tool, acc_info in self.accuracy_store.items()
         }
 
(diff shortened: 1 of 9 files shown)


packages/valory/skills/decision_maker_abci/policy.py
```diff
@@ -137,12 +137,14 @@ def update_weighted_accuracy(self) -> None:
             for tool, acc_info in self.accuracy_store.items()
         }
 
-    def select_tool(self, randomness: RandomnessType) -> Optional[str]:
+    def select_tool(self, randomness: RandomnessType = None) -> Optional[str]:
```
cyberosa, line 140: you do not need to change the calls to this function so that only once the seed is past as a parameter?

cyberosa (approved):

PR #296, opened 2024-07-17
[v0.17.0]: Update the policy's logic

Adamantios:


packages/valory/skills/decision_maker_abci/policy.py
```diff
@@ -131,9 +131,8 @@ def best_tool(self) -> str:
     def update_weighted_accuracy(self) -> None:
         """Update the weighted accuracy for each tool."""
         self.weighted_accuracy = {
-            tool: (acc_info.accuracy / 100)
-            * (acc_info.requests - acc_info.pending)
-            / self.n_requests
+            tool: acc_info.accuracy
```
Adamantios, line 134: Removed division by `100` so that the weighted accuracy is on the same scale as the accuracy, i.e., `[0 - 100]`, not `[0 - 1]`.

cyberosa, line 134: You forgot to add the "/ 100" at the acc_info.accuracy

cyberosa (changes requested):

packages/valory/skills/decision_maker_abci/policy.py
Adamantios: Did you see https://github.com/valory-xyz/trader/pull/296#discussion_r1681337009 above?

cyberosa: the weighted accuracy cannot be between 0 and 100 because we are adding an extra number to the tool accuracy, which is the volume of requests factor

cyberosa: since we are dividing now the volume factor by 2 this means the maximum number this variable can reach is 0.5 so the maximum number the weighted accuracy could reach is 100.5

Adamantios: 81c547d7d202568758c2fa2cc04a63b24b0e7bf1

packages/valory/skills/decision_maker_abci/utils/scaling.py
```diff
@@ -0,0 +1,65 @@
...
+        elif num > max_:
+            max_ = num
+
+    return min_, max_
+
+
+def scale_value(
+    value: float,
+    min_max_bounds: tuple[float, float],
+    scale_bounds: tuple[float, float] = (0, 1),
```
cyberosa, line 46: We cannot change the target range to a different value from [0, 1] because it will impact now the value of the dynamic kelly fraction, so I request removing this possibility from here and simplifying this function

packages/valory/skills/decision_maker_abci/utils/scaling.py
```diff
@@ -0,0 +1,65 @@
...
+) -> float:
+    """Perform min-max scaling on a value."""
+    min_, max_ = min_max_bounds
+    current_range = max_ - min_
+    # normalize between 0-1
+    std = (value - min_) / current_range
```
cyberosa, line 52: just after this line you can do `return std`

cyberosa (commented):
I prefer to remove the option to change the target range because it will affect the kelly fraction

packages/valory/skills/decision_maker_abci/utils/scaling.py
Adamantios: There is no reason to remove that from the util method. The change is blocked upstream.

packages/valory/skills/decision_maker_abci/utils/scaling.py
Adamantios: https://github.com/valory-xyz/trader/pull/296#discussion_r1683018968

cyberosa (approved):

PR #303, opened 2024-07-26
Feat/nevermined v2

0xArdi:
This PR updates nevermined config to use the latest subscription.


packages/valory/skills/decision_maker_abci/states/order_subscription.py
```diff
@@ -67,7 +67,7 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
             return None
 
         sync_data, event = update
-        agreement_id = self.most_voted_payload_values[2]
+        agreement_id = self.most_voted_payload_values[3]
```
0xArdi, line 70: Because of the recent MultisigTx payload changes.

annasambrook (approved):

packages/valory/skills/decision_maker_abci/states/order_subscription.py
Adamantios: You could create an instance of the `SubscriptionPayload` and access the property to be safe instead of using the index.

Adamantios (approved):

packages/valory/skills/decision_maker_abci/states/order_subscription.py
0xArdi: @Adamantios https://github.com/valory-xyz/trader/issues/304

PR #325, opened 2024-09-12
fix: disable acn

0xArdi:
This PR disables the ACN for the trader.

### DIFF
--- a/packages/valory/agents/trader/aea-config.yaml
+++ b/packages/valory/agents/trader/aea-config.yaml
@@ -365,6 +365,7 @@ cert_requests:
   not_before: '2023-01-01'
   public_key: ${str:02d3a830c9d6ea1ae91936951430dee11f4662f33118b02190693be835359a9d77}
   save_path: .certs/acn_cosmos_9005.txt
+is_abstract: true
 ---
 public_id: valory/ledger:0.19.0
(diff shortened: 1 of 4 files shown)


packages/valory/agents/trader/aea-config.yaml
```diff
@@ -365,6 +365,7 @@ cert_requests:
   not_before: '2023-01-01'
   public_key: ${str:02d3a830c9d6ea1ae91936951430dee11f4662f33118b02190693be835359a9d77}
   save_path: .certs/acn_cosmos_9005.txt
+is_abstract: true
```
0xArdi, line 368: I couldn't make this configurable, there's an issue:
```
Updating hashes...
Error: Configuration of component (connection, valory/p2p_libp2p_client:0.1.0) is not valid. The following errors occurred during validation:
 - is_abstract: '${bool:false}' is not of type 'boolean'

```

0xArdi: @Adamantios looks like an issue in core.

Adamantios: I think it's a known limitation, we cannot override anything on the top level.

Adamantios (approved):

PR #342, opened 2024-10-25
Fix tool selection

Adamantios:
We should only select relevant tools even if we have global information about others.

### DIFF
--- a/packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
+++ b/packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
@@ -267,7 +267,7 @@ class StorageManagerBehaviour(DecisionMakerBaseBehaviour, ABC):
 
         return True
 
-    def _update_accuracy_store(self, local_tools: List[str]) -> None:
+    def _update_accuracy_store(self) -> None:
         """Update the accuracy store file with the latest information available"""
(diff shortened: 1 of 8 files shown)


packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
```diff
@@ -276,9 +276,17 @@ def _update_accuracy_store(self, local_tools: List[str]) -> None:
...
+        # remove tools which are irrelevant
+        for tool in accuracy_store.copy():
+            if tool not in self.mech_tools:
+                accuracy_store.pop(tool, None)
+
         # update the accuracy store using the latest accuracy information (only entered during the first period)
         for row in reader:
             tool = row[self.acc_info_fields.tool]
+            if tool not in accuracy_store.keys():
+                continue
```
cyberosa, line 288: what is the rationale behind this? Ignoring a tool not present at the local store?

annasambrook (approved):

packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
Adamantios: Exactly. Keep in mind the store only contains the relevant tools.

PR #346, opened 2024-10-30
Hotfix: no default value for sample bet index

cyberosa:
When running the latest trader version with the Benchmarking mode active, an error was found because there was no default value for the index at the synchronized data

self.synchronized_data.sampled_bet_index

In order to have the Benchmarking running again we need to introduce a default value.


Adamantios (changes requested):
I do not agree with this approach because it changes the original flow. Instead, I would suggest to only edit the benchmark related path.

PR #354, opened 2024-11-15
Refactor (trader_pearl): Take ledger address from env-var

OjusWiZard:


packages/valory/services/trader_pearl/service.yaml
```diff
@@ -126,7 +126,7 @@ type: connection
 config:
   ledger_apis:
     ethereum:
-      address: http://host.docker.internal:8545
+      address: ${GNOSIS_LEDGER_RPC:str:http://host.docker.internal:8545}
       chain_id: 100
       default_gas_price_strategy: eip1559
       poa_chain: false
```
jmoreira-valory, line 132: please also override these ones.

jmoreira-valory: see for example https://github.com/valory-xyz/optimus/blob/main/packages/valory/services/optimus/service.yaml#L113-L116

jmoreira-valory (approved):

PR #344, opened 2024-10-28
[DO NOT MERGE] Release/0.18.4

jmoreira-valory:
[DO NOT MERGE] Release/0.18.4

PR #356, opened 2024-11-20
Update Pearl trader

jmoreira-valory:
Update Pearl trader for the refactored branch on olas-operate.

:warning: Important: This update is for the current trader_pearl used in olas-operate #344, which is referenced here: https://github.com/valory-xyz/olas-operate-app/blob/da917f717745a09f4548d0d239f4e4e6e8394748/frontend/constants/serviceTemplates.ts#L7  and introduces the changes required for the refactored branch https://github.com/valory-xyz/olas-operate-app/pull/428, which mainly concerns the introduction of environment variables.

### DIFF
--- a/packages/valory/services/trader_pearl/service.yaml
+++ b/packages/valory/services/trader_pearl/service.yaml
@@ -38,7 +38,7 @@ models:
       tendermint_url: ${TENDERMINT_URL:str:http://localhost:26657}
       tendermint_p2p_url: ${TM_P2P_ENDPOINT_NODE_0:str:node0:26656}
       creator_per_subgraph:
-        omen_subgraph: ${OMEN_CREATORS:list:["0x89c5cc945dd550BcFfb72Fe42BfF002429F46Fec"]}
+        omen_subgraph: ${OMEN_CREATORS:list:["0xFfc8029154ECD55ABED15BD428bA596E7D23f557"]}
       slot_count: ${SLOT_COUNT:int:2}
       opening_margin: ${OPENING_MARGIN:int:300}
       languages: ${LANGUAGES:list:["en_US"]}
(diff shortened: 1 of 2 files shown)


OjusWiZard:
diff from `bafybeidicxsruh3r4a2xarawzan6ocwyvpn3ofv42po5kxf7x6ck7kn22u` to this `trader_pear/service.yaml` is this: https://www.diffchecker.com/HY2yJFMe/

jmoreira-valory:
@Adamantios confirm that the current trader_pearl configuration works for 0.18.2 / 0.18.7

Adamantios:
> @Adamantios confirm that the trader_pearl configuration in this PR is appropriate for 0.18.2 / 0.18.7

The configs look good from a framework's perspective. Could you elaborate on that more?

Adamantios (approved):

jmoreira-valory:
The comment is about if you could check if the  pearl_trader.yaml is missing (or has an inconvenient value) for any config params that have been introduced recently, in terms of of tools/multibets/simulation/... etc.

Adamantios:
> The comment is about if you could check if the pearl_trader.yaml is missing (or has an inconvenient value) for any config params that have been introduced recently, in terms of of tools/multibets/simulation/... etc.

You could diff with the main service config and use the exact same defaults. I think we could merge the two service files anyway.

packages/valory/services/trader_pearl/service.yaml
```diff
@@ -74,15 +73,22 @@ models:
       slippage: 0.01
       store_path: ./data
```
OjusWiZard, line 74: can we have this like `${STORE_PATH:str:/data/}`? and others also of the same format so that they can be overridable from the environment variables?

jmoreira-valory: Yes I don't see any issue with making this change.

jmoreira-valory:
Closed by https://github.com/valory-xyz/trader/pull/381/

PR #358, opened 2024-11-22
Hotfix/staking kpi reachability

keshav1998:


annasambrook (dismissed):

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -93,6 +94,7 @@ def read_bets(self) -> None:
             with open(self.bets_filepath, READ_MODE) as bets_file:
                 try:
                     self.bets = json.load(bets_file, cls=BetsDecoder)
+
```
Adamantios, line 97:
```suggestion
```

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -154,6 +163,11 @@ def _update_bets(
             if self.synced_time >= bet.openingTimestamp - self.params.opening_margin:
                 bet.blacklist_forever()
 
+        # Extract the last queue number from the bets
+        # This is used to determine the next queue number
+        if self.bets:
+            self.current_queue_number = max(bet.queue_no for bet in self.bets)
```
Adamantios, line 169: Isn't this the same as `self.current_queue_number -= 1`? Also, does the logic change if this gets removed completely since `current_queue_number` is never accessed again after this point?

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -99,15 +99,49 @@ def processable_bet(self, bet: Bet) -> bool:
...
+        The sampling logic follows the specified priority logic:
+        1. Filter out all the bets that have a processed_timestamp != 0 to get a list of new bets.
+        2. If the list of new bets is not empty:
+           2.1 Order the list in decreasing order of liquidity (highest liquidity first).
+           2.2 For bets with the same liquidity, order them in decreasing order of market closing time (openingTimestamp).
+        3. If the list of new bets is empty:
+           3.1 Order the bets in decreasing order of invested_amount.
+           3.2 For bets with the same invested_amount, order them in increasing order of processed_timestamp (least recently processed first).
+           3.3 For bets with the same invested_amount and processed_timestamp, order them in decreasing order of liquidity.
+           3.4 For bets with the same invested_amount, processed_timestamp, and liquidity, order them in decreasing order of market closing time (openingTimestamp).
```
Adamantios, line 113: Why increasing order of processed timestamp and why decreasing order of market closing time? Sorry if this already mentioned in the spec.

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -138,6 +141,12 @@ def _update_bets(
         self,
     ) -> Generator:
         """Fetch the questions from all the prediction markets and update the local copy of the bets."""
+
+        # Extract the last queue number from the bets
+        # This is used to determine the next queue number
+        if self.bets:
+            self.current_queue_number = max(bet.queue_no for bet in self.bets) + 1
```
Adamantios, line 148: This increases the queue's number in each period! Instead, it should be increasing when it is fully consumed.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -99,15 +99,49 @@ def processable_bet(self, bet: Bet) -> bool:
...
-        return self.bets.index(max(bets))
+
+        # Filter out all the best with the smallest queue number
+        least_queue_number = min([bet.queue_no for bet in bets])
+        priority_bets = [bet for bet in bets if bet.queue_no == least_queue_number]
+
+        # Filter out all the bets that have a processed_timestamp == 0
+        new_in_priority_bets = [bet for bet in bets if bet.processed_timestamp == 0]
+
+        if new_in_priority_bets:
```
Adamantios, line 126: Here we always give priority to the new bets arriving. However, the spec says that priority to new bets should only be done on startup, i.e., `current_queue_number == 0`

Adamantios (changes requested):
This needs a bit more work to align with the spec.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
annasambrook: This is laid out in the specs doc in the 'What should be the ordering of the markets–Rational' section

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -340,6 +340,21 @@ def check_balance(self) -> WaitableConditionType:
...
+        self.read_bets()
+        # Update the bet's invested amount, the new bet amount is added to previous invested amount
+        self.bets[
+            self.synchronized_data.sampled_bet_index
+        ].invested_amount += self.synchronized_data.bet_amount
+
+        # Update bet transaction timestamp
+        self.bets[
+            self.synchronized_data.sampled_bet_index
+        ].transaction_processed_timestamp = self.synced_timestamp
```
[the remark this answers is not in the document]
annasambrook: Done.

packages/valory/skills/decision_maker_abci/behaviours/reedem.py
```diff
@@ -966,6 +969,9 @@ def async_act(self) -> Generator:
             if not success:
                 return None
 
+            if self.synchronized_data.tx_submitter == BetPlacementRound.auto_round_id():
+                self.update_bet_transaction_information()
```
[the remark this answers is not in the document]
annasambrook: Done.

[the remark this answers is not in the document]
Adamantios: What I meant was, why are we checking `self.synchronized_data.tx_submitter == BetPlacementRound.auto_round_id()` here? It might not be intuitive for people who are not familiar with the tx settlement multiplexer and the trader's code.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
Adamantios: From the spec:

> The mech is likely to give out a more well-informed response 
    a. true for older markets
    b. Amongst the older markets–more likely to happen for markets that have been processed less recently

Therefore, I think that a. translates to *increasing* order of market creation - not closing - time.

To be clarified -> maybe the spec meant "true for markets closing sooner", therefore, *increasing* order of market closing time.

packages/valory/skills/market_manager_abci/behaviours.py
Adamantios: What confuses me is that this part of the code comes after using the value here:
https://github.com/valory-xyz/trader/blob/6e580904e713aebd0f3aece152c8bccd89d8663e/packages/valory/skills/market_manager_abci/behaviours.py#L156

and then next time it is used again it takes a new value here:
https://github.com/valory-xyz/trader/blob/6e580904e713aebd0f3aece152c8bccd89d8663e/packages/valory/skills/market_manager_abci/behaviours.py#L148

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
keshav1998: > More formally: The set of all available markets should be ordered as follows: 
> 
> Decreasing order of investment amount (highest investment first)
> Increasing order of last_processed_time (least recently processed first)
> Decreasing order of liquidity (highest liquidity first)
> Decreasing order of market_closing_time (added for completeness, as a tie-breaker when the above are all same)


@Adamantios  I have followed this final laid plan by Juli in the doc, will confirm with here

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -526,7 +526,6 @@ def _update_selected_bet(
             # update the bet's timestamp of processing and its number of bets for the given
             sampled_bet = self.sampled_bet
             sampled_bet.n_bets += 1
-            sampled_bet.processed_timestamp = self.synced_timestamp
```
Adamantios, line 529: I wonder why we only moved this update into `update_bet_transaction_information` and left the `n_bets` update here. Isn't that inconsistent?

We could remove the decrease in the blacklisting behaviour if that was your concern.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -103,15 +103,82 @@ def processable_bet(self, bet: Bet, now: int) -> bool:
...
+
+            if all_bets_not_processed:
+                # if none of the bets have been processed, then we should set the current_queue_number to 0
+                # for all none blacklisted bets
+                for bet in self.bets:
+                    if bet.queue_no != -1:
+                        bet.queue_no = 0
+
+                self.store_bets()
+
```
Adamantios, line 162: Why are you storing here?

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -103,15 +103,82 @@ def processable_bet(self, bet: Bet, now: int) -> bool:
...
+                bets.sort(
+                    key=lambda bet: (
+                        bet.invested_amount,
+                        -bet.processed_timestamp,  # Increasing order of processed_timestamp
+                        bet.scaledLiquidityMeasure,
+                        bet.openingTimestamp,
+                    ),
+                    reverse=True,
+                )
+                return self.bets.index(bets[0])
```
Adamantios, line 147: This is duplicated below. Let's extract a method.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
keshav1998: yup that was my exact concern, I didn't want to change many bheaviours, will fix it then

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
keshav1998: ahh sorry, I didn't see that async act already has this, will remove this

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -48,16 +47,7 @@ def __init__(self, **kwargs: Any) -> None:
 
     def setup(self) -> None:
```
keshav1998, line 48: @Adamantios  I have emptied setup as randomness is no longer a factor in deciding upon a bet

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -351,6 +352,26 @@ def check_balance(self) -> WaitableConditionType:
...
+        sampled_bet.n_bets += 1
+        # Update Queue number for priority logic
+        if sampled_bet.queue_status == QueueStatus.TO_PROCESS:
+            sampled_bet.queue_status = QueueStatus.PROCESSED
+        elif sampled_bet.queue_status == QueueStatus.PROCESSED:
+            sampled_bet.queue_status = QueueStatus.REPROCESSED
+        else:
+            raise ValueError(
+                f"Invalid queue number {sampled_bet.queue_status} detected. This bet should not have been sampled"
+            )
```
Adamantios, line 372:
```suggestion
            raise ValueError(
                f"Invalid queue status {sampled_bet.queue_status} detected. This bet should not have been sampled!"
            )
```

packages/valory/skills/decision_maker_abci/behaviours/blacklisting.py
```diff
@@ -45,9 +46,13 @@ def _blacklist(self) -> None:
...
         sampled_bet_index = self.synchronized_data.sampled_bet_index
         sampled_bet = self.bets[sampled_bet_index]
+
         # the question is blacklisted, i.e., we did not place a bet on it,
         # therefore, we decrease the number of bets which was increased on sampling
-        sampled_bet.n_bets -= 1
+        if sampled_bet.queue_status == QueueStatus.TO_PROCESS:
+            sampled_bet.queue_status = QueueStatus.PROCESSED
+        elif sampled_bet.queue_status == QueueStatus.PROCESSED:
+            sampled_bet.queue_status = QueueStatus.REPROCESSED
```
Adamantios, line 55: Let's extract a method for this and reuse it here and in `update_bet_transaction_information`.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -237,8 +253,10 @@ class BetsEncoder(json.JSONEncoder):
 
     def default(self, o: Any) -> Any:
         """The default encoder."""
-        if dataclasses.is_dataclass(o):
+        if dataclasses.is_dataclass(o) and not isinstance(o, type):
```
Adamantios, line 256: Can you explain this?

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -261,6 +279,7 @@ def hook(data: Dict[str, Any]) -> Union[Bet, PredictionResponse, Dict[str, Bet]]
         # if this is a `Bet`
         bet_annotations = sorted(Bet.__annotations__.keys())
         if bet_annotations == data_attributes:
+            data["queue_status"] = QueueStatus(data["queue_status"])
```
Adamantios, line 282: Let's ensure this is compatible with the old json version too, in case someone upgrades but already has some bets stored.

```suggestion
            if "queue_status" in data:
                data["queue_status"] = QueueStatus(data["queue_status"])
```

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -150,20 +153,63 @@ def _update_bets(
...
                 bet.blacklist_forever()
 
-        # truncate the bets, otherwise logs get too big
-        bets_str = str(self.bets)[:MAX_LOG_SIZE]
-        self.context.logger.info(f"Updated bets: {bets_str}")
+    def _bet_freshness_check_and_update(self) -> None:
+        """Check the freshness of the bets."""
+        fresh_statuses = {QueueStatus.FRESH}
+        all_bets_fresh = all(
+            bet.queue_status in fresh_statuses
```
Adamantios, line 176:
```suggestion
        all_bets_fresh = all(
            bet.queue_status is QueueStatus.FRESH
```

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -150,20 +153,63 @@ def _update_bets(
...
+            for bet in self.bets
+            if bet.queue_status is not QueueStatus.EXPIRED
+        )
+
+        if all_bets_fresh:
+            for bet in self.bets:
+                if bet.queue_status is not QueueStatus.EXPIRED:
+                    bet.queue_status = QueueStatus.TO_PROCESS
```
Adamantios, line 184:
```suggestion
        if not all_bets_fresh:
            return

        for bet in self.bets:
            if bet.queue_status is not QueueStatus.EXPIRED:
                bet.queue_status = QueueStatus.TO_PROCESS
```

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -150,20 +153,63 @@ def _update_bets(
...
+
+            # fetch checkpoint status and if reached requeue all bets
+            if self.synchronized_data.is_checkpoint_reached:
+                self._requeue_all_bets()
+
+            # blacklist bets that are older than the opening margin
+            # if trader ran after a long time
+            # helps in resetting the queue number to 0
+            if self.bets:
+                self._blacklist_expired_bets()
```
Adamantios, line 200: Let's move this into `setup()`.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -74,32 +64,83 @@ def processable_bet(self, bet: Bet, now: int) -> bool:
...
-        # rebetting is allowed only if we have already placed at least one bet in this market.
-        # conversely, if we should not rebet, no bets should have been placed in this market.
-        if self.should_rebet ^ bool(bet.n_bets):
-            return False
+        # check if bet queue number is processable
+        processable_statuses = {
+            QueueStatus.TO_PROCESS,
+            QueueStatus.PROCESSED,
+            QueueStatus.REPROCESSED,
+        }
```
Adamantios, line 72: Would this cause an exception here?

https://github.com/valory-xyz/trader/blob/3aa2568368673fbae87da8cf0f2433b9c09676e7/packages/valory/skills/decision_maker_abci/behaviours/base.py#L369-L372

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -74,32 +64,83 @@ def processable_bet(self, bet: Bet, now: int) -> bool:
...
+
+        for bet in bets:
+            if bet.queue_status == QueueStatus.TO_PROCESS:
+                to_process_bets.append(bet)
+            elif bet.queue_status == QueueStatus.PROCESSED:
+                processed_bets.append(bet)
+            elif bet.queue_status == QueueStatus.REPROCESSED:
+                reprocessed_bets.append(bet)
+
+        return to_process_bets, processed_bets, reprocessed_bets
```
Adamantios, line 112:
```suggestion
        bets_by_status = defaultdict(list)
        for bet in bets:
            bets_by_status[bet.queue_status].append(bet)
        return bets_by_status[QueueStatus.TO_PROCESS], bets_by_status[QueueStatus.PROCESSED], bets_by_status[QueueStatus.REPROCESSED]
```

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -114,16 +155,27 @@ def _sample(self) -> Optional[int]:
...
             self.context.logger.warning(msg)
             return None
 
+        # sample a bet using the priority logic
         idx = self._sampled_bet_idx(available_bets)
-        sampled_bet = self.bets[idx]
+        if idx:
+            sampled_bet = self.bets[idx]
+        else:
+            return None
```
Adamantios, line 176: I wonder why `_sampled_bet_idx`'s return value is optional. Can it return `None`?

Adamantios (changes requested):

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -140,27 +140,30 @@ def is_staking_kpi_met(self) -> Generator[None, None, bool]:
...
         # data on the trading branch so that it is available in the
         # cross-period persistent keys.
         if self.is_first_period:
             self.context.logger.debug(f"{self.is_first_period=}")
+            yield  # ensures this is a generator
             return False
 
         self.context.logger.debug(f"{self.params.disable_trading=}")
         if self.params.disable_trading:
+            yield  # ensures this is a generator
```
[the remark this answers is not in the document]
keshav1998: @Adamantios I have added this fix because of the pyright warnings I was getting.
1. Basically the if blocks that were returning True or False directly, essentially was mismatching from the expected return type which was Generator.
2. The return type was expanded from Just Generator to Generator[None, None, bool]:

> The return type `Generator[None, None, bool]` indicates that this function is a generator that yields `None` values and ultimately returns a `bool`. This type annotation is appropriate because:
> 
> - The function uses `yield` to pause execution and produce values.
> - The function's final return value is a boolean (`True` or `False`).
> 
> This ensures that the function's behavior and type signature are consistent with its intended use as a generator.

3. `_compute_stop_trading` method is expected to return a `Generator`, but in the case of the first period check, it returns a `False` directly. This is causing the type mismatch warning from **Pyright**.

To fix this, I ensured that `_compute_stop_trading` always returns a `Generator`. I achieved this by using `yield` in the first period check and disable trading check  as well. 

basically the functionality is still the same but the code aligns much better with return type

packages/valory/skills/market_manager_abci/bets.py
keshav1998: > Pyright: Argument of type "DataclassInstance | type[DataclassInstance]" cannot be assigned to parameter "obj" of type "DataclassInstance" in function "asdict"
>   Type "DataclassInstance | type[DataclassInstance]" is not assignable to type "DataclassInstance"
>     "__dataclass_fields__" is defined as a ClassVar in protocol

@Adamantios  I was getting above pyright issues, basically is_dataclass check would hold true for both dataclass instance and type[DataclassInstance], we wouldn't have faced any issue because during loading we use a bets file, but the check condition was not correct.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
keshav1998: Theoretically and logically both we will never reach this else check, actually we will never reach this check, but have added it for code readability and understanding

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
keshav1998: it can never return None, as `_sample` function only sets `available_bets` list when it's not empty, logically we can remove it

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
keshav1998: @Adamantios Done

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
Adamantios: What confuses me is that we accept markets with reprocessed status but then in `update_bet_transaction_information` we do not.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -262,6 +313,14 @@ def hook(data: Dict[str, Any]) -> Union[Bet, PredictionResponse, Dict[str, Bet]]
         bet_annotations = sorted(Bet.__annotations__.keys())
         if bet_annotations == data_attributes:
             return Bet(**data)
+        else:
+            # fetch missing attributes from the data
+            missing_attributes = set(bet_annotations) - set(data_attributes)
+            new_attributes = {"queue_status", "invested_amount"}
+            if missing_attributes == new_attributes:
+                data["queue_status"] = QueueStatus(0)
+                data["invested_amount"] = 0
```
Adamantios, line 322: I wonder if this is necessary since there are already default values in the dataclass definition.

If I am not missing something, then we could reformat this code to something like:
1. Create constant for new attributes.
2. Calc the sets and the missing attributes.
3. if `bet_annotations` are the same as the `data_attributes` or `missing_attributes` are the `new_attributes` then `return Bet(**data)`

Adamantios: Also, we need to make sure that the users can also roll back to previous releases with the edited `bets.json` files. Therefore we should also add a check on whether the new attributes are present in the file but not in the dataclass and then pop them before decoding.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
keshav1998: Yup I have fixed this, we should accept all actually until kpi has been met, you check the next_status function in bets.py

Adamantios (dismissed):

Adamantios (approved):

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -140,27 +140,30 @@ def is_staking_kpi_met(self) -> Generator[None, None, bool]:
...
 
-        stop_trading_conditions = []
         self.context.logger.debug(f"{self.params.stop_trading_if_staking_kpi_met=}")
         if self.params.stop_trading_if_staking_kpi_met:
             staking_kpi_met = yield from self.is_staking_kpi_met()
             self.context.logger.debug(f"{staking_kpi_met=}")
-            stop_trading_conditions.append(staking_kpi_met)
 
-        return any(stop_trading_conditions)
+            return staking_kpi_met
```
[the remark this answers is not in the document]
jmoreira-valory: The idea of the list is to have an array of boolean conditions that indicate independently if the trader should stop trading, for example:
- staking kpi met
- set stop trading manually
- (... other kpi met, e.g., number of trades per day... [not implemented]... )

If any of these conditions is true, the trader would stop trading.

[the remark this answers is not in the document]
Adamantios: So if I understand correctly it is not necessary at the moment but would be useful in the future if we introduced more conditions. Right, @jmoreira-valory ?

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -34,6 +35,46 @@
...
+        return self
+
+    def next_status(self) -> "QueueStatus":
+        """Get the next status in the queue."""
+        if self == QueueStatus.TO_PROCESS:
+            return QueueStatus.PROCESSED
+        elif self == QueueStatus.PROCESSED:
+            return QueueStatus.REPROCESSED
+        else:
+            return QueueStatus.FRESH
```
annasambrook, line 75: How does a bet move from `FRESH` to `TO_PROCESS`?

annasambrook (dismissed):
Just one question, but other than that LGTM.

annasambrook (dismissed):

packages/valory/skills/market_manager_abci/bets.py
keshav1998: when all the bets, that are present in the list of bets that are FRESH or not EXPIRED.

Adamantios (approved):

Adamantios (approved):

annasambrook (approved):

PR #362, opened 2024-11-28
hotfix found for the payload issue:

cyberosa:

PR #363, opened 2024-11-29
Fix issues introduced by the benchmarking mode changes

Adamantios:
Fixes the issues introduced in #347 and replaces #362.


packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -151,8 +151,8 @@ def async_act(self) -> Generator:
         """Do the action."""
         with self.context.benchmark_tool.measure(self.behaviour_id).local():
             idx = self._sample()
-            benchmarking_finished = False
-            day_increased = False
+            benchmarking_finished = None
+            day_increased = None
```
Adamantios, line 155: This was causing an issue in the **normal flow**.

packages/valory/skills/decision_maker_abci/rounds.py
```diff
@@ -218,6 +219,7 @@ class DecisionMakerAbciApp(AbciApp[Event]):
             Event.DONE: SubscriptionRound,
             Event.NONE: FinishedWithoutDecisionRound,
             Event.NO_MAJORITY: SamplingRound,
+            Event.ROUND_TIMEOUT: SamplingRound,
```
Adamantios, line 222: This was causing an issue in the **normal flow**.

packages/valory/skills/decision_maker_abci/states/sampling.py
```diff
@@ -48,6 +48,7 @@ class SamplingRound(UpdateBetsRound):
         get_name(SynchronizedData.benchmarking_finished),
         get_name(SynchronizedData.simulated_day),
     )
+    synchronized_data_class = SynchronizedData
```
Adamantios, line 51: This was causing an issue in the **normal flow**.

keshav1998 (approved):

cyberosa (approved):

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
cyberosa: Still trying to understand why None is a valid value and False not.

packages/valory/skills/decision_maker_abci/states/sampling.py
Adamantios: Can you elaborate on that?

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
Adamantios: Because otherwise the NONE event cannot ever be emitted from the round.

PR #366, opened 2024-12-06
Fix (CI): Build image from support autonomy version

OjusWiZard:


.github/workflows/release.yaml
```diff
@@ -76,6 +76,7 @@ jobs:
             shell: bash
             run: |
               echo "Building Docker Images"
+              poetry install
```
0xArdi, line 79: is the env being activated below?

OjusWiZard: yes, after this it happens on the line `source env.sh || exit 1`

dagacha (dismissed):

Adamantios (approved):

dagacha (approved):

PR #367, opened 2024-12-06
Update the bets so that we map all the investments per vote

Adamantios:


packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -518,8 +518,6 @@ def _update_selected_bet(
         self.context.logger.info(
             f"with the timestamp:{datetime.fromtimestamp(active_sampled_bet.processed_timestamp)}"
         )
-        if prediction_response is not None:
-            active_sampled_bet.n_bets += 1
```
Adamantios, line 522: @cyberosa this is now automatically calculated from the stored investments. Where should the investments be updated in the benchmarking flow?

keshav1998 (approved):

annasambrook (approved):

PR #372, opened 2024-12-20
Implement a mechanism for detecting and recovering quickly from unreliable tools

Adamantios:


packages/valory/skills/decision_maker_abci/skill.yaml
```diff
@@ -317,6 +317,9 @@ models:
         response_timeout: 300
       agent_balance_threshold: 10000000000000000
       expected_mech_response_time: 300
+      mech_invalid_response: Invalid Response
```
Adamantios, line 320: @0xArdi could you please share the exact, case-sensitive, value that the mech responds with when there is an error, for the `mech_invalid_response` configuration above?

annasambrook (dismissed):

packages/valory/skills/decision_maker_abci/skill.yaml
0xArdi: It varies from tool to tool, some tools will respond differently.

Adamantios: Could you share a list?

keshav1998 (approved):

annasambrook (approved):

PR #378, opened 2025-01-02
new info_utility field in csv Bencharming files

cyberosa:


cyberosa:
the info_utility is not used right now in the logic of the benchmarking, this was meant to add more information but it is not a relevant feature.

PR #382, opened 2025-01-09
Feat/benchmarking convergence

cyberosa:
This branch fixes the Benchmarking mode running at the trader. It introduces some changes regarding

* bets queue mechanism 
* nr mech calls limit support


packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -311,14 +312,12 @@ def hook(data: Dict[str, Any]) -> Union[Bet, PredictionResponse, Dict[str, Bet]]
...
-            if missing_attributes == new_attributes:
-                data["queue_status"] = QueueStatus(0)
-                data["invested_amount"] = 0
-                return Bet(**data)
+        elif "id" in data_attributes:
+            common_attributes = set(bet_annotations) & set(data_attributes)
+            data = {key: data[key] for key in common_attributes}
+            if "queue_status" in data:
+                data["queue_status"] = QueueStatus(data["queue_status"])
+            return Bet(**data)
```
Adamantios, line 320: Can you explain this part?

annasambrook (dismissed):

packages/valory/skills/market_manager_abci/bets.py
annasambrook: Please also add comments to explain this in the code.

packages/valory/skills/decision_maker_abci/behaviours/bet_placement.py
Adamantios, line 1: Please revert. Context:

https://github.com/valory-xyz/trader/blob/bf25df739121080aff6f0fb01adab50699106340/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L977-L986

packages/valory/skills/market_manager_abci/bets.py
keshav1998: @Adamantios it was during the time we had a discussion that bet.json didn't have backward compatibility and all you and Anna were still working on the sell token feature and were going to handle this later. basically the new investments field in Bets dataclass was also being treated as a dataclass instance and files couldn't be read during benchmarking mode.

@annasambrook this can be replaced with the current logic of bets on main, this was a backport fix for then because this branch was created out of sell token feature branch that had investments as a dict with investment values for both yes and no.

I can see that chnage has been reverted so this code is not needed.

packages/valory/skills/decision_maker_abci/behaviours/bet_placement.py
keshav1998: @Adamantios so the reason behind this is we don't go into transaction multiplexer in benchmarking mode and thus the condition in redeem file will never be true.

annasambrook (approved):

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -311,15 +312,16 @@ def hook(data: Dict[str, Any]) -> Union[Bet, PredictionResponse, Dict[str, Bet]]
...
-            if missing_attributes == new_attributes:
-                data["queue_status"] = QueueStatus(0)
-                data["invested_amount"] = 0
-                return Bet(**data)
-
+        # if the data contains an id key, but does not match the bet attributes exactly, process it as a bet
+        # this is necessary for backwards compatibility of the bets.json file
+        elif "id" in data_attributes:
+            # Extract only the attributes that exist in both Bet and data to ensure compatibility
+            common_attributes = set(bet_annotations) & set(data_attributes)
```
Adamantios, line 319:
```suggestion
            common_attributes = bet_annotations.keys() & data_attributes.keys()
```

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -311,15 +312,16 @@ def hook(data: Dict[str, Any]) -> Union[Bet, PredictionResponse, Dict[str, Bet]]
...
+            data = {key: data[key] for key in common_attributes}
+            # Convert queue_status to a QueueStatus enum if present in data
+            if "queue_status" in data:
+                data["queue_status"] = QueueStatus(data["queue_status"])
```
Adamantios, line 323: Would be more preferable to do that when casting:

https://github.com/valory-xyz/trader/blob/19f1fbffc730ec358934e9831899917810cb0fcb/packages/valory/skills/market_manager_abci/bets.py#L196

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -311,15 +312,16 @@ def hook(data: Dict[str, Any]) -> Union[Bet, PredictionResponse, Dict[str, Bet]]
...
-            # fetch missing attributes from the data
-            missing_attributes = set(bet_annotations) - set(data_attributes)
-            new_attributes = {"queue_status", "invested_amount"}
-            if missing_attributes == new_attributes:
-                data["queue_status"] = QueueStatus(0)
-                data["invested_amount"] = 0
-                return Bet(**data)
-
+        # if the data contains an id key, but does not match the bet attributes exactly, process it as a bet
+        # this is necessary for backwards compatibility of the bets.json file
```
Adamantios, line 316: I am not sure if that's the case because we have been changing the filename every time we perform a breaking change.

Adamantios (approved):

packages/valory/skills/market_manager_abci/bets.py
annasambrook: This won't work because `bet_annotations` and `data_attributes` are already a list of keys. They need to be sets so we can compare the overlap of the sets.

packages/valory/skills/market_manager_abci/bets.py
annasambrook: The casting is only run in `__post_init__` which isn't run when we initialise using `Bet(**data)`

packages/valory/skills/market_manager_abci/bets.py
annasambrook: I've removed the line about the backwards compatibility

packages/valory/skills/market_manager_abci/bets.py
Adamantios: That should not be the case. This snippet does not raise an `AssertionError`:

```python
from dataclasses import dataclass


@dataclass
class Dummy:
    test: str
    
    def __post_init__(self):
        self.test = "init called"

a = {"test": "this"}
b = Dummy(**a)
assert b.test != "that", "Init was not called!"
assert b.test == "init called", "Init was not called!"
```

annasambrook (dismissed):

Adamantios (approved):

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -91,7 +99,7 @@ def read_bets(self) -> None:
         """Read the bets from the agent's data dir as JSON."""
         self.bets = []
 
-        if self.shared_state.first_read:
+        if (not self.benchmarking_mode.enabled) and self.shared_state.first_read:
```
Adamantios, line 102:
```suggestion
        if not self.benchmarking_mode.enabled and self.shared_state.first_read:
```

Adamantios (approved):

keshav1998 (approved):

PR #381, opened 2025-01-07
Update trader_pearl service

jmoreira-valory:
Update trader_pearl service

PR #388, opened 2025-01-17
Chore: Update trader_pearl

OjusWiZard:


Adamantios (commented):
This is even closer to what I suggested in https://github.com/valory-xyz/trader/pull/381#pullrequestreview-2536803175. Cannot we just delete this service definition?

jmoreira-valory:
> This is even closer to what I suggested in [#381 (review)](https://github.com/valory-xyz/trader/pull/381#pullrequestreview-2536803175). Cannot we just delete this service definition?

Indeed the idea is to eventually delete it. We are working to make it converge to the 'normal' trader. We can't do it right now due to some technical reasons on Pearl: to update a service we require it has the SAME public ID (this will fail if we suddently replace trader_pearl by trader). We'll investigate how to solve that.

jmoreira-valory:
Closed in favour of #381

jmoreira-valory:
Closed in favour of https://github.com/valory-xyz/trader/pull/381

PR #390, opened 2025-01-21
chore: update tools accuracy hash

annasambrook:
Update the tools accuracy hash


Adamantios (changes requested):
Why update the default values and not simply set an environment variable?

cyberosa (approved):

PR #411, opened 2025-04-16
Fix/pearl epoch change

annasambrook:
Fix for pearl epoch change failing


keshav1998 (dismissed):

packages/valory/services/trader_pearl/service.yaml
```diff
@@ -113,7 +113,7 @@ models:
       redeeming_batch_size: ${REDEEMING_BATCH_SIZE:int:1}
       slippage: ${SLIPPAGE:float:0.01}
       policy_epsilon: ${POLICY_EPSILON:float:0.1}
-      store_path: ${STORE_PATH:str:/data/}
+      store_path: ${STORE_PATH:str:null}
```
Adamantios, line 116: Can you explain this?

Adamantios (dismissed):

packages/valory/services/trader_pearl/service.yaml
keshav1998: From what I can understand, for the Pearl agent the store_path comes as an computed env var(auto calculated for each agent) in the new operator and the path for that depends on where the agent is stored for a run the .operate folder and this path is something like ".operate/../sc-xxx/persistant_data"

cyberosa (approved):
No direct impact of this but we need to test later the benchmarking mode again.

keshav1998 (approved):

Adamantios (approved):

PR #423, opened 2025-05-20
Exp/tool selection randomness

keshav1998:
# Trader-Agent Params Tweaks   🚦

| Parameter (code constant) | **Original value** | **New value** | Why this matters now |
|---------------------------|--------------------|---------------|----------------------|
| `epsilon` (exploration share) | **0.10** | **0.30** | 30 % of calls are random, so every tool retains a healthy daily slice of traffic and no single model can monopolise volume. |
| `VOLUME_FACTOR_REGULARIZATION` | **0.50** | **0.15** | Cuts the historic-volume bonus from **0.50 × share** to **0.15 × share**, ensuring real accuracy—not past traffic—drives rankings. |
| Upper bound of `UNSCALED_WEIGHTED_ACCURACY_INTERVAL` | **100.5** | **80.5** | Shrinks the scaling window to the 47 – 67 % zone where raw scores live, so 1 pp accuracy edges remain visible after rescaling. |
| `TOOL_QUARANTINE_DURATION` | **18 000 s** (5 h) | **10 800 s** (3 h) | A tool that flukes two failures returns 2 h sooner, preventing less count performers from being sidelined most of the day by brief outages. |

---

## 1 · Higher Exploration — `epsilon = 0.30`

Random traffic per day = ε × Total daily calls

* 450 daily-active agents × 61 calls = **27 450 calls/day**  
* Random traffic = 0.30 × 27 450 ≈ **8 235**  
* Random per tool (≈ 12 tools) ≈ **685 calls/day**

Every model—even brand-new or recently quarantined—gets ~685 exploratory calls daily.

---

## 2 · Gentler Volume Bonus — `VOLUME_FACTOR_REGULARIZATION = 0.15`

raw_score = accuracy% + 0.15 × 100 × (calls_i / calls_all)

| Tool (10-day projection) | Traffic share | Bonus with 0.50 | **Bonus with 0.15** |
|--------------------------|---------------|-----------------|---------------------|
| superforcaster           | 17 % | +8.5 pp | **+2.6 pp** |
| prediction-request-rag   | 7 %  | +3.5 pp | **+1.1 pp** |

A 1 pp accuracy edge now overcomes the bonus, so exploit oscillates among the best two-three tools instead of locking onto one.

---

## 3 · Tighter Scaling Window — Cap 80.5

scaled_score = (raw_score + 0.5) / 81

| Tool | Raw score | Scaled (old cap 100.5) | **Scaled (cap 80.5)** |
|------|-----------|------------------------|-----------------------|
[...]

PR #424, opened 2025-05-22
Feat/update tool selection policy

cyberosa:
The current random tool selection logic just picks a random tool from the list. Since we are increasing the exploration of the trader with new tools we want to be sure that we did not end picking more tools with low accuracy so we replaced the random selection by a new logic where the tool is selected from the top five in the list with accuracy > 50%.

This pull request cannot be approved and merged til this previous one is completed
https://github.com/valory-xyz/trader/pull/423


packages/valory/skills/trader_abci/models.py
```diff
@@ -106,7 +106,7 @@
 AccuracyInfoFields = BaseAccuracyInfoFields
 
 
-MARGIN = 5
+MARGIN = 10
```
Adamantios, line 109: Unrelated to the PR.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -33,6 +33,7 @@
 CONFIDENCE_FIELD = "confidence"
 INFO_UTILITY_FIELD = "info_utility"
 BINARY_N_SLOTS = 2
+MECH_TOOL_FIELD = "mech_tool"
```
Adamantios, line 36: Not used here.

packages/valory/skills/decision_maker_abci/behaviours/base.py
Adamantios, line 1: Changes here are unrelated to the PR.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
Adamantios, line 1: Changes here are unrelated to the PR.

packages/valory/skills/decision_maker_abci/policy.py
```diff
@@ -242,7 +256,6 @@ def select_tool(self, randomness: RandomnessType = None) -> Optional[str]:
 
         if not self.has_updated or random.random() < self.eps:  # nosec
             return self.random_tool
-
```
Adamantios, line 245: ?

packages/valory/skills/decision_maker_abci/policy.py
```diff
@@ -163,8 +163,22 @@ def has_updated(self) -> bool:
...
+        ]
+        tools_with_acc.sort(key=lambda x: x[1], reverse=True)
+        tool_names: List[str] = [tool for tool, _ in tools_with_acc]
+
+        if len(tool_names) > 5:
+            tool_names = tool_names[:5]
+
+        if len(tool_names) == 1:
+            return tool_names[0]
+        return random.choice(tool_names)  # nosec
```
Adamantios, line 181: I strongly disagree with this approach. This is no more an e greedy policy and the "exploration" is no more exploration but just a random selection among the top 5 performing tools. The whole purpose of the exploration is to search for new, hopefully better, options. We could either:
1. Implement a decay schedule as suggested in the past.
2. Remove the exploration if we do not want it and simply maintain some sort of greedy policy.

packages/valory/skills/market_manager_abci/bets.py
cyberosa: Fixed

cyberosa:
Closing because we are going to implement the new dynamic pricing policy so we archive it in case we want to go back in the future.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -92,6 +92,7 @@
 QUOTE = '"'
 TWO_QUOTES = '""'
 INIT_LIQUIDITY_INFO = LiquidityInfo()
+MECH_TOOL_FIELD = "mech_tool"
```
Adamantios, line 96: Ideally, it should be configurable as with the other fields.

PR #427, opened 2025-05-30
Replace filtering with `eth_getLogs`

Adamantios:
Replaces filtering operations with simple logs retrieval.

The problem with using `eth_newFilter` is that we are creating filters that are stored in the memory of the RPC node, causing issues on the providers' side.

The `get_logs` method handles this by uninstalling the filter after performing the operation.


Totoual (approved):
LGTM

packages/valory/contracts/conditional_tokens/contract.yaml
```diff
@@ -21,4 +21,8 @@ dependencies:
     version: ==2.28.1
   urllib3:
     version: ==1.26.16
+  eth-utils:
+    version: ==2.2.0
+  eth-abi:
+    version: ==4.0.0
```
Totoual, line 27: Do we need the HexBytes in the dependencies ?

PR #430, opened 2025-06-09
Feat/agent info http

Skanislav:
This PR introduces a new endpoint to the `trader_abci` skill to return information about running agent on endpoint like this. `staking_abci` was adjusted to store and retrieve the `agent_ids` using `service_registry` contract.

```
http://localhost:8716/agent-info

{"address": "0x0133b9F11F8eE185823882d2599E65ceA4c71Cd1", "agent_ids": [25], "service_id": 2181}
```

PR #432, opened 2025-06-11
feat: introduce http endpoint to serve agents' UI

Skanislav:
[First PR in chain](https://github.com/valory-xyz/trader/pull/430)

This PR introduces endpoint for serving the static files using Agent's http handler.

Screenshot 2025-06-16 at 15 59 16


Adamantios (dismissed):

Adamantios (approved):

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -103,25 +119,114 @@ def setup(self) -> None:
...
+                    Path(Path(__file__).parent, self.agent_profile_path, "index.html"),
+                    "r",
+                    encoding="utf-8",
+                ) as file:
+                    index_html = file.read()
+
+                # Send the HTML response
+                self._send_ok_response(http_msg, http_dialogue, index_html)
+        except FileNotFoundError:
+            self._handle_not_found(http_msg, http_dialogue)
```
Adamantios, line 232: I think this was missed. Is the `_handle_not_found` implemented?

Skanislav: It's also never called as file existence is checked here: https://github.com/valory-xyz/trader/blob/feat/sell-tokens-review-bets/packages/valory/skills/trader_abci/handlers.py#L209

And in case of non-existent route we just return index.html

Adamantios: Then why is this try-catch needed here?

Adamantios: Opened #475.

PR #460, opened 2025-07-03
Backend Implementation for ChatUI

LOCKhart07:
This PR adds backend support for switching parameters via prompt input.

**Params supported**
- trading strategy
- mech tool

**Changes**
- Uses genai connection by @dvilelaf 
- ChatUI updates are persisted using a JSON file
- SharedState is not used. Values are directly read from JSON file (Fallback to param from yaml)
   - Edit: SharedState is used now for reading from. Values are written to both json file and sharedState
- Requires GENAI_API_KEY to be set with a valid Gemini API key

**Note**
- Minimum Python version increased to 3.9 due to google-generativeai dependency

### DIFF
--- a/packages/valory/skills/trader_abci/prompts.py
+++ b/packages/valory/skills/trader_abci/prompts.py
@@ -37,8 +37,8 @@ Configuration details:
 - Trading strategy: "{current_trading_strategy}"
     -- Available strategies:
         --- "kelly_criterion_no_conf": Uses the Kelly Criterion formula, a well-known method in finance to optimize profits while maximizing the long-term return on an investment. The AI's predicted probability and market parameters are used to compute the bet amount, which is then adjusted based on the tool's weighted accuracy (higher accuracy increases trust in the suggested amount).
-        --- "bet_amount_per_threshold": A static betting strategy using a mapping from confidence thresholds to fixed bet amounts. For example, with a mapping like {{"0.6": 60000000000000000, "0.7": 90000000000000000, ...}}, higher AI confidence leads to higher bet amounts.
-        --- "mike_strat": Similar to "bet_amount_per_threshold", but the fixed amount from the mapping is multiplied by the AI's confidence (e.g., for confidence 0.6 and mapping value 60000000000000000, the bet is 0.6 * 60000000000000000).
+        --- "bet_amount_per_threshold": A static betting strategy using a mapping from confidence thresholds to fixed bet amounts. For example, with a mapping like {{"0.6": 60000000000000000, "0.7": 90000000000000000, ...}}, higher AI confidence leads to higher bet amounts. Here, 0.6 means 60 percent confidence, and 60000000000000000 means the amount in wei.
+        --- "mike_strat": Similar to "bet_amount_per_threshold", but the fixed amount from the mapping is multiplied by the AI's confidence (e.g., for confidence 0.6 and mapping value 60000000000000000, the bet is 0.6 * 60000000000000000). Again, 0.6 means 60 percent confidence, and 60000000000000000 means wei.
 - Mech tool:
     -- Available tools: {available_tools}
     -- Can be set to \""""
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -17,13 +17,13 @@
         "contract/valory/market_maker/0.1.0": "bafybeiepwclhyg7oo7wcffmben42wmwrxgb4ovoomw2zbjqljcy7b2neke",
         "skill/valory/market_manager_abci/0.1.0": "bafybeihievdcieyf5lad7iamrgkuahxlcjyuqc4kzomi4chcy2pap47ehu",
         "skill/valory/decision_maker_abci/0.1.0": "bafybeifblbnvspgaz4au3gkss5uyqtlnscfglu3mqlmilc2jlum3z6eroq",
-        "skill/valory/trader_abci/0.1.0": "bafybeiho7igjnqr2ejlt3oudtuoxctj5rtwlpi4hhmvcdeuagf44rzgria",
+        "skill/valory/trader_abci/0.1.0": "bafybeibqokdnotwcyradnjijetv4hpcexpvmc5rlz4s75cvvh6uo4v5o3q",
         "skill/valory/tx_settlement_multiplexer_abci/0.1.0": "bafybeibnuuoxhq67a7wocx3fjs3skcvmqfzzczwhrbt4nlmf3rqfcyz7ji",
         "skill/valory/staking_abci/0.1.0": "bafybeifupwkfbxa4c4jogpudvzwt5rkgfuedgd65sj2bf2z5ver4phq64m",
         "skill/valory/check_stop_trading_abci/0.1.0": "bafybeie7jbl24emwbujs34yojjaooyn5niuh3aqagv3nlqg4dg5rkuxyci",
-        "agent/valory/trader/0.1.0": "bafybeibzo6daa4o5doowxkapdutocds4madghwabai636e3l6hgm4bnc5q",
-        "service/valory/trader/0.1.0": "bafybeiddx3xa3plhg476mu25niypok5i736xl2ydtdbhupswkrt42viaky",
-        "service/valory/trader_pearl/0.1.0": "bafybeic7akzsrabybpx3zjokstv2xwokdcmcguytgnpxdqglslndj4v3zi"
+        "agent/valory/trader/0.1.0": "bafybeid4ssq4juy2eetsl7vearbknftkkofg2f2c5ighoj2as2ic5qtlzq",
+        "service/valory/trader/0.1.0": "bafybeibcy25w5pqc2hxx6btmdghzefx4vm7df7zy33c4dwyu2lm3xqzqxe",
+        "service/valory/trader_pearl/0.1.0": "bafybeibeq4nl2mfdvd4rrsysr5usoayijkemjvmwug4gskusa7sabdiw44"
     },
     "third_party": {
         "protocol/valory/acn_data_share/0.1.0": "bafybeih5ydonnvrwvy2ygfqgfabkr47s4yw3uqxztmwyfprulwfsoe7ipq",
--- a/packages/valory/agents/trader/aea-config.yaml
+++ b/packages/valory/agents/trader/aea-config.yaml
@@ -58,7 +58,7 @@ skills:
 - valory/tx_settlement_multiplexer_abci:0.1.0:bafybeibnuuoxhq67a7wocx3fjs3skcvmqfzzczwhrbt4nlmf3rqfcyz7ji
 - valory/market_manager_abci:0.1.0:bafybeihievdcieyf5lad7iamrgkuahxlcjyuqc4kzomi4chcy2pap47ehu
 - valory/decision_maker_abci:0.1.0:bafybeifblbnvspgaz4au3gkss5uyqtlnscfglu3mqlmilc2jlum3z6eroq
-- valory/trader_abci:0.1.0:bafybeiho7igjnqr2ejlt3oudtuoxctj5rtwlpi4hhmvcdeuagf44rzgria
+- valory/trader_abci:0.1.0:bafybeibqokdnotwcyradnjijetv4hpcexpvmc5rlz4s75cvvh6uo4v5o3q
 - valory/staking_abci:0.1.0:bafybeifupwkfbxa4c4jogpudvzwt5rkgfuedgd65sj2bf2z5ver4phq64m
 - valory/check_stop_trading_abci:0.1.0:bafybeie7jbl24emwbujs34yojjaooyn5niuh3aqagv3nlqg4dg5rkuxyci
 - valory/mech_interact_abci:0.1.0:bafybeigkizx3ecohtyshr6tq55zvp4sl5cou344a2t6nvwfgund4xnrj4a
--- a/packages/valory/services/trader/service.yaml
+++ b/packages/valory/services/trader/service.yaml
@@ -7,7 +7,7 @@ license: Apache-2.0
 fingerprint:
   README.md: bafybeigtuothskwyvrhfosps2bu6suauycolj67dpuxqvnicdrdu7yhtvq
 fingerprint_ignore_patterns: []
-agent: valory/trader:0.1.0:bafybeibzo6daa4o5doowxkapdutocds4madghwabai636e3l6hgm4bnc5q
+agent: valory/trader:0.1.0:bafybeid4ssq4juy2eetsl7vearbknftkkofg2f2c5ighoj2as2ic5qtlzq
 number_of_agents: 4
 deployment:
   agent:
--- a/packages/valory/services/trader_pearl/service.yaml
+++ b/packages/valory/services/trader_pearl/service.yaml
@@ -8,7 +8,7 @@ license: Apache-2.0
 fingerprint:
   README.md: bafybeibg7bdqpioh4lmvknw3ygnllfku32oca4eq5pqtvdrdsgw6buko7e
 fingerprint_ignore_patterns: []
-agent: valory/trader:0.1.0:bafybeibzo6daa4o5doowxkapdutocds4madghwabai636e3l6hgm4bnc5q
+agent: valory/trader:0.1.0:bafybeid4ssq4juy2eetsl7vearbknftkkofg2f2c5ighoj2as2ic5qtlzq
 number_of_agents: 1
 deployment:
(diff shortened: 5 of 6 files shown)


packages/dvilela/connections/genai/connection.py
```diff
@@ -0,0 +1,241 @@
...
+
+        self.put_envelope(response_envelope)
+
+    def _get_response(self, payload: dict) -> Tuple[Dict, bool]:
+        """Get response from Genai."""
+
+        AVAILABLE_MODELS = [
+            "gemini-1.5-flash",
+            "gemini-1.5-pro",
+            "gemini-2.0-flash-exp",
```
cyberosa, line 177: why not gemini-2.5-flash? which is one of the latest

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -313,6 +314,79 @@ def _mock_balance_check(self) -> None:
         self.wallet_balance = self.benchmarking_mode.native_balance
         self._report_balance()
 
+    def _store_chatui_param(self, param_name: str, value: Any) -> None:
+        """Store chatui param to JSON file."""
+        chatui_store_path = self.params.store_path / CHATUI_PARAM_STORE
+        # If the file exists, load and update it; otherwise, create a new store
+        if os.path.exists(chatui_store_path):
+            with open(chatui_store_path, "r") as f:
+                current_store: dict = json.load(f)
```
cyberosa, line 323: maybe some try/except here in case there are issues with the storage process

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -172,6 +196,155 @@ def _send_ok_response(
...
+        :param http_dialogue: the original HttpDialogue
+        """
+        self.context.logger.info(
+            f"LLM response payload: {llm_response_message.payload}"
+        )
+
+        llm_response = json.loads(llm_response_message.payload).get("response", "{}")
+        llm_response_json = json.loads(llm_response)
+
+        updated_trading_strategy: str = llm_response_json.get("trading_strategy", None)
```
cyberosa, line 288: The trading strategy cannot be just any string. There are only 3 options available so we should check that the value is one from the valid options otherwise to throw an error.

LOCKhart07: I'm using structured output calling with gemini where I'm passing it an enum that has the 3 strategy strings. So in practice, the model should never return anything outside of those.

That said, I can add an explicit check to validate the response just to make sure it’s never a concern.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -64,6 +81,15 @@
...
 PREDICT_AGENT_PROFILE_PATH = "predict-ui-build"
+CHATUI_PARAM_STORE = "chatui_param_store.json"
+
+AVAILABLE_TRADING_STRATEGIES = frozenset(
+    {
+        "kelly_criterion_no_conf",
+        "bet_amount_per_threshold",
+        "mike_strat",
+    }
+)
```
Adamantios, line 92: You can reuse the existing enum instead.

```python
from packages.valory.skills.trader_abci.prompts import TradingStrategy
```

```suggestion
AVAILABLE_TRADING_STRATEGIES = frozenset(strategy.value for strategy in TradingStrategy)
```

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -172,6 +204,170 @@ def _send_ok_response(
...
+        Send a message and set up a callback for the response.
+
+        :param message: the Message to send
+        :param dialogue: the Dialogue context
+        :param callback: the callback function upon response
+        :param callback_kwargs: optional kwargs for the callback
+        """
+        self.context.outbox.put_message(message=message)
+        nonce = dialogue.dialogue_label.dialogue_reference[0]
+        self.context.state.req_to_callback[nonce] = (callback, callback_kwargs or {})
```
Adamantios, line 224: Why inconsistent types?

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -172,6 +204,170 @@ def _send_ok_response(
...
+                        f"Unsupported trading strategy: {updated_trading_strategy}"
+                    )
+                else:
+                    # Store the trading strategy in the chatui param store
+                    self._store_chatui_param(
+                        "trading_strategy", updated_trading_strategy
+                    )
+                    updated_params.update(
+                        {"trading_strategy": updated_trading_strategy}
+                    )
```
Adamantios, line 316: Use early return instead to avoid too much nesting and make the code more readable.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -172,6 +204,170 @@ def _send_ok_response(
...
+        updated_params = {}
+        if updated_agent_config:
+            updated_trading_strategy: str = updated_agent_config.get(
+                "trading_strategy", None
+            )
+            if updated_trading_strategy:
+                if updated_trading_strategy not in AVAILABLE_TRADING_STRATEGIES:
+                    self.context.logger.error(
+                        f"Unsupported trading strategy: {updated_trading_strategy}"
+                    )
```
Adamantios, line 308: Why send ok response in this case?

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -172,6 +204,170 @@ def _send_ok_response(
...
+            http_dialogue,
+            {
+                "updated_params": updated_params,
+                "llm_response": llm_response_json,
+                "llm_message": llm_message,
+                "message": (
+                    "Params successfully updated"
+                    if updated_params
+                    else "No Params updated"
+                ),
```
Adamantios, line 329: Why send an ok response in this case?

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -172,6 +204,170 @@ def _send_ok_response(
...
+    def _store_chatui_param(self, param_name: str, value: Any) -> None:
+        """Store chatui param."""
+        chatui_store_path = self.context.params.store_path / CHATUI_PARAM_STORE
+        # If the file exists, load and update it; otherwise, create a new store
+        if os.path.exists(chatui_store_path):
+            with open(chatui_store_path, "r") as f:
+                current_store: dict = json.load(f)
+
+        else:
+            current_store = {}
```
Adamantios, line 342: Logic is duplicated in the method below.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -172,6 +204,170 @@ def _send_ok_response(
...
+            current_store: dict = json.load(f)
+
+        return current_store.get(param_name, None)
+
+    def get_trading_strategy(self) -> str:
+        """Get the trading strategy."""
+        trading_strategy = self._load_chatui_param("trading_strategy")
+        if trading_strategy is None:
+            trading_strategy = self.context.params.get("trading_strategy", None)
+        return trading_strategy
```
Adamantios, line 365:
```suggestion
        return self._load_chatui_param("trading_strategy") or self.context.params.get("trading_strategy")
```

packages/valory/skills/trader_abci/handlers.py
Adamantios, line 1: Missing constants, e.g., `trading_strategy`, `response`, `updated_agent_config`, etc.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -230,3 +426,41 @@ def _handle_get_static_file(
...
+        nonce = srr_msg.dialogue_reference[
+            0
+        ]  # Assuming dialogue_reference is accessible
+        callback, kwargs = self.context.state.req_to_callback.pop(nonce, (None, {}))
+
+        if callback is None:
+            super().handle(message)
+        else:
+            dialogue = self.context.srr_dialogues.update(srr_msg)
+            callback(srr_msg, dialogue, **kwargs)
```
Adamantios, line 466:
```suggestion
            return
        dialogue = self.context.srr_dialogues.update(srr_msg)
        callback(srr_msg, dialogue, **kwargs)
```

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -313,6 +314,83 @@ def _mock_balance_check(self) -> None:
...
+        try:
+            with open(param_path, "r") as f:
+                current_store: dict = json.load(f)
+
+            return current_store.get(param_name, None)
+        except Exception:
+            self.context.logger.info(
+                "ChatUI param store file not found or is invalid, returning None."
+            )
+            return None
```
Adamantios, line 349: Duplicates some logic of the above method.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -313,6 +314,83 @@ def _mock_balance_check(self) -> None:
...
+        if the parameter value in the YAML configuration has changed since it
+        was last stored.
+
+        :param param_name: The name of the parameter to fetch.
+        :return: The value of the parameter.
+        """
+        config_value = getattr(self.params, param_name)
+
+        param_json = self._load_chatui_param(param_name)
+        initial_json = self._load_chatui_param(f"initial_{param_name}")
```
Adamantios, line 366: Could use a `list` to hold history instead.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -313,6 +314,83 @@ def _mock_balance_check(self) -> None:
...
+        if param_json is None:
+            self._store_chatui_param(param_name, config_value)
+            param_json = config_value
+
+        # If the configured value does not match the initial value in the db,
+        # the user has reconfigured it and we need to update it:
+        if config_value != initial_json:
+            self._store_chatui_param(param_name, config_value)
+            self._store_chatui_param(f"initial_{param_name}", config_value)
+            initial_json = config_value
```
Adamantios, line 383: What is the effect of storing the `config_value` to the `initial_json`?

Adamantios (changes requested):
Using strings to retrieve the chatui params is unsafe.

Instead, some basic validation should be performed after reading, and the params should be set in some form of structure.

pyproject.toml
```diff
@@ -81,6 +81,7 @@ jsonschema = "<4.4.0,>=4.3.0"
 openapi-core = "==0.15.0"
 openapi-spec-validator = "<0.5.0,>=0.4.0"
 pyinstaller = "==6.8.0"
+google-generativeai = "==0.8.2"
```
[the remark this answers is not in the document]
LOCKhart07: Yes, your assumption is correct.

After discussing with @dvilelaf, decided against using the mech because of its response latency, where sometimes it could take around 20 minutes to respond. This delay would lead to poor UX for ChatUI feature, hence opted for going with the 'connection' approach even though it requires a API key(free).

packages/valory/skills/trader_abci/models.py
```diff
@@ -173,3 +173,5 @@ def setup(self) -> None:
 
         for event, override in event_to_timeout_overrides.items():
             TraderAbciApp.event_to_timeout[event] = override
+
+        self.req_to_callback: Dict[str, Callable[..., Any]] = {}
```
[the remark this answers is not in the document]
LOCKhart07: Used here

[packages/valory/skills/trader_abci/handlers.py:L224](https://github.com/valory-xyz/trader/pull/460/files#diff-7d7fbe10a9275c5857cf06e8e7ae8b4448c235b2468b40a50bcec2e3d9a6a9c2R223)
`        self.context.state.req_to_callback[nonce] = (callback, callback_kwargs or {})`


If you mean it should be defined somewhere else, please let me know and I'll make the changes.

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: It could've been that the user just asked a question, eg. "What are the available trading strategies?". This would be fullfilled by llm_message key in response

But you're point stands. I'm removing the 'message' key and adding a issues key(list of strings) which will carry problems like "unsupported trading strategy."

packages/valory/skills/decision_maker_abci/behaviours/base.py
LOCKhart07: No actual effect, apart from making sure the variables reflect their actual source atp. 
That said, while revisiting this, I found some real issues with the logic, so I’ve reworked it a bit.

packages/valory/skills/decision_maker_abci/models.py
```diff
@@ -387,11 +400,13 @@ def _ensure_chatui_store(self) -> Any:
...
 
-        with open(chatui_store_path, "w") as f:
-            json.dump(current_store, f, indent=4)
+        self._set_json_store(current_store)
 
         self._chat_ui_params = ChatUIParams(**current_store)
 
+        if self._chat_ui_params is None:
+            raise ValueError("The chat UI parameters have not been set!")
+
```
Adamantios, line 409: Context: https://github.com/valory-xyz/trader/pull/460#discussion_r2195508353

I noticed just now that this is not possible, right? I also noticed that you kept the check for typing reasons in the `chat_ui_params` property. Therefore, let's remove this check from here as it is unnecessary.

packages/valory/skills/decision_maker_abci/models.py
```diff
@@ -323,6 +346,52 @@ def setup(self) -> None:
...
+        chatui_store_path = self.context.params.store_path / CHATUI_PARAM_STORE
+        try:
+            if os.path.exists(chatui_store_path):
+                with open(chatui_store_path, "r") as f:
+                    current_store: dict = json.load(f)
+            else:
+                current_store = {}
+        except (FileNotFoundError, json.JSONDecodeError):
+            current_store = {}
```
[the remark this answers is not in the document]
Adamantios: Second comment was missed.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -91,6 +91,7 @@
 NEW_LINE = "\n"
 QUOTE = '"'
 TWO_QUOTES = '""'
+CHATUI_PARAM_STORE = "chatui_param_store.json"
```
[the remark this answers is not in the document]
LOCKhart07: Would it be reasonable to do this refactor as part of the migration to a seperate skill?

packages/valory/skills/decision_maker_abci/models.py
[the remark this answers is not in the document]
Adamantios: What you said is correct. However, this is not a component; this is in the shared state.

Adamantios (approved):

packages/dvilela/connections/genai/connection.py
LOCKhart07: https://github.com/valory-xyz/trader/issues/468

Adamantios (changes requested):
I do not agree with #479.

LOCKhart07:
> I do not agree with #479. We should never move third-party packages to dev.

https://github.com/valory-xyz/trader/issues/482

packages/valory/skills/trader_abci/prompts.py
```diff
@@ -0,0 +1,84 @@
...
+
+CHATUI_PROMPT = (  # nosec
+    """
+You are an expert assistant tasked with helping users update an agent's trading configuration.
+
+Configuration details:
+- Trading strategy: "{current_trading_strategy}"
+    -- Available strategies:
+        --- "kelly_criterion_no_conf": Uses the Kelly Criterion formula, a well-known method in finance to optimize profits while maximizing the long-term return on an investment. The AI's predicted probability and market parameters are used to compute the bet amount, which is then adjusted based on the tool's weighted accuracy (higher accuracy increases trust in the suggested amount).
+        --- "bet_amount_per_threshold": A static betting strategy using a mapping from confidence thresholds to fixed bet amounts. For example, with a mapping like {{"0.6": 60000000000000000, "0.7": 90000000000000000, ...}}, higher AI confidence leads to higher bet amounts.
```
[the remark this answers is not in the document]
LOCKhart07: Just to confirm, you mean as in '0.6' here is 60 percent confidence and 60000000000000000 is WEI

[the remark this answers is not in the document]
LOCKhart07: Added in c6c5c851b2949b6b4e09f445f66034fa9a07b1a7

cyberosa (approved):

PR #453, opened 2025-06-30
feat(sell tokens 2): schedule reviews for selling

Skanislav:
First PR in the chain: #437 

This PR aims to introduce logic for scheduling periodic market processing (after KPI is met) and triggering a sell if the prediction has changed.

PR #437, opened 2025-06-17
feat(sell tokens 1): behaviour to sell bets

Skanislav:
The first PR in the chain. 

Aims to add tools for selling the bets with no knowledge of how the selling is triggered. Heavily inspired by #350 but with addressed comments and some fixed bugs.

PR #487, opened 2025-07-18
fix(sell tokens 3): fixes for CI 

Skanislav:
First PR in chain: https://github.com/valory-xyz/trader/pull/437
Second PR in chain: https://github.com/valory-xyz/trader/pull/453

This PR only carries the fixes for passing the CI to reduce the scope of PRs above. 

It will also contain the final hashes for testing

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -16,14 +16,14 @@
         "contract/valory/relayer/0.1.0": "bafybeihypuljybkocl6iiacy52py7i5iqxdli3ily66q7b3nego4qajvne",
         "contract/valory/market_maker/0.1.0": "bafybeigzfyzbixum5cqvcbyc24dqwce25dr6nyqvgjm2eqpt6tgo2perd4",
         "skill/valory/market_manager_abci/0.1.0": "bafybeigzldcghoxbg73e53ha2v3cayshswrrgeg4bqpcwd6al237v2ssle",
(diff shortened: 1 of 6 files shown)


packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -110,7 +110,11 @@ def should_review_bets(self, is_staking_kpi_met: bool) -> bool:
             return False
 
         last_review = self.synchronized_data.get_last_review_timestamp()  # type: ignore
-        
+
+        if not self.context.params.review_period_seconds:
+            self.context.logger.info("Review period is not set")
+            return False
```
Adamantios, line 116: Generally, it is a good practice to enforce the configs to be set by using the `_ensure` method in the models when looking for the configs' values in the kwargs.

Skanislav: Done here: https://github.com/valory-xyz/trader/pull/487/commits/a668d138daf3253687d5508577439e156e83effb

Adamantios: Perfect, so now we do not need this check.

Skanislav: Removed here: https://github.com/valory-xyz/trader/pull/487/commits/4061ec086db8e7f90eac79c271da7e5ba567c2c9

LOCKhart07 (approved):

Issue #472, opened 2025-07-14
Refactor: Change content types to an enum


LOCKhart07:
context: https://github.com/valory-xyz/trader/pull/471#discussion_r2205360334

PR #497, opened 2025-07-25
Refactor: Change content types to an enum

LOCKhart07:
#472


packages/valory/skills/trader_abci/composition.py
```diff
@@ -124,6 +126,7 @@
     FinishedMarketManagerRound: CheckStopTradingRound,
     FinishedCheckStopTradingRound: RandomnessRound,
     FinishedWithSkipTradingRound: RedeemRound,
+    FinishedWithReviewBetsRound: RandomnessRound,
```
Skanislav, line 129: https://github.com/valory-xyz/trader/blob/develop/packages/valory/skills/trader_abci/composition.py#L123

these changes are already in master. Something shady going on

Skanislav (approved):

packages/valory/skills/trader_abci/composition.py
Adamantios, line 1: These changes should not be present. Is your base out of sync?

packages/valory/skills/chatui_abci/handlers.py
```diff
@@ -84,15 +84,28 @@ class HttpMethod(Enum):
...
+class HttpContentType(Enum):
+    """Enum for HTTP content types."""
+
+    HTML = "Content-Type: text/html\n"
+    JS = "Content-Type: application/javascript\n"
+    JSON = "Content-Type: application/json\n"
+    CSS = "Content-Type: text/css\n"
+    PNG = "Content-Type: image/png\n"
+    JPG = "Content-Type: image/jpeg\n"
+    JPEG = "Content-Type: image/jpeg\n"
```
Adamantios, line 96: We could avoid some duplication.

```suggestion
class HttpContentType(Enum):
    """Enum for HTTP content types."""
    HTML = "text/html"
    JS = "application/javascript"
    JSON = "application/json"
    CSS = "text/css"
    PNG = "image/png"
    JPG = "image/jpeg"
    JPEG = "image/jpeg"

    @property
    def header(self) -> str:
        return f"Content-Type: {self.value}\n"
```

Then, instead of `value`, we could use `header`.

Adamantios (changes requested):

packages/valory/skills/trader_abci/composition.py
LOCKhart07: Yes. But only the last PR would be merged to develop.
@Skanislav Could you help with this

packages/valory/skills/chatui_abci/handlers.py
LOCKhart07: e1c2674b4d7f5319602b25e0e2a8f3a0ab4a7d4c

PR #504, opened 2025-08-04
fix: agent doesn't crash if policy is not set

Skanislav:
This fix should help with handling the cases when the blacklisting round was called, and for some reason, the policy was not set.


packages/valory/skills/decision_maker_abci/behaviours/blacklisting.py
```diff
@@ -66,6 +66,14 @@ def async_act(self) -> Generator:
...
 
         with self.context.benchmark_tool.measure(self.behaviour_id).local():
+            if not self.policy:
+                success = yield from self._setup_policy_and_tools()
+                if not success:
+                    self.context.logger.info(
+                        "Tool selection failed, skipping blacklisting"
+                    )
+                    return
+
```
Adamantios, line 76: The check is redundant as it's performed in the `_setup_policy_and_tools`.

```suggestion
            success = yield from self._setup_policy_and_tools()
            if not success:
                self.context.logger.info(
                    "Tool selection failed, skipping blacklisting"
                )
                return

```

Adamantios (approved):

PR #505, opened 2025-08-04
feat(sell-tokens 3): logic needed for benchmarking mode 

Skanislav:
Last PR in the chain which changes related to benchmarking mode for selling feature


packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -569,9 +569,30 @@ def should_sell_outcome_tokens(
...
+                if question_id not in bets_mapping:
+                    bets_mapping[question_id] = []
+                bets_mapping[question_id].append(row_number)
+        return bets_mapping
+
     def async_act(self) -> Generator:
         """Do the action."""
 
+        if self.benchmarking_mode.enabled:
+            if len(self.shared_state.bet_id_row_manager) == 0:
```
cyberosa, line 592: The initialization happens on the decision_request state, so why do we need to initialize it here again? We cannot go into decision receive without going first into decision_request.

cyberosa: https://github.com/valory-xyz/trader/blob/main/packages/valory/skills/decision_maker_abci/behaviours/decision_request.py#L105

Skanislav: Yes, you are right. I think I copied it from bet placement behavior

LOCKhart07 (approved):

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
Skanislav: Removed here: 86ef4f11

cyberosa (approved):

packages/valory/skills/decision_maker_abci/behaviours/sell_outcome_tokens.py
```diff
@@ -133,18 +133,21 @@ def async_act(self) -> Generator:
...
             if self.synchronized_data.vote is None:
-                raise ValueError("Vote is not set")
+                raise ValueError("The round was called with no vote, nothing to sell. Most likely it's a bug.")
 
             # we flip the vote to the opposite one to buy the outcome token
-            opposite_vote = self.sampled_bet.opposite_vote(self.outcome_index)
+            outcome_index = self.sampled_bet.opposite_vote(self.outcome_index)
+
+            if self.sell_amount:
+                sell_amount = self.sell_amount
```
Adamantios, line 142:
```suggestion
            if self.sell_amount:
                sell_amount = self.sell_amount
                outcome_index = self.sampled_bet.opposite_vote(self.outcome_index)
```

packages/valory/skills/market_manager_abci/behaviours.py
```diff
@@ -222,7 +222,7 @@ def get_active_bets(self) -> Generator[None, None, Dict[str, Dict[str, int]]]:
             self.synchronized_data.safe_contract_address.lower()
         )
         if user_positions is None:
-            return {}
+            return []
```
Adamantios, line 225: Does not match the specified return type:

https://github.com/valory-xyz/trader/blob/f100392287ed1d58f9529aa59ffea23a260b313c/packages/valory/skills/market_manager_abci/behaviours.py#L213

Adamantios (approved):

PR #507, opened 2025-08-08
chore: update hashes to use the right ABI for mech-marketplace

Skanislav:
```
        "contract/valory/mech_marketplace/0.1.0": "bafybeibrojfogyvfv5pvcozjvjks7hfhsqmxzuod7cdkkqzybv376bpysu",
        "skill/valory/mech_interact_abci/0.1.0": "bafybeibxvbgtj5pi24e2fh7kskdcv72buevqzaykw4lirkr4n7qn5ba2re"
```

To build this version we had to move both to the `dev` packages, apply the change locally, lock hashes, push them to IPFS and then return packages back to the third_party. As we had some problems with using hash directly. 

This version contains a variant of the fix which uses `requestData` and not `data` here. 
https://github.com/valory-xyz/mech-interact/blob/main/packages/valory/contracts/mech_marketplace/contract.py#L197


dvilelaf:
We can release this as a hotfix, but the proper way to do it is:
* Make the fix in the source repo: https://github.com/valory-xyz/mech-interact (it was extracted from IEKit recently into its own repo)
* Import the package from IPFS in all target repos

LOCKhart07 (approved):
LGTM

dvilelaf (approved):

dagacha (approved):
Tested - LGTM

PR #522, opened 2025-08-21
updating irrelevant tools list

cyberosa:
After latest accuracy computation we found there are two tools not performing well so we are going to remove them from the options available to the traders.
The tools are: 
prediction-offline-sme 
prediction-url-cot-claude 
prediction-url-cot

### DIFF
--- a/packages/valory/services/trader_pearl/service.yaml
+++ b/packages/valory/services/trader_pearl/service.yaml
@@ -8,7 +8,7 @@ license: Apache-2.0
 fingerprint:
   README.md: bafybeibg7bdqpioh4lmvknw3ygnllfku32oca4eq5pqtvdrdsgw6buko7e
 fingerprint_ignore_patterns: []
-agent: valory/trader:0.1.0:bafybeig2ued5fzjpxf5g5girkyrxe3mki7ynjto4u3aivefqf4bhc7qkg4
+agent: valory/trader:0.1.0:bafybeibdi7oq27h4ubor3kasn3yre3lijz4xlahdhudakcy5upvtjgktzi
 number_of_agents: 1
 deployment:
   agent:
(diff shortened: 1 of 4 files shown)


packages/valory/services/trader_pearl/service.yaml
```diff
@@ -116,7 +116,7 @@ models:
...
       store_path: ${STORE_PATH:str:/data/}
       irrelevant_tools: ${IRRELEVANT_TOOLS:list:["native-transfer","prediction-online-lite","claude-prediction-online-lite","prediction-online-sme-lite","prediction-request-reasoning-lite","prediction-request-reasoning-claude-lite","prediction-offline-sme","deepmind-optimization",
-        "deepmind-optimization-strong","openai-gpt-3.5-turbo","openai-gpt-3.5-turbo-instruct","openai-gpt-4","openai-text-davinci-002","openai-text-davinci-003","prediction-online-sum-url-content","prediction-online-summarized-info","stabilityai-stable-diffusion-512-v2-1","stabilityai-stable-diffusion-768-v2-1","stabilityai-stable-diffusion-v1-5","stabilityai-stable-diffusion-xl-beta-v2-2-2"]}
+        "deepmind-optimization-strong","openai-gpt-3.5-turbo","openai-gpt-3.5-turbo-instruct","openai-gpt-4","openai-text-davinci-002","openai-text-davinci-003","prediction-online-sum-url-content","prediction-online-summarized-info","stabilityai-stable-diffusion-512-v2-1","stabilityai-stable-diffusion-768-v2-1","stabilityai-stable-diffusion-v1-5","stabilityai-stable-diffusion-xl-beta-v2-2-2","prediction-url-cot-claude","prediction-url-cot"]}
```
LOCKhart07, line 119: `prediction-offline-sme` is missing here. Is that on purpose?

cyberosa: no, it was already removed long time ago. Check again the first line and you will find it :)

Adamantios (commented):
Why update the defaults and not the env vars? This is why these are configurable so that we do not need to unnecessarily create new hashes and push them on IPFS.

PR #530, opened 2025-08-28
fix(selling-1): sell-tokens approval

Skanislav:
This PR fixes issue with approval required to allow sell outcome tokens. Initially approval was granted on the market contract but it suppose to be conditional tokens approved to be sold. 

PRs in chain:
- https://github.com/valory-xyz/trader/pull/537
- https://github.com/valory-xyz/trader/pull/648

Remaining work:

- This PR fixes issue with approval
- Next one implements required step to initiate selling from bet sampling
- Sell amount calculation #648 needs to be used instead of `calc_sell_amount`


packages/valory/skills/decision_maker_abci/states/decision_receive.py
```diff
@@ -94,13 +94,12 @@ def end_block(self) -> Optional[Tuple[SynchronizedData, Enum]]:
         if event == Event.DONE and synced_data.vote is None:
             return synced_data, Event.TIE
 
-        self.context.logger.info(f"Should be sold. {synced_data=}")
```
Skanislav, line 97: This one would not be missed

PR #532, opened 2025-09-01
fix(1): weighted accuracy weights

Skanislav:
This PR introduced a small change related to the weight of selected tools. The new accuracy hash contains best best-picked tools, and existing user history might affect a tool selection depending on the historical amount of requests. 

In this PR, I changed the value from `0.25` to `0.1`, decreasing the weight by almost 2.5 times. 
It resulted in the expected behaviour, which is described in the `test_e_greedy_policy_select_tool_weighted_accuracy_calls_decreased ` test.

For a situation like this, the selected tool would still be `prediction-request-reasoning`, as it has the higher accuracy. If we change back the `VOLUME_FACTOR_REGULARIZATION =0.25` then the `prediction-request-rag` would be selected

```
e_greedy_policy_mock.accuracy_store.update(
        {
            "prediction-request-rag": AccuracyInfo(
                accuracy=0.57965, pending=-6, requests=5021
            ),
            "prediction-request-reasoning": AccuracyInfo(
                accuracy=0.66411, pending=-6, requests=521
            ),
        }
```

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -16,14 +16,14 @@
         "contract/valory/relayer/0.1.0": "bafybeihypuljybkocl6iiacy52py7i5iqxdli3ily66q7b3nego4qajvne",
         "contract/valory/market_maker/0.1.0": "bafybeigzfyzbixum5cqvcbyc24dqwce25dr6nyqvgjm2eqpt6tgo2perd4",
         "skill/valory/market_manager_abci/0.1.0": "bafybeif3zwrjq6suryleltxd3vkta5zeiczyrzpgsbwjlpo26g2gfxf4iq",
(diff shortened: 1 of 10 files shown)


Adamantios (approved):
LGTM, how did we decide `0.1` instead of `0`?

Skanislav:
> LGTM, how did we decide `0.1` instead of `0`?

Heuristics and binary search to find a non-zero value that would still slightly affect the weight. Have no problem with changing it to 0. I'm a bit afraid completely disabling this logic

LOCKhart07 (dismissed):

Adamantios (approved):

LOCKhart07 (approved):

PR #539, opened 2025-09-03
sync develop with current main

solarw:


Skanislav:
What will we do with https://github.com/valory-xyz/trader/pull/538/files PR?

PR #541, opened 2025-09-03
Remove the obsolete NVM subscription logic

Adamantios:
This is now replaced by the latest mech-interact version. 
See also #526.


Adamantios:
> I notice this build is missing few commits from develop and the previous version. Might need a merge or rebase. Otherwise tested agent locally, successfully making requests and placing bets

Will `develop` sync on the base branch after merging.

PR #549, opened 2025-09-09
feat: added workflow for pushing hashes from RCs

LOCKhart07:
Adds a GH action to push all images to IPFS whenever a release PR is opened

Actual `push-packages` step is copied from release workflow

### DIFF
--- a/.github/workflows/release_candidate.yaml
+++ b/.github/workflows/release_candidate.yaml
@@ -1,0 +1,32 @@
+name: Release Candidate Flow
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    branches:
+      - main
+
+jobs:
+  publish-packages:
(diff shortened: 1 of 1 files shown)


.github/workflows/release_candidate.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    branches:
+      - main
+
+jobs:
+  publish-packages:
+    if: github.head_ref =~ '^v[0-9]+\.[0-9]+\.[0-9]+$'
```
Adamantios, line 11: Needs to alos end with `-rc{number}`.

LOCKhart07: I checked the previous release branches and they don't have the rc thing
https://github.com/valory-xyz/trader/pull/542
https://github.com/valory-xyz/trader/pull/534

PR #561, opened 2025-09-10
Fix/chatui prompt fixes

LOCKhart07:
- Adds html support to LLM Reasoning
- Mike strat has been temporarily removed, so the user will not be able to select it
- Added information about the UI strategy names to chatUI prompts

https://linear.app/valory-xyz/issue/PREDICT-337
https://linear.app/valory-xyz/issue/PREDICT-338

image


Skanislav (approved):
🙏

Adamantios (approved):

PR #568, opened 2025-09-12
APT - PR comments

LOCKhart07:


packages/valory/skills/agent_performance_summary_abci/models.py
```diff
@@ -89,7 +89,7 @@ def _read_existing_performance_summary(self) -> Optional[AgentPerformanceSummary
             self.context.logger.warning(
                 f"Could not read existing agent performance summary: {e}"
             )
-            return None
+            return AgentPerformanceSummary()
```
Adamantios, line 92: This means the method does not return `Optional` anymore:

https://github.com/valory-xyz/trader/blob/02b9e61ad0a3ae2f0cf2d3647bc15283b2bf9d1d/packages/valory/skills/agent_performance_summary_abci/models.py#L80

PR #585, opened 2025-10-02
1. Funding mechanism: Integrated funds manager skill 

LOCKhart07:
PR chain:
1. https://github.com/valory-xyz/trader/pull/585
2. https://github.com/valory-xyz/trader/pull/586
3. https://github.com/valory-xyz/trader/pull/587
4. https://github.com/valory-xyz/trader/pull/591


Skanislav (approved):
Integration looks okay if providing rpcs as parameter is a subject to change

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -96,6 +98,11 @@ def agent_ids(self) -> List[int]:
         """Get the agent ids."""
         return json.loads(self.staking_synchronized_data.agent_ids)
 
+    @property
+    def funds_status(self) -> FundRequirements:
+        """Get the fund status."""
+        return self.context.shared_state[GET_FUNDS_STATUS_METHOD_NAME]()
```
LOCKhart07, line 104: Won't have to maintain an extra variable this way

```suggestion
        return self.context.shared_state[
            FundsManagerBehaviour.get_funds_status.__name__
        ]()
```

LOCKhart07: https://github.com/valory-xyz/trader/issues/588

.github/workflows/common_checks.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
           tox -e check-abci-docstrings
           tox -e check-abciapp-specs
           tox -e check-handlers
-          tox -e analyse-service
+          # tox -e analyse-service # analyse-service checks do not work properly as it is checking funds manager skill which is not an ABCI skill
```
Adamantios, line 115: Let's open an issue.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -215,3 +228,11 @@ def _handle_get_static_file(
...
         except FileNotFoundError:
             self._handle_not_found(http_msg, http_dialogue)
+
+    def _handle_get_funds_status(
+        self, http_msg: HttpMessage, http_dialogue: HttpDialogue
+    ) -> None:
+        """Handle a fund status request."""
+        self._send_ok_request_response(
+            http_msg, http_dialogue, self.funds_status.get_response_body()
+        )
```
Adamantios, line 238: Thinking out loud for future improvements:

- Is there any chance for the calculation to fail at some point?
- I would assume yes, so is there any way to identify these failures?
- Maybe we should identify them and not return ok response in these cases.

PR #593, opened 2025-10-03
Fix images build

Adamantios:
Pinning `click` temporarily as a workaround to address a breaking change in its defaults, which is introduced due to a very wide range of `click` used in `open-autonomy`.


Adamantios:
Tested and seems to work in [this run](https://github.com/valory-xyz/trader/actions/runs/18222982443/job/51887089183). Notice how it fails due to the commit not being tagged, but the image build succeeds.

PR #609, opened 2025-10-07
Partial roi part of total roi description

LOCKhart07:
Now inline with spec


Skanislav (approved):
I believe this is a product decision. Please reflect it in the PR description for historical reference

Adamantios (approved):

PR #617, opened 2025-10-15
[Predict-407] Helper scripts for easier development

LOCKhart07:
Two main scripts
1. `make ci-linter-checks` To check all ci linters locally
2. `make run-agent` To run agent locally while developing

Feel free to try out the actual scripts before approving


Makefile
```diff
@@ -213,3 +213,29 @@ endif
 .PHONY: check-agent-runner
 check-agent-runner:
 	python check_agent_runner.py
+
+.PHONY: ci-linter-checks
```
LOCKhart07: Added for developer to run locally before pushing instead of waiting for github ci

pyproject.toml
```diff
@@ -68,7 +68,7 @@ hexbytes = "==0.3.1"
 packaging = "*"
 py-ecc = "==6.0.0"
 pytz = "==2022.2.1"
-typing_extensions = "==4.13.2"
+typing_extensions = "<=4.13.2,>=3.10.0.2"
```
LOCKhart07: Changed by `scrips/check_dependencies`

run_agent.sh
```diff
@@ -1,8 +1,47 @@
-rm -r trader
-find . -empty -type d -delete  # remove empty directories to avoid wrong hashes
+cleanup() {
```
LOCKhart07: Taken as is from https://github.com/valory-xyz/mech-tools-dev/blob/main/run_agent.sh

scripts/aea-config-replace.py
```diff
@@ -0,0 +1,128 @@
...
+
+    # Load the aea config
+    with open(Path(AGENT_NAME, "aea-config.yaml"), "r", encoding="utf-8") as file:
+        config = list(yaml.safe_load_all(file))
+
+    # Search and replace all the secrets
+    for path, var in PATH_TO_VAR.items():
+        try:
+            new_value = os.getenv(var)  # pylint: disable=E1101
+            if new_value is None:
```
LOCKhart07: This is a change from original implementation. Doesn't make the user add ALL the variables defined here. It just defaults to the ones defined in aea config

PR #624, opened 2025-10-23
Remove unnecessary RPC calls

Adamantios:
Double-checking whether a market [is finalized](https://github.com/valory-xyz/trader/blob/008d335d45b93af851b100bdd0ff999c4549d87a/packages/valory/skills/decision_maker_abci/behaviours/reedem.py#L798) is unnecessary, plus a conflict has not been found in the logs yet.

Therefore, removing it is sensible, especially taking into consideration the amount in which it affects the number of RPC calls, therefore providers' subscription costs. For each trade a service has performed, a call will be made at each period. A period runs every 2-5 minutes, or even 1 minute if the KPI goals have been met. This can easily explode to thousands of extra requests per hour for the average service - especially in combination with https://github.com/valory-xyz/open-autonomy/issues/2368.

Additionally, bumping to [mech-interact@v0.21.3](https://github.com/valory-xyz/mech-interact/blob/v0.21.3/packages/packages.json#L19) improves the number of RPC calls when waiting for a request in the new mech marketplace.


Skanislav (dismissed):
I highlighted it in the past. So happy to have this one removed

packages/valory/skills/decision_maker_abci/behaviours/reedem.py
LOCKhart07: `_finalized` variable doesn't seem to be used anymore

PR #637, opened 2025-10-29
fix: change http server timeout to 30 seconds

LOCKhart07:


LOCKhart07:
Already approved here https://github.com/valory-xyz/trader/pull/628

PR #640, opened 2025-10-29
x402: add logic to swap some balances from xdai to usdc in agent eoa

LOCKhart07:
Everytime the agent starts, some xDAI funds from agent eoa are swapped to USDC.e to use for x402
This also happens each time the fund-status api is called.

Implementation is adapted from Optimus' implementation

PR chain:
1. https://github.com/valory-xyz/trader/pull/640
2. https://github.com/valory-xyz/trader/pull/641

### DIFF
--- a/packages/valory/skills/trader_abci/handlers.py
+++ b/packages/valory/skills/trader_abci/handlers.py
@@ -309,7 +309,7 @@ class HttpHandler(BaseHttpHandler):
                 self.context.logger.warning(f"No RPC URL for {chain}")
                 return None
 
-            # Commented for future bebugging purposes:
+            # Commented for future debugging purposes:
             # Note that you should create only one HTTPProvider with the same provider URL per python process,
             # as the HTTPProvider recycles underlying TCP/IP network connections, for better performance.
             # Multiple HTTPProviders with different URLs will work as expected.


packages/valory/skills/trader_abci/handlers.py
```diff
@@ -259,8 +280,225 @@ def _handle_get_funds_status(
...
+            ]
+
+            usdc_contract = w3.eth.contract(
+                address=Web3.to_checksum_address(usdc_address), abi=erc20_abi
+            )
+            balance = usdc_contract.functions.balanceOf(
+                Web3.to_checksum_address(eoa_address)
+            ).call()
+
+            balance_usdc = balance / 10**6  # USDC.e has 6 decimals
```
Skanislav: small improvement: even though it would work for usdce and rest of stables to be it sounds more robust to also fetch the token decimals right after balance to do a conversion

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -77,6 +83,9 @@
 GNOSIS_CHAIN_NAME = "gnosis"
 XDAI_ADDRESS = "0x0000000000000000000000000000000000000000"
 WRAPPED_XDAI_ADDRESS = "0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d"
+USDC_E_ADDRESS = "0x2a22f9c3b484c3629090FeED35F17Ff8F88f76F0"
+GNOSIS_CHAIN_ID = 100
+SLIPPAGE = "0.005"
```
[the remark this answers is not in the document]
Skanislav: I'd also comment here that it represents 0.5%. Regarding the value, I checked CowSwap for the xdai/usdce pair, and they approximate 0.25%. In the lifi documentation, example value is 0.3%. Should be safe to go with 0.003
https://docs.li.fi/api-reference/get-a-quote-for-a-token-transfer

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -259,8 +280,225 @@ def _handle_get_funds_status(
...
+                if isinstance(tx_request.get("gasLimit"), str)
+                else self.params.default_gas_limit
+            )
+
+            tx_data = {
+                "to": Web3.to_checksum_address(tx_request["to"]),
+                "data": tx_request["data"],
+                "value": tx_value,
+                "gas": tx_gas,
+                "gasPrice": gas_price,
```
Skanislav: Also, is there any particular reason not to use EIP-1559? Could be good for POC, but may introduce an overpayment risk

packages/valory/skills/trader_abci/handlers.py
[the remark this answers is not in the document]
LOCKhart07: https://github.com/valory-xyz/trader/pull/640/commits/9bc790c98f3eff2d12ddc65a9e9599475f1db57c

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -259,8 +280,225 @@ def _handle_get_funds_status(
...
+
+            tx_value = (
+                int(tx_request["value"], 16)
+                if isinstance(tx_request["value"], str)
+                else tx_request["value"]
+            )
+            tx_gas = (
+                int(tx_request.get("gasLimit", "0x7a120"), 16)
+                if isinstance(tx_request.get("gasLimit"), str)
+                else self.params.default_gas_limit
```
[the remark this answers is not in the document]
LOCKhart07: https://github.com/valory-xyz/trader/pull/640/commits/ec250d271895fecc04c83cabef24d8564e142667

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: Opened a issue https://github.com/valory-xyz/trader/issues/642

packages/valory/services/trader_pearl/service.yaml
```diff
@@ -147,6 +147,10 @@ models:
       use_multi_bets_mode: ${USE_MULTI_BETS_MODE:bool:false}
       coingecko_olas_in_usd_price_url: ${COINGECKO_OLAS_IN_USD_PRICE_URL:str:https://api.coingecko.com/api/v3/simple/token_price/xdai?contract_addresses=0xcE11e14225575945b8E6Dc0D4F2dD4C570f79d9f&vs_currencies=usd}
       genai_api_key: ${GENAI_API_KEY:str:""}
+      x402_payment_requirements: ${X402_PAYMENT_REQUIREMENTS:dict:{"threshold":0.2,"top_up":0.25}}
```
OjusWiZard: why don't we also take the token address? Because what if USDC address differs? I'm not sure if it's even the same in optimims and gnosis.

OjusWiZard (approved):

packages/valory/services/trader_pearl/service.yaml
LOCKhart07: The USDC address is set in the agent code. As it's specific to trader, only keeping the appropriate address for gnosis is fine imo

PR #660, opened 2025-11-18
[Predict-444] feat: add feature flag for APT

LOCKhart07:
Feature flag is to disable APT for QS

Hashes in https://github.com/valory-xyz/trader/pull/661


packages/valory/services/trader/service.yaml
```diff
@@ -151,6 +151,7 @@ type: skill
         gnosis_ledger_rpc: ${GNOSIS_LEDGER_RPC:str:https://rpc.gnosischain.com/}
         lifi_quote_to_amount_url: ${LIFI_QUOTE_TO_AMOUNT_URL:str:https://li.quest/v1/quote/toAmount}
         use_x402: ${USE_X402:bool:false}
+        is_agent_performance_summary_enabled: ${IS_AGENT_PERFORMANCE_SUMMARY_ENABLED:bool:true}
```
Adamantios: Missing from the rest of the agents.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -268,6 +268,17 @@ def _save_agent_performance_summary(
...
+        if not self.params.is_agent_performance_summary_enabled:
+            self.context.logger.info(
+                "Agent performance summary is disabled. Skipping fetch and save."
+            )
+            payload = FetchPerformanceDataPayload(
+                sender=self.context.agent_address,
+                vote=False,
+            )
+            yield from self.finish_behaviour(payload)
+            return
```
Adamantios: I believe there is no need to return, because of the `self.set_done()` in the `finish_behaviour`.

PR #662, opened 2025-11-18
Add support for discoverability

Adamantios:
- Bumps to the latest mech-interact which supports discoverability.
- Adjusts trader's code to the new mech-interact that supports discoverability.

The changes are based on the instructions [here](https://github.com/valory-xyz/mech-interact/blob/v0.22.0/UPGRADING.md#v0213-to-v0220-built-with-open-aea1650-and-open-autonomy01911).


poetry.lock
LOCKhart07: Why did poetry lock change?

Adamantios: Correct, I simply updated the existing dependencies' versions, i.e., I ran `poetry update`.

LOCKhart07 (dismissed):

DIvyaNautiyal07 (approved):

LOCKhart07 (approved):

PR #672, opened 2025-12-10
Bump open autonomy to `v0.21.4`

LOCKhart07:
Has almost the same changes as https://github.com/valory-xyz/trader/pull/584 which has been approved previously

New changes:
- open-autonomy v0.21.4 instead of v0.21.3
- Older mech-interact version without any discoverability changes. From https://github.com/valory-xyz/mech-interact/pull/32 and https://github.com/valory-xyz/mech-interact/pull/33
- gitleaks utility script
- Updated third party hashes


LOCKhart07:
Closing as the open-aea version used has issues

PR #675, opened 2025-12-12
Refactor/improve performance

DIvyaNautiyal07:
## PR Description:

---

### Refactor: Cache API Endpoint Data in Behaviour Round

This PR refactors the agent API endpoints to pre-cache data during the behaviour round execution instead of querying subgraphs on-demand in handlers.

### Solution

Move data fetching to the behaviour round and cache results in `agent_performance.json`:

1. Behaviour fetches and stores data during `FetchPerformanceDataRound`
2. Handlers read from file
3. For prediction-history, since it is a paginated response, we store only latest 200 bets and fetch more than that on demand
### Changes

__1. Models (`models.py`)__

- Added `PredictionHistory` dataclass (renamed from cache terminology)
- Extended `AgentPerformanceSummary` to store endpoint data

__2. Behaviour (`behaviours.py`)__

- `_fetch_agent_details_data()` - Fetches agent metadata
- `_fetch_agent_performance_data()` - Calculates metrics with balance checks
- `_fetch_prediction_history()` - Fetches latest 200 predictions
- All data written to `agent_performance.json` each round

__3. Handlers (`handlers.py`)__

- Agent details endpoint now reads from file
- Performance endpoint now reads from file
- Predictions endpoint uses file-first approach with subgraph fallback

__4. Shared Helper (`predictions_helper.py`)__

- New `PredictionsFetcher` class to fetch prediction history and calculate per bet pnl
- Shared between behaviour and handler
- Eliminates code duplication

### DIFF
--- a/packages/valory/skills/agent_performance_summary_abci/handlers.py
+++ b/packages/valory/skills/agent_performance_summary_abci/handlers.py
@@ -22,6 +22,7 @@
 
 import json
 from datetime import datetime
+from dataclasses import asdict
 from enum import Enum
 from http import HTTPStatus
 from typing import Any, Callable, Dict, List, Optional, Union, cast
@@ -191,17 +192,18 @@ class HttpHandler(BaseHttpHandler):
         details = summary.agent_details
         
         if not details or not details.id:
-            formatted_response = {
-                "id": None,
-                "created_at": None,
-                "last_active_at": None,
-            }
-        else:
-            formatted_response = {
-                "id": details.id,
-                "created_at": details.created_at,
-                "last_active_at": details.last_active_at,
-            }
+            self._send_internal_server_error_response(
+                http_msg, 
+                http_dialogue,
+                {"error": "Agent details not available. Data may not have been fetched yet or there was an error retrieving it."}
+            )
+            return
+            
+        formatted_response = {
+            "id": details.id,
+            "created_at": details.created_at,
+            "last_active_at": details.last_active_at,
+        }
         
         self.context.logger.info(f"Responding with agent details: {formatted_response}")
         self._send_ok_response(http_msg, http_dialogue, formatted_response)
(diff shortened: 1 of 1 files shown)


packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -225,6 +231,162 @@ def _get_prediction_accuracy(self):
...
+                skip=0,
+                status_filter=None
+            )
+            
+            return PredictionHistory(
+                total_predictions=result["total_predictions"],
+                stored_count=len(result["items"]),
+                last_updated=self.shared_state.synced_timestamp,
+                items=result["items"],
+            )
```
Adamantios: Why inside the `try`?

packages/valory/skills/agent_performance_summary_abci/handlers.py
```diff
@@ -179,69 +187,23 @@ def _handle_get_agent_details(
...
-            "id": trader_agent.get("id", safe_address),
-            "created_at": self._format_timestamp(trader_agent.get("blockTimestamp")),
-            "last_active_at": self._format_timestamp(trader_agent.get("lastActive")),
-        }
+        if not details or not details.id:
+            formatted_response = {
+                "id": None,
+                "created_at": None,
+                "last_active_at": None,
+            }
```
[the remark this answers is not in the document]
DIvyaNautiyal07: I just thought it's better to return null values with 200 OK rather than error responses because if there's a temporary issue fetching data (subgraph down, RPC timeout), the UI can still render gracefully with null values instead of showing error states.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: addressed here - [a845d1f](https://github.com/valory-xyz/trader/pull/675/commits/a845d1f89c1b981d53f4bfcc19f2e1c10d7ffd7a)

DIvyaNautiyal07:
> What is the advantage of using this hybrid approach? Also, what do you mean by this:
> 
> > For prediction-history, since it is a paginated response, we store only latest 200 bets and fetch more than that on demand

@Adamantios With the hybrid approach, the agent performance data is pre-computed periodically by the agent itself (during the AgentPerformanceSummary round), rather than being calculated on-demand for every API request. This makes API responses much faster. Also, by storing the data in agent_performance.json, the UI can always display the last known metrics even if the AgentPerformanceSummary round hasn't run recently or is currently executing.

For prediction history, we store the latest 200 predictions in agent_performance.json for instant access. If users request additional pages (page 2, 3, etc.), we fetch those on-demand from the subgraph. This keeps the file size manageable while still providing fast access to recent data.

PR #783, opened 2026-02-03
Fix/redeeming

LOCKhart07:
### Problem
Bets were not being redeemed despite having winning positions. The system stopped working around late January 2026 for some users

### Root Cause
Python's `bytes.hex()` method strips the `0x` prefix from hex strings, while the results from subgraph always returns condition IDs with the `0x` prefix. This format mismatch caused dictionary lookups to fail in three critical locations:

1. **Primary lookup** (line 787-803): Checking if a position is unredeemed
   - Local: `condition_id.hex()` → `"1ae56459..."`
   - Subgraph dict keys: `"0x1ae56459..."`
   - Result: Lookup fails, system thinks position is already redeemed

2. **Trade filtering** (line 381): Filtering already-redeemed positions
   - Comparison failed due to format mismatch
   - Previously redeemed positions kept reappearing

3. **Claiming state tracking** (line 858): Marking positions as being claimed
   - IDs stored without `0x` prefix
   - Filter in `utils.py` couldn't match them against subgraph data
   - Positions appeared unredeemed even after successful redemption

### Solution
Added `0x` prefix normalization at all three points where `.hex()` output is compared against subgraph data:

1. **Defensive bidirectional normalization** for unredeemed position lookup (lines 787-803)
   - Normalizes both the candidate condition ID and all dictionary keys
   - Handles potential future format changes
   - Uses consistent inline conditional style

2. **Format normalization** in `_filter_trades()` (line 381)
   - Ensures condition IDs have `0x` prefix when comparing against `redeemed_condition_ids`

3. **Format normalization** in `_prepare_safe_tx()` (line 858)
   - Stores condition IDs with `0x` prefix in `claiming_condition_ids`
   - Ensures consistency when moved to `claimed_condition_ids` after transaction success

### Impact
- **Fixes**: Bet redemptions now work correctly for all winning positions
- **Safe**: Changes only affect local variable formatting in lookup operations
- **No side effects**: Original Trade objects remain unchanged
- **Defensive**: Handles both formats (with/without `0x`) for future robustness

### Testing
- Verified with 100+ subgraph trades (all have `0x` prefix)
- Analyzed 23 safe addresses showing redemption failures since late January
- **Confirmed locally with multiple Omenstrat safes which have unredeemed winning positions which are able to be redeemed only after applying the fix**

### Notes
The exact trigger for this issue remains unclear. The bug appears to be latent in the codebase but only manifested for certain users starting late January 2026. Possible explanations include subgraph changes, or indexing causing the data to finally become 'proper'. The defensive fix ensures consistent behavior regardless of the underlying cause.

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -14,16 +14,16 @@
         "contract/valory/relayer/0.1.0": "bafybeicaz3xmdv3eanbzkw63nurx2l2yoacl2wsalxktt5or6oix5rkcxa",
         "contract/valory/market_maker/0.1.0": "bafybeidwc7yvxufnupyjzge2ycbos2t24hoyaeosvzj3z63ccqdzotygdm",
         "skill/valory/market_manager_abci/0.1.0": "bafybeieqeutpxfr56yl4d2ik6kvnjuljrv5jtushqhhbp65gml6jyplyky",
-        "skill/valory/decision_maker_abci/0.1.0": "bafybeihpubhnx3aj3zdicghitlx4v6ffsbngbmbccxnetxrzxuoalfmkvu",
-        "skill/valory/trader_abci/0.1.0": "bafybeieekt33vsfjccajd6pelkmh6gbsbxbxl47uanha5aprejkxuu5cza",
-        "skill/valory/tx_settlement_multiplexer_abci/0.1.0": "bafybeidbuwx5nu2ts6binu5uaxesijyvbpoajgqkxvuq6scz2jbelnz3im",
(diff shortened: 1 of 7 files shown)


packages/valory/skills/decision_maker_abci/behaviours/reedem.py
```diff
@@ -378,7 +378,8 @@ def _filter_trades(self) -> None:
         self.trades = {
             trade
             for trade in self.trades
-            if trade.fpmm.condition.id.hex().lower() not in redeemed_condition_ids
+            if ("0x" + trade.fpmm.condition.id.hex()).lower()
```
DhairyaPatel7: Do we always have the condition Id NOT prefixed with "0x" in the subgraph data

PR #890, opened 2026-03-25
Feat/trader request context enrichment

DIvyaNautiyal07:
# feat: add market context to mech request payloads (benchmark schema v2.0)

## Summary

- Enrich mech request payloads with market context so the benchmark can measure prediction edge over market
- Add `to_request_context()` method to `Bet` that builds context from existing bet data
- Read `spread` from Polymarket Gamma API and derive `amm_fee` from Omen AMM fee
- Wire context through `DecisionRequestBehaviour.setup()` into `MechMetadata`
- Add `market_spread` field to `Bet` dataclass
- Fix Polymarket bets missing `market` field — set `_current_market = "polymarket_client"` in `PolymarketFetchMarketBehaviour.setup()` and backfill existing persisted bets with empty market on refresh
- Use labeled Yes outcome index for `market_prob` instead of hard-coding index 0
- 27 new tests across 3 test files

## Context

Currently, when the trader asks a mech to make a prediction, the IPFS payload contains only `prompt`, `tool`, and `nonce`. The benchmark has no way to measure whether the mech's prediction was better than what the market already knew.

After this change, the payload includes a `request_context` with the market state at request time:

```json
{
  "schema_version": "2.0",
  "prompt": "Will BTC hit $150k by end of 2026? Yes No",
  "tool": "prediction-online",
  "nonce": "a1b2c3d4-...",
  "request_context": {
    "market_id": "0xdef456abc...",
    "type": "polymarket",
    "market_prob": 0.65,
    "market_liquidity_usd": 450000.0,
    "market_close_at": "2026-12-31T00:00:00Z",
    "market_spread": 0.02
  }
}
```

The benchmark can now compute: "The market thought 65% yes, the mech said 72% yes — the mech had a +7% edge."

## Dependencies

Requires [mech-interact PR](https://github.com/valory-xyz/mech-interact) that adds `schema_version` and `request_context` fields to `MechMetadata` (branch `feat/mech-metadata-request-context`).

### `packages/valory/skills/market_manager_abci/bets.py`

- Added `WEI = 10**18` constant
- Added `MARKET_TO_PLATFORM` mapping (`omen_subgraph` -> `omen`, `polymarket_client` -> `polymarket`)
- Added `market_spread: Optional[float] = None` field to `Bet` dataclass
- Added `to_request_context()` method that builds the context dict from existing bet data
- `market_prob` now derived from labeled Yes outcome via `outcomes.index(BinaryOutcome.YES.value)` instead of hard-coded index 0

### `packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py`

- Set `self._current_market = "polymarket_client"` in `setup()` so Polymarket bets get a valid `market` field (was empty string, causing `to_request_context()` to return `None`)
- Backfill existing persisted bets: when `_process_chunk()` updates an existing bet, if its `market` field is empty, set it to `self._current_market`
- Parse `spread` from Gamma API response and include as `market_spread` in bet_dict
- Validate spread is a float in `[0, 1]` range; invalid or missing values default to `None`

### `packages/valory/skills/decision_maker_abci/behaviours/decision_request.py`

- Call `sampled_bet.to_request_context()` in `setup()`
- Pass result as `request_context=` kwarg to `MechMetadata`
- Removed debug print statements

## request_context fields by platform

| Field | Omen | Polymarket |
|-------|------|------------|
| `market_id` | `bet.id` (FPMM contract address) | `bet.condition_id` (on-chain condition ID) |
| `type` | `"omen"` | `"polymarket"` |
| `market_prob` | `outcomeTokenMarginalPrices[yes_idx]` | `outcomeTokenMarginalPrices[yes_idx]` |
| `market_liquidity_usd` | `scaledLiquidityMeasure` (xDAI ≈ USD) | `scaledLiquidityMeasure` (USD) |
| `market_close_at` | `openingTimestamp` as ISO 8601 UTC | `openingTimestamp` as ISO 8601 UTC |
| `amm_fee` | `fee / 10^18` (e.g. 0.02 = 2%) | not present |
| `market_spread` | not present | from Gamma API `spread` field |

`yes_idx` is derived via `outcomes.index(BinaryOutcome.YES.value)` — uses the labeled outcome list rather than assuming index 0.

## Validation guards

- NaN/Inf `market_prob` excluded via `math.isfinite()`
- `market_spread` outside `[0, 1]` excluded (catches NaN, Inf, negatives, values > 1)
- `openingTimestamp <= 0` excluded
- Unknown platform returns `None` (no context)
- `None` values stripped from context dict
- Bets with `outcomes=None` (blacklisted) safely produce `market_prob=None` — these bets are filtered out during sampling and never reach `to_request_context()` in practice

## Backward compatibility

- Mech ignores extra fields — reads only `prompt` and `tool`
- Old traders without this change produce `request_context: null` in the payload (MechMetadata default)
- `BetsDecoder` fallback path handles Bet objects with or without `market_spread`
- Consensus safe: `request_context` is deterministic from shared `sampled_bet`
- Existing persisted Polymarket bets with `market=""` are backfilled on next fetch cycle

### `test_bets.py` — `TestBetToRequestContext` (22 tests)

- [x] Omen bet returns correct context with all fields including `amm_fee`
- [x] Polymarket bet uses `condition_id` as `market_id` with `market_spread`
- [x] Omen falls back to `bet.id` when no `condition_id`
- [x] Unknown platform returns `None`
- [x] `None` values stripped from context
- [x] Valid timestamp produces ISO 8601 UTC with Z suffix
- [x] `market_prob` uses labeled Yes outcome index (mutation-resistant)
- [x] `MARKET_TO_PLATFORM` mapping verified
- [x] Empty `condition_id` falls back to `bet.id`
- [x] NaN, Inf, -Inf `market_prob` excluded
- [x] Zero liquidity included (not stripped)
- [x] Blacklisted bet still returns context without crash
- [x] Omen `amm_fee` derived from fee in wei
- [x] Omen zero fee excludes `amm_fee`
- [x] Polymarket `market_spread` from field
- [x] Polymarket no spread excludes field
- [x] `amm_fee` and `market_spread` are independent
- [x] NaN, Inf, negative, > 1 `market_spread` excluded
- [x] Context is JSON serializable

### `test_behaviours_polymarket.py` (3 tests)

- [x] Valid market reads `spread` from API response
- [x] Invalid spread value results in `None`
- [x] Missing spread defaults to `None` in bet_dict

### `test_decision_request.py` (2 tests)

- [x] `setup()` calls `to_request_context()` and passes result to `MechMetadata`
- [x] `None` return from `to_request_context()` handled correctly

---

Generated with [Claude Code](https://claude.com/claude-code)


packages/packages.json
```diff
@@ -48,27 +48,27 @@
...
+        "connection/valory/abci/0.1.0": "bafybeia2fpbej2tdklgprvuvgtjwswlluzdkbfbj2tbvguzavexn3ulpai",
+        "connection/valory/ipfs/0.1.0": "bafybeifcwxo3l56mlcucpdaoqhiczzkbyemrgtffoxzsqx67g567jcfeya",
+        "skill/valory/mech_interact_abci/0.1.0": "bafybeiawzhnm6ubgcotdp3vco5aw3koo4befon7z6pyinkxo5gucoguraa",
+        "skill/valory/abstract_abci/0.1.0": "bafybeidf6cjcl4ra7ryj3xm74x7o7jfpg6wvgutn7lcijcajcdvh33657q",
+        "skill/valory/abstract_round_abci/0.1.0": "bafybeiamwgraw77dsawauz7fxabiqsa344vrieiljw37eufbl55ijpdmha",
+        "skill/valory/transaction_settlement_abci/0.1.0": "bafybeie2lj22nrliky4xuobpihpaglhju6x4oaq56765ueoxgcqgjahb34",
+        "skill/valory/registration_abci/0.1.0": "bafybeiajjkzebtdlvnsjhhqojgomf2cy6ciridpx3purwbj7frfble5spm",
+        "skill/valory/reset_pause_abci/0.1.0": "bafybeiepj7ek22rgw2blhu3s423xu4u7hjcptluvkcduhvweyrahbzsfkq",
+        "skill/valory/termination_abci/0.1.0": "bafybeigmozjv5tm2nzoj4kwtzo433m75su3mkmk5ulgcasjpomi3b6uaom",
+        "skill/valory/funds_manager/0.1.0": "bafybeigprgxloujy5wpfa4xn7xuwsxafq5r67z4sffjg3kjfovucmlk6mq"
```
LOCKhart07: Funds manager changed?

LOCKhart07 (changes requested):

bennyjo:
## Cross-repo review: benchmark schema v2.0

Reviewed this PR alongside mech#422, mech-predict#160, mech-predict#161, and mech-interact#61 for cross-repo compatibility.

### MEDIUM: Unexplained `funds_manager` hash change

Reviewer LOCKhart07 already flagged this — the `funds_manager` third-party package hash changed in `packages.json`. This PR only adds market context fields and should not affect `funds_manager`. This suggests either:

- An unintended dependency was pulled in during `autonomy packages lock`
- The branch was rebased onto a commit that included this change

Please confirm whether this is intentional or revert the stray hash change to keep the diff clean.

### Deployment ordering note

This PR imports the new `request_context` kwarg on `MechMetadata`, which is added in mech-interact#61. That PR **must** be merged and its package published before this one can deploy, otherwise the `MechMetadata` constructor will reject the unknown kwarg at runtime.

---

_Cross-repo review covering: mech#422, trader#890, mech-predict#160, mech-predict#161, mech-interact#61_

mariapiamo (commented):
1. This PR assumes outcomeTokenMarginalPrices[0] always corresponds to the benchmark probability we want to serialize, but other existing code paths appear to use a different outcome-index mapping. Could we derive market_prob from the labeled Yes outcome instead of hard-coding index 0
2. From a benchmarking-flow perspective, the currently fetched request-context fields look sufficient to support a first end-to-end implementation, naming aside. We already have the key contemporaneous market context needed to start wiring the loop: market identifier, platform/type, probability snapshot, liquidity snapshot, spread, and close time. The remaining gaps from proposal #139 feel more like completeness/future-analysis improvements than blockers for a first full benchmarking flow.
3. I'll check next subgraph

LOCKhart07 (dismissed):

bennyjo (dismissed):

mariapiamo (approved):
The market_prob fix looks good now. One small cleanup I’d suggest is updating the test name/docstring in test_bets.py that still refers to using the first index, since the intended behavior is now to use the index of the labeled Yes outcome. But is Not a blocker for me

LOCKhart07 (approved):

mariapiamo:
The benchmark-related payload changes do not appear to break [autonolas-subgraph](https://github.com/valory-xyz/autonolas-subgraph) or [autonolas-subgraph-studio](https://github.com/valory-xyz/autonolas-subgraph-studio). The request-side enrichment comes from [trader PR #890](https://github.com/valory-xyz/trader/pull/890) and is carried in the mech envelope by [mech-interact PR #61](https://github.com/valory-xyz/mech-interact/pull/61), while the response-side enrichment is added in [mech PR #422](https://github.com/valory-xyz/mech/pull/422).

These changes are additive: they add fields like schema_version, request_context, executed_at, cost_dict, and richer metadata, but they preserve the existing keys the current subgraphs parse.
For payload-shape compatibility, the relevant blast radius is limited to the mech IPFS readers in [autonolas-subgraph/subgraphs/mech](https://github.com/valory-xyz/autonolas-subgraph/tree/main/subgraphs/mech) and [autonolas-subgraph/subgraphs/marketplace](https://github.com/valory-xyz/autonolas-subgraph/tree/main/subgraphs/marketplace). Those readers still only depend on a narrow stable subset of the payload:

request side: prompt and tool
response side: result and metadata.model


I did not find the same mech request/response IPFS dependency in autonolas-subgraph-studio, including [predict-omen](https://github.com/valory-xyz/autonolas-subgraph-studio/tree/main/subgraphs/predict/predict-omen) and [predict-polymarket](https://github.com/valory-xyz/autonolas-subgraph-studio/tree/main/subgraphs/predict/predict-polymarket), nor in unrelated subgraphs like service-registry, staking, tokenomics, governance, liquidity, babydegen, or mech-marketplace.

So the benchmark flow can already run by reading the richer payload from IPFS directly, while the current subgraphs remain compatible for request/delivery discovery. For example, the current subgraph can expose prompt, tool, and the raw content, but new fields like request_context are only recoverable from that raw content and are not yet indexed as first-class schema fields. If we want those benchmark fields to be directly queryable through subgraphs later, that would be a follow-up schema/mapping enhancement rather than a compatibility fix.\


Note: the existing subgraph was tested against a trader request with same format, would be great to test against the mech response as well. I am blocked here from mech fixes and deployment for that

PR #899, opened 2026-03-31
chore: migrate from Poetry to uv with tox-uv

DhairyaPatel7:
## Summary
- Migrate dependency management from Poetry to uv (PEP 621 standard)
- Add tox-uv plugin for faster CI venv creation
- Replace poetry.lock with universal uv.lock

## Verified locally
- uv lock --check
- uv sync --all-groups
- tox -e black-check
- tox -e check-packages
- tox -e check-hash
- tox -e liccheck
- tox -e safety

## Test plan
- [ ] CI lock_check passes on all OSes
- [ ] CI linter checks pass
- [ ] CI test matrix passes


packages/valory/connections/polymarket_client/tests/test_connection.py
```diff
@@ -1972,11 +1972,14 @@ def test_json_decode_error_caught_and_retried(self) -> None:
...
+        with (
+            patch(
+                "packages.valory.connections.polymarket_client.connection.requests.get",
+                return_value=mock_response,
+            ),
+            patch(
+                "packages.valory.connections.polymarket_client.connection.time.sleep",
+            ),
         ):
             result, error = conn._request_with_retries("https://example.com/api")
```
LOCKhart07, line 1982: Have we gone back to an older black version? Or is this a newer version?

DIvyaNautiyal07 (approved):

packages/valory/connections/polymarket_client/tests/test_connection.py
DhairyaPatel7: It is the same version (black 26.1.0, pinned by tomte 0.6.2), but tox-uv resolves a slightly different compiled binary than regular tox+pip. The binary produces minor formatting differences on a few files. The version hasn't changed — it's the same black==26.1.0 that tomte pins.

LOCKhart07 (approved):

PR #902, opened 2026-04-09
docs: tool selection pipeline spec (draft)

LOCKhart07:
## Summary

- Merges three source proposals into a single tool selection pipeline spec: the original [Google Doc](https://docs.google.com/document/d/14qbNuJJ9eK2_YFjlwkyx_eO_vulU6BMk9UhHdiBjPFc/edit?tab=t.0#heading=h.q1k3gt1ns4rd) (two-gate filtering), the [Benchmark Proposal Update gist](https://gist.github.com/DIvyaNautiyal07/20c01c4b70304ff1d860372dab404e6b) (Brier-first metrics + dynamic routing), and the local `BUG_REPORT_NVM_SUBSCRIPTION_DEADLOCK.md`.
- Defines a three-layer pipeline with clean ownership: Layer 1 catalog gates in [`valory-xyz/mech-interact`](https://github.com/valory-xyz/mech-interact), Layer 2 segment-aware tool selection in `trader`, Layer 3 benchmark loop in [`valory-xyz/mech-predict`](https://github.com/valory-xyz/mech-predict).
- Addressed a thorough review pass — all 16 review points resolved via locked decisions or clarifications (see commit `58b448a4`).
- Tracks **19 locked decisions**, 3 open questions with recommended defaults, 1 genuinely open item, 1 future/out-of-scope item.

## What's in the spec

- **Architectural principles** — mech selection is mech-interact's job; tool selection is the caller's; classifier is the contract for cross-repo agreement.
- **Layer 1 — catalog gates**: cold-start grace (20 global requests, bypasses reputation only), category gate (hybrid IPFS schema), capability gate (`allowed_payment_types` + `payment_type` cached on `MechInfo`, with explicit prerequisite refactors for `PaymentType` relocation and `_get_payment_type` sharing), reputation gate (Wilson lower bound + separate `gate_liveness_half_life_seconds` parameter, existing `HALF_LIFE_SECONDS` unchanged), name gate (existing `irrelevant_tools` kept as a first-class fourth dimension). Composition order spelled out with logical and efficient-execution variants.
- **Layer 2 — category-aware `EGreedyPolicy`**: nested `Dict[str, Dict[str, AccuracyInfo]]` shape (chosen to require zero encoder/decoder changes), one new optional column on the existing `tools_accuracy_hash` CSV, shared `classify_category()` adopted from mech-predict, `tradeable_categories` config in trader, precise definition of `_candidates_for_category` fallback rule.
- **Layer 3 — benchmark loop**: Brier-first ranking, `by_tool_category` cross-breakdown, conditional accuracy when disagreeing, disagreement-stratified Brier. Net-new CSV export + IPFS publish steps.
- **Migration notes** — per-file implementation checklist covering `policy.py`, `storage_manager.py`, `tool_selection.py`, `polymarket_fetch_market.py`, `bets.py`, `skill.yaml`, and the shared classifier file.
- **Adjacent NVM FSM safety fixes** — orthogonal to selection, ship alongside Layer 1.
- **Rollout sequence** — 12 dependency-ordered steps with parallelisable markers, including explicit prerequisite refactors (`PaymentType` relocation, `_get_payment_type` refactor) before the capability gate work.

## Status

- 19 decisions locked
- 3 open questions with recommended defaults (Wilson 95%, hybrid IPFS schema, `min_mech_score = 0.3`)
- 1 genuinely open: min `n` per cell for routing
- 1 future item: data-driven `tradeable_categories` from disagreement-stratified Brier

## Explicitely want reviewer input

- **Classifier sync mechanism** — the spec currently proposes hash-pin with committed `.sha256` files + nightly cross-repo GitHub Action. Alternatives considered: commit-pin (consumer stores canonical SHA), git submodule, tiny PyPI package. Hash-pin picked because it's the minimum tooling that catches both local and cross-repo drift, but this is explicitly marked open for reviewer input on the exact sync approach.

## Test plan

- [ ] Documentation only — no code changes, no tests to run
- [ ] Spec review by trader / mech-interact / mech-predict owners
- [ ] Resolve remaining open questions before implementation begins
- [ ] Coordinate with existing [`BENCHMARK_BRANCH_PLAN.md`](https://github.com/valory-xyz/mech-predict/blob/main/BENCHMARK_BRANCH_PLAN.md) and [`WORKFLOW_SPLIT_PLAN.md`](https://github.com/valory-xyz/mech-predict/blob/main/WORKFLOW_SPLIT_PLAN.md) in mech-predict before opening Layer 3 work

CI is intentionally skipped for these commits (docs-only).

### DIFF
--- a/TOOL_SELECTION_SPEC.md
+++ b/TOOL_SELECTION_SPEC.md
@@ -47,31 +47,34 @@ ELIGIBILITY (catalog gates)  →  PERFORMANCE (segment routing)  →  FEEDBACK (
 
 ## Layer 1 — Catalog gates (mech-interact)
 
-All gates run inside `mech_interact_abci/behaviours/mech_info.py::populate_tools()` (or its successor) at pool construction time. Pool is rebuilt per period.
+All four gates run in `populate_tools()` (`behaviours/mech_info.py:71-100`), the same function that today applies the `irrelevant_tools` filter at line 96. Mechs that fail a mech-level gate (category, capability, reputation) are skipped with `continue`, so their `relevant_tools` are never populated. The name gate runs at tool level within the mech's tool list (same site as today's `relevant_tools = set(res) - self.params.irrelevant_tools`). The `mech_tools` property (`states/base.py:319-324`) stays read-only — it simply unions `relevant_tools` across mechs, and all filtering has already happened upstream.
 
-### Gate composition order
+### Gate composition order (logical)
 
 ```
 for mech in candidate_mechs:
-    if mech.received_requests < COLD_START_GRACE:    # 1. Grace bypass
-        admit(mech)
+    if not category_match(mech, task_type):            # 1. Category gate
+        skip(mech); continue
+    if not capability_match(mech, allowed_payment_types):  # 2. Capability gate
+        skip(mech); continue
+    if mech.received_requests < COLD_START_GRACE:     # 3. Grace bypass (reputation only)
+        admit(mech, filter_tools_by=irrelevant_tools)
         continue
-    if not category_match(mech, task_type):           # 2. Category gate
-        skip(mech); continue
-    if not capability_match(mech, allowed_payment_types):  # 3. Capability gate
-        skip(mech); continue
     if mech.gate_score < min_mech_score:              # 4. Reputation gate
         skip(mech); continue
-    admit(mech)
+    admit(mech, filter_tools_by=irrelevant_tools)      # 5. Name gate (per-tool)
 ```
 
-Order rationale: grace runs first (otherwise it'd be redundant). Category is cheap and discards the most volume. Capability requires `payment_type` cached on `MechInfo`. Reputation is last because the score computation is the most expensive.
+**Key property:** grace exempts a mech from Wilson/reputation checks only. Category, capability, and name gates still apply to mechs in the grace window — because those aren't about evidence, they're about correctness. A new image-generation mech can't fake its way into a prediction agent's pool by being new.
+
+**Implementation may reorder gates for efficiency** as long as the AND-semantics are preserved. In particular: reputation and grace data (`received_requests`, `self_delivered`, `last_delivered`) is already on `MechInfo` from the subgraph query, so checking reputation/grace first saves the cost of fetching IPFS metadata and the payment-type contract call for mechs that would be excluded anyway. Efficient execution order is typically reputation-or-grace → capability → IPFS fetch → category → name. The final pool is identical to the logical order because all gates are AND conditions.
 
 ### 1. Cold-start grace period
 
-**Locked.** A mech with `received_requests < 20` (counted from the subgraph, **global** across all agents) bypasses all reputation/capability/category checks and enters the pool.
+**Locked.** A mech with `received_requests < 20` (counted from the subgraph, **global** across all agents) bypasses **only the reputation gate**. Category, capability, and name gates still apply — grace is a reputation exception, not a correctness exception.
 
 - **Rationale:** prevents the death spiral where a new mech can't earn reputation because the gate excludes it before it gets a chance.
+- **Why only reputation:** Wilson lower bound needs evidence to compute a meaningful score. Category (wrong type of tool), capability (unpayable), and name (explicit blocklist) are not about evidence — they're about whether the mech belongs in the pool at all. A new image-generation mech is never valid for a prediction agent regardless of age.
 - **Why global:** if the mech has already been hammered by other agents on the network and proven bad, this agent shouldn't grant it a fresh local trial. Subgraph already exposes `received_requests` per mech.
 - **Why 20:** at 20/20 successful deliveries, Wilson lower bound ≈ 0.83 → `gate_score ≈ 0.90` — solidly clear of the reputation gate. At 0/20, Wilson ≈ 0 → fails immediately at request #21. Clean handoff.
 
@@ -120,22 +123,30 @@ allowed_payment_types: null   # default: no gate, today's behaviour
 allowed_payment_types: ["NATIVE", "TOKEN_USDC", "TOKEN_OLAS"]
 ```
 
-**Vocabulary:** the existing `PaymentType` enum (`mech_interact_abci/behaviours/request.py:85-91`):
+**Vocabulary:** the existing `PaymentType` enum (currently `mech_interact_abci/behaviours/request.py:71-83`, relocating to `mech_interact_abci/payment_types.py` per prerequisite below):
 - `NATIVE`
 - `TOKEN_USDC`
 - `TOKEN_OLAS`
 - `NATIVE_NVM`
 - `TOKEN_NVM_USDC`
 
-**Plumbing (locked):** today, payment type is discovered *per request* in `_fetch_and_validate_payment_type()` (`behaviours/request.py:690`), which calls `_get_payment_type()` (`behaviours/request.py:634-647`) against `MechMM.contract_id`. The wrapper, the contract method, and the data flow all already exist — they're just wired into the request path instead of pool construction.
+The related constants (`TOKEN_PAYMENT_TYPES`, `NVM_PAYMENT_TYPES`, `PAYMENT_TYPE_TO_NVM_CONTRACT`) at `request.py:85-91` move to the same new module.
+
+**Plumbing (locked):** today, payment type is discovered *per request* in `_fetch_and_validate_payment_type()` (`behaviours/request.py:690`), which calls `_get_payment_type()` (`behaviours/request.py:634-647`) against `MechMM.contract_id`. The contract call exists, but as currently written it's bound to `MechRequestBehaviour`, reads `self.priority_mech_address` implicitly, and stores the result in a fixed placeholder (`MechRequestBehaviour.mech_payment_type`). It can't be called as-is from `MechInformationBehaviour`, so the plumbing is a real refactor, not a wiring move.
+
+**Prerequisites:**
+
+1. **Move `PaymentType` to a neutral module.** Create `packages/valory/skills/mech_interact_abci/payment_types.py` and move the `PaymentType` enum + related constants from `behaviours/request.py:71-91` into it. `behaviours/request.py` and `states/base.py` both import from the new module. Reason: `MechInfo` lives in `states/base.py`, and today `behaviours/` imports from `states/` — never the reverse. Adding `payment_type: Optional[PaymentType]` to `MechInfo` without this move creates a circular import.
+
+2. **Refactor `_get_payment_type()` into the shared base.** Move `_get_payment_type` (and `_mech_mm_contract_interact` if it's also hard-coded to `self.priority_mech_address`) from `MechRequestBehaviour` into `MechInteractBaseBehaviour` (`behaviours/base.py`). Signature becomes `_get_payment_type_for(self, mech_address: str) -> Generator[...]` and stores the fetched value in a generic placeholder attribute (e.g. `self._last_fetched_payment_type`) that callers read after the `yield from`. `MechRequestBehaviour` updates its usage to pass its own `priority_mech_address` and copy the result into `self._mech_payment_type`. `MechInformationBehaviour` calls it per candidate mech with `mech.address`.
 
 **Implementation:**
 
-1. **`MechInfo` field:** add `payment_type: Optional[PaymentType] = None` to the dataclass at `states/base.py:168-272`.
-2. **Behaviour-level cache:** add `_payment_type_cache: Dict[str, PaymentType]` as an instance attribute on `MechInformationBehaviour` (`behaviours/mech_info.py:48`). Lives for the agent's lifetime; cold starts re-fetch.
-3. **Populate in `populate_tools()`:** after the IPFS metadata fetch for each mech (already a per-mech HTTP call), check the cache. On hit, copy into `mech.payment_type`. On miss, `yield from self._get_payment_type_for(mech.address)`, store on cache and on `mech.payment_type`.
-4. **Capability filter:** in the `mech_tools` property (`states/base.py:319-324`) or alongside the existing `irrelevant_tools` filter, drop mechs whose `payment_type ∉ self.params.allowed_payment_types` (when the param is non-`None`).
-5. **Cold start cost:** ~50 sequential eth_calls × ~100ms ≈ ~5 seconds, once per agent lifetime. Subsequent periods amortise to ~0 (only any new mechs require a fresh call).
+1. **`MechInfo` field:** add `payment_type: Optional[PaymentType] = None` to the dataclass at `states/base.py:168-272`, importing `PaymentType` from the new neutral module.
+2. **Behaviour-level cache:** add `_payment_type_cache: Dict[str, PaymentType]` as an instance attribute on `MechInformationBehaviour` (`behaviours/mech_info.py:48`). Cache key is **`mech.address`** — mech addresses are contract addresses, immutable for the life of that deployment. Mech redeployment at a new address is a cache miss → fresh fetch, which is the correct invalidation semantic. Cold starts re-fetch; in-memory only for v1.
+3. **Populate in `populate_tools()`:** alongside the existing IPFS metadata fetch for each mech, check the cache. On hit, copy into `mech.payment_type`. On miss, `yield from self._get_payment_type_for(mech.address)`, read `self._last_fetched_payment_type`, store on cache and on `mech.payment_type`.
+4. **Capability filter in `populate_tools()`:** the filter runs in the same function as the existing `irrelevant_tools` filter (at `mech_info.py:96`), not in the read-only `mech_tools` property. Mechs whose `payment_type ∉ self.params.allowed_payment_types` (when the param is non-`None`) are skipped with `continue` so their `relevant_tools` are never populated.
+5. **Cold-start cost:** at ~50 mechs with ~100-200ms per eth_call, ~5-10s total cold-start overhead on top of the existing per-mech HTTP metadata fetch (~5-25s). Worst-case total is ~10-35s per cold start, which is at or above the current `MechInformationRound` timeout of **30s** (`rounds.py:168-170`). The existing retry path (`mech_tools_api.increment_retries()`) recovers from `ROUND_TIMEOUT` by re-running `populate_tools` on the next period. **Validate during implementation with real RPC latency measurements; may need to bump the round timeout to 60s if cold starts consistently exceed the budget.** Not pre-emptively changing the timeout because the decision should be evidence-driven.
 6. **Persistence (deferred):** in-memory only for v1. If durability across agent restarts becomes desirable, plumb through `storage_manager.py` (same pattern as `tools_accuracy_hash`) as a follow-up.
 
 **Subgraph extension considered and deferred.** The query at `graph_tooling/queries/mechs_info.py:24-57` does not currently expose `payment_type`. Adding it would eliminate the contract calls but requires subgraph migration; not worth the cost at this scale.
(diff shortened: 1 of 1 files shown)


DIvyaNautiyal07 (commented):
## Design Review — Logical Concerns

### 1. Reputation gate measures delivery, not accuracy — a confident liar passes forever

The reputation gate uses `self_delivered / received_requests` (Wilson lower bound). This measures **whether the mech responded**, not **whether its answer was correct**. A mech that always responds with `p_yes = 0.5` for every question gets 100% delivery rate, maximum Wilson score, and sails through the reputation gate permanently.

The spec says "Layer 3 benchmarks handle per-tool quality." But Layer 3 feeds into Layer 2 (the CSV that EGreedyPolicy consumes), not into Layer 1. There's no feedback loop from accuracy back to catalog eligibility. A consistently wrong but reliably-delivering mech stays in the catalog forever — it just gets deprioritized in tool selection. It still consumes pool slots, request quota, and adds noise.

The reputation gate gives a false sense of quality filtering — it's really a liveness/spam filter. The spec should either:
- Be explicit about this limitation (it filters dead/spam mechs, not bad-quality ones)
- Or add an accuracy-based eviction mechanism in Layer 1 that consumes Layer 3 scores

---

### 2. Staleness asymmetry between live and benchmark data

The spec says:
> `update_accuracy_store(tool, winning)`: update to write `accuracy_store[tool][""]` (aggregate row). Per-category cells are updated by Layer 3 via the IPFS CSV refresh, not by per-resolution updates in the trader.

This creates a split-brain: the aggregate cell is updated live (every bet resolution), but per-category cells are only updated when the benchmark publishes a new CSV. So:

- Tuesday: benchmark says tool X is 75% on politics (from last week's data)
- Wednesday–Friday: tool X goes 2/10 on politics (live bet resolutions)
- The aggregate cell drops to reflect the bad streak. But the per-category cell still says 75%.
- Category-aware routing keeps sending politics questions to tool X based on stale data.

The spec has no discussion of update cadence for the CSV, no staleness bounds, and no mechanism for the trader to signal back "this tool is doing badly on this category right now." Per-category routing could actively harm performance if the benchmark data is stale, because it overrides the trader's live aggregate signal with week-old per-category data. At minimum, the spec should define the update cadence and discuss what happens when live and benchmark signals diverge.

---

### 3. Cross-repo classifier sync is ongoing maintenance cost for uncertain value

The spec proposes duplicating `category_keywords.py` across trader and mech-predict, with SHA256 hash files, CI checks in both repos, a nightly cross-repo GitHub Action, and coordinated PRs for every classifier update. This is meaningful ongoing engineering overhead for a ~50-line file.

The sync mechanism is solving a problem created by the architecture choice (duplicate files in two repos). Alternatives that avoid the problem:
- **Single shared package** (even tiny)
- **Runtime fetch** from a single canonical location (IPFS, since you already use it)
- **Trader doesn't classify at all**: mech-predict classifies and publishes `(tool, category, accuracy)` rows. Trader needs `Bet.category` at `select_tool` time, but could use a simpler mechanism — Polymarket already exposes API tags, and for Omen the category could be embedded in the CSV row lookup rather than requiring the trader to run the same keyword classifier independently.

The last option is the simplest: the benchmark owns classification, the CSV carries the labels, and the trader just needs a lightweight way to tag incoming markets (API tag or a simple lookup) without duplicating the full keyword classifier and maintaining cross-repo sync infrastructure.

---

### 4. No success metrics defined

The spec doesn't define how to measure whether the pipeline is working. What does "better" look like after rollout?

- Higher prediction accuracy?
- Better Brier score?
- More profit?
- Fewer NVM deadlocks?
- Smaller `irrelevant_tools` list?

Without success metrics, you can't evaluate the rollout or make go/no-go decisions at each of the 12 steps. Suggest adding a "How we'll know it's working" section with concrete metrics and thresholds for each layer.

LOCKhart07:
Thanks @DIvyaNautiyal07 — going point by point with proposed resolutions. Want sign-off before I edit the spec.

---

### 1. Reputation gate measures delivery, not accuracy

Concede. It's a liveness/spam filter. Proposed fix: **rename only** — "reputation gate" → **"delivery gate"**, plus an explicit one-liner in the section: *this gate measures whether the mech responds, not whether it's right.*

**Not adding an accuracy-eviction path in Layer 1.** Accuracy quality is Layer 3's job and already has a home: the name gate (`irrelevant_tools`, locked decision #12) is the hook if we ever want to translate Layer 3 signals into exclusion. Keeping mech-interact free of Brier/accuracy awareness is a deliberate architectural principle ("mech-interact never sees Brier scores, market platforms, or question categories"). I'd rather leave the gap explicit than cross that boundary.

### 2. Staleness asymmetry between live aggregate and per-category CSV

Concede. Proposed additions to Layer 2:

- **Update cadence (lock it):** Layer 3 re-pins the CSV **daily**, aligned to `benchmark_flywheel.yaml`. Needs confirmation from the mech-predict side — @DIvyaNautiyal07 does daily match what you're planning?
- **Freshness guard on per-category cells:** `_candidates_for_category` and `_weighted_accuracy_for` require the cell's `max` timestamp to be within `cell_freshness_window` (default **14 days**) of "now". Stale cells fall through to the aggregate. One-line check, no new plumbing.
- **Divergence policy:** if `|aggregate_live − aggregate_csv| > divergence_threshold` (default **0.10**), skip per-category routing for that tool and use the live aggregate. Covers the "tool degraded this week, CSV hasn't caught up" case directly.
- Both thresholds go in as Open Questions — placeholders, tune post-rollout.

### 3. Cross-repo classifier sync overhead

Agreed, hash-pin + nightly cross-repo action is heavy. **Propose phasing**:

- **Phase 1 (initial rollout): no shared classifier in trader.**
  - Polymarket: use the existing API tag as `Bet.category` (already exposed).
  - Omen: `Bet.category = None`, aggregate fallback — same as today. Omen gets Layer 1 + Layer 2 aggregate routing, no per-category routing initially.
  - Zero cross-repo sync, zero `.sha256`, zero nightly workflow. `POLYMARKET_CATEGORY_KEYWORDS` / `_validate_market_category()` stay or are simplified.
- **Phase 2 (deferred):** if Layer 3 numbers show Omen per-category routing is worth the sync cost, re-evaluate then — tiny PyPI package, IPFS runtime fetch, or mech-predict embedding category hints keyed by market id. Decision made with data in hand.

Reverses locked decisions #11 and #18 and removes ~100 lines of the spec. Explicitly flagging this because it's the biggest structural change in these replies.

### 4. Success metrics

Concede — adding a new section. Annotated below so we agree on the *why* for each metric, not just the metric.

**Layer 1 — is the gate doing real work without over-filtering?**

- **Pool size before/after each gate enables.** Validates the gate is actually excluding mechs, and quantifies how many. Expect meaningful reduction (~10–20% per gate); zero means the gate is a no-op and should be re-examined; >50% means we over-filtered and are starving selection.
- **Zero NVM-deadlock incidents in the 2 weeks post capability-gate enable.** This is the concrete bug the capability gate exists to kill. If we still see NVM deadlocks, the gate isn't wired correctly — binary pass/fail check.

**Layer 2 — is category routing actually better than aggregate routing?**

- **Per-category Brier vs aggregate Brier on the same tool, for cells above `N_MIN_CELL`.** The core question — is routing by category picking more accurate tools than the aggregate would have? If per-category cells aren't Brier-better than aggregate, the whole Layer 2 exercise is noise.
- **Tool-selection entropy per category.** Are we actually routing differently across categories, or collapsing to one winner everywhere? Low entropy across all categories means the mechanism exists but doesn't do anything interesting — same tool wins regardless, so we added complexity for no routing benefit.
- **Fraction of selections hitting aggregate fallback vs per-category cell.** Operational visibility — tells us what share of the traffic is actually benefiting from Layer 2 vs falling through to pre-Layer 2 behaviour. Softer signal than the first two, but cheap to emit and useful for interpreting them.

**Layer 3 — is the benchmark producing usable data?**

- **CSV cell coverage: count of `(tool, category)` cells with `n ≥ N_MIN_CELL`.** The benchmark is only useful to the trader at cells it's sufficiently filled. If most cells sit below the floor, Layer 2 is routing on aggregate and we effectively have no per-category system. Coverage is the capacity metric.
- **Brier leaderboard stability week-over-week.** Churn test — if the "best tool" for politics flips every week, either the sample is too small (underpowered benchmark) or tools are genuinely volatile (routing decisions from last week don't apply this week). Either way, stability tells us whether the CSV is trustworthy as routing input.

**Overall — is the system doing its actual job?**

- **Trader P&L delta (7-day rolling, pre vs post rollout).** The ultimate test. All the machinery above is instrumental; this is what the work is for.
- **Bet volume sanity.** Over-gating (Layer 1 too strict, Layer 2 quarantining everything) could starve the agent of bets. Flat P&L at half the volume is not a win. Volume is the "did we break something" guardrail.

Each step in the rollout sequence (§12) gets go/no-go against the metrics relevant to that layer.

---

Holding off on spec edits until you've had a chance to push back, especially on #3. Once aligned I'll apply as one commit per concern.

mariapiamo:
[P2] Wilson metric is not formally specified
Refs: line 160 (line 160), line 195 (line 195).
The spec cites “Wilson lower bound (95%)” but does not define full math and edge-case rules (n=0, 0/0, validity/clamping), so implementations can diverge.

[P2] Category authority and wrong-category handling are not explicit enough
Refs: line 242 (line 242), line 257 (line 257).
The spec should explicitly state who sets each category type and deterministic behavior for wrong/unknown categories.

[P2] New-tool promotion policy is underspecified
Ref: line 342 (line 342).
There is no precise trigger/evidence/guardrail policy for promoting newly added tools that benchmark better.

[P3] 0.6 / 0.4 gate-score weights are heuristic without explicit rationale
Ref: line 162 (line 162).
Operationally plausible, but no calibration or policy basis is documented.

[P3] Cold-start threshold (20) appears heuristic
Refs: line 74 (line 74), line 183 (line 183).
Conceptually valid, but not tied to explicit empirical calibration or formal minimum-evidence policy.

Add explicit category authority lines:
Tool category = mech self-declared metadata.
Market category = shared classify_category(title).
Tradeable set = trader tradeable_categories.

Add wrong-category behavior lines:
Wrong tool declaration = controlled by reputation/benchmark + irrelevant_tools kill switch.
Wrong market classification = aggregate fallback + logging for classifier fixes.
Unknown category = deterministic fallback to uncategorized/aggregate.

Add new-tool promotion policy section with:
Trigger, minimum evidence, and rollout safety guardrails (ramp, rollback, exposure cap).

LOCKhart07:
Thanks @mariapiamo. Proposed resolutions below.

---

### [P2] Wilson metric not formally specified (lines 160, 195)

Concede. Adding full formula + edge cases to the spec:

```
n = received_requests
k = self_delivered

if n == 0:
    wilson_reliability = 0.0
else:
    p_hat = k / n
    z = 1.96                       # 95% confidence
    denom = 1 + z**2 / n
    center = p_hat + z**2 / (2*n)
    margin = z * sqrt((p_hat*(1 - p_hat) + z**2 / (4*n)) / n)
    wilson_reliability = max(0.0, (center - margin) / denom)
```

Edge cases: `n == 0 → 0.0` (fails gate, grace bypass is the intended admission path); `k > n` treated as data error, clamp to `k = n`; final value clamped to `[0, 1]`.

### [P2] Category authority and wrong-category handling (lines 242, 257)

Concede — and flagging that the spec currently uses "category" for two distinct taxonomies. They need to be split cleanly in the revision:

- **Tool/mech type** (`prediction`, `image_generation`) — Layer 1, mech-interact. Mechs self-declare in IPFS metadata; trader picks via `task_type` param.
- **Market topic** (`politics`, `travel`, `crypto`, `health`) — Layer 2, trader. Used for per-`(tool, topic)` Brier routing.

**Tool/mech type (Layer 1):**

- **Authority** = mech self-declared in IPFS metadata. Example:

  ```json
  {
    "tools": [
      { "name": "prediction-online",  "category": "prediction" },
      { "name": "image-gen-sd",       "category": "image_generation" }
    ]
  }
  ```

  Parser fallback chain: per-tool `category` → mech-level `category` → `"uncategorized"`. See the existing IPFS metadata schema section in the spec for the mech-level and legacy shapes.

- **Wrong self-declaration** (e.g. image-gen mech declares `category: prediction`): the delivery gate does **not** catch this — the mech will happily respond with garbage. Caught by (a) `irrelevant_tools` kill switch once observed, (b) Layer 3 benchmarks over time (Brier collapses), (c) eventually the name gate excludes the offending tool.
- **Unknown / missing type** (legacy mechs with no `category` field): parser falls back to `"uncategorized"`. Trader's `task_type = prediction` excludes these by design (caller opted in to strict typing).

**Market topic (Layer 2):**

- **Authority (Phase 1)** = Polymarket API tag for Polymarket; `None` for Omen. (Phase 2 may introduce a shared classifier — deferred pending Layer 3 numbers, per my reply to @DIvyaNautiyal07 above.)
- **Wrong market classification** (Polymarket API tag mislabels a market): aggregate fallback at Layer 2 (`category=""` cell) + log the mismatch for upstream fix.
- **Unknown / missing market topic**: deterministic fallback to the aggregate row. Already the behaviour in `_candidates_for_category`.

**Tradeable set** = trader's `tradeable_categories` config in `market_manager_abci/skill.yaml`. This refers to **market topics**, not tool types.

Spec edit will update terminology throughout so the distinction is never implicit.

### [P2] New-tool promotion policy (line 342)

**Deferring** — flagging as a follow-up, not in scope for this spec. Rationale: the benchmark is the promotion signal, eps-greedy exploration already gives new tools traffic, and the spec's Layer 3 leaderboard-stability metric already surfaces the underlying risk. Adding a probation/ramp mechanism is complexity for a problem we haven't observed. Will add an Open Question entry pointing at this so it's tracked.

### [P3] 0.6 / 0.4 gate-score weights (line 162)

Concede as heuristic. Adding rationale:

> Reliability weighted higher (0.6) than liveness (0.4) because reliability is the stronger signal — a 100%-reliable idle mech is more useful than a 50%-reliable active one. Weights are placeholders; tune against real `gate_score` distributions post-rollout.

Added to Open Questions: calibration of `(0.6, 0.4)` against observed mech distributions.

### [P3] Cold-start threshold = 20 (lines 74, 183)

Already half-addressed at line 79 ("Why 20" — at 20/20 Wilson ≈ 0.83 → clean handoff to the gate). Adding one more line acknowledging it's a math-driven placeholder, not empirically calibrated against real mech-onboarding data.

Added to Open Questions: empirically validate `COLD_START_GRACE = 20` against observed onboarding distributions (how many global requests does a legit mech typically accumulate in its first week?).

---

Will apply these as spec edits once DIvya's review is also resolved — doing it as one commit per concern so each is reviewable on its own.

DIvyaNautiyal07:
**#1 (delivery gate rename):** Agreed — "delivery gate" is clearer. Makes sense keeping accuracy awareness out of Layer 1.

**#2 (staleness / daily cadence):** Daily matches — `benchmark_flywheel.yaml` already runs at `0 6 * * *` UTC. Freshness guard (14 days) and divergence threshold (0.10) are reasonable starting defaults.

One note: the CSV currently lands as a GitHub Actions artifact, not an IPFS pin. The spec assumes IPFS-based delivery. We'll need to bridge that gap before Phase 1 — just flagging so it's tracked.

**#3 (phased classifier):** Agree with Phase 1. Polymarket API tags are sufficient to start, Omen on aggregate-only is no regression from today. Reversing #11 and #18 is the right call — sync overhead wasn't justified for an unproven benefit. Decide Phase 2 with data in hand.

**#4 (success metrics):** Solid. Two small notes:
- For "zero NVM-deadlock" — what's the detection mechanism? Grepping agent logs, or an alert? Worth specifying so we actually measure it.
- P&L delta — should we define the baseline window (how many days of pre-rollout data we compare against)?

LGTM to proceed with spec edits once these are settled.

mariapiamo:
[P1] Grace semantics are internally contradictory
grace is defined as bypassing only the delivery gate, but the “revised composition” still admits early and skips tool-type/capability checks.
Refs: TOOL_SELECTION_SPEC.md (line 79), TOOL_SELECTION_SPEC.md (line 255), TOOL_SELECTION_SPEC.md (line 615).

[P2] Omen category handling is contradictory across sections
Phase 1 says Omen stays Bet.category = None, but call-site text says both Polymarket and Omen become non-None post-migration.
Refs: TOOL_SELECTION_SPEC.md (line 286), TOOL_SELECTION_SPEC.md (line 492), TOOL_SELECTION_SPEC.md (line 534), TOOL_SELECTION_SPEC.md (line 628).

[P2] “Precise new-tool promotion policy” is still not fully specified
It is now documented as deferred open question with candidate ideas, not a locked policy.
Ref: TOOL_SELECTION_SPEC.md (line 640).

bennyjo:
Three discussion-opener comments — none intended to block the PR, all intended to align on direction before the spec gets enshrined.

---

## 1. Can we re-anchor the spec against what mech-interact already does?

Reading through Layer 1, a lot of the pseudocode describes mechanisms that already exist in `mech-interact`, and I think the spec would be clearer (and the review easier) if it framed itself as a *delta* against current code rather than a new pipeline.

Things that already exist and aren't credited as such:
- Subgraph-level filtering: `MechsSubgraph.filter_info()` at `models.py:66-75` (metadata validity, `delivery_rate_cap`) and `ignored_mechs` baked into the GraphQL query at `queries/mechs_info.py:33`.
- Tool-level filter: `relevant_tools = set(res) - self.params.irrelevant_tools` at `behaviours/mech_info.py:96`.
- Tool-supporting filter: `relevant_mechs_info` at `states/base.py:313-317` already skips mechs that don't support the chosen tool.
- Composite ranking: `MechInfo.__lt__` at `states/base.py:221-240` already blends `delivery_rate_metric`, `liveness` (half-life 3600s, `base.py:154-165`), and `delivered_ratio_smoothed` (Laplace `+8/+9`, `base.py:267-271`), karma as tiebreak.
- Top-1 vs full ranking: `priority_mech` (`states/base.py:327-333`), `ranked_mechs` / `ranked_mechs_addresses` (`states/base.py:346-363`), and `get_priority_mech_address()` at `behaviours/request.py:554-573` which already iterates the ranking and skips penalized mechs.

The parts that are genuinely new are worth calling out explicitly: `allowed_payment_types` + cached `payment_type` on `MechInfo`, the IPFS `category` field and the tool-type gate, the hard `min_mech_score` threshold that converts existing soft ranking into hard exclusion, and the category-keyed `EGreedyPolicy` store. Everything else reads to me as tuning or renaming of existing behaviour.

One specific tension I'd like the authors' take on: the spec keeps Laplace for ranking but argues Wilson for the gate, on the grounds that Laplace is "wrong for a gate" because it gives new mechs an 89% prior. But the existing ranking gives new mechs an 89% prior *on purpose* — that's a grace-by-prior pattern. Are we sure two different priors on the same underlying signal is what we want, or is the grace period an attempt to paper over having picked different ones?

Suggestion: a pass over the spec that identifies each change as "modify X at `file:line`" would make the scope much more legible, and might surface places where the existing hook is already enough.

---

## 2. Should we be moving away from `irrelevant_tools` entirely?

I'd like to open a discussion on this before we lock it in as a fourth gate.

My concern: in an open mech marketplace, a name-based blocklist is structurally chasing its tail — any actor can spin up a new tool under a new name, and the list requires human curation that doesn't scale with the marketplace. It also couples trader *policy* with tool *naming*, which the marketplace controls.

My instinct is that traders should be able to figure out what tools to use themselves, rather than being told explicitly what to use or avoid. The spec's own split hints at this:
- "Wrong-tool-type exclusions" is arguably what `task_type` + capability declarations are *for*.
- "Within-type inferior variants" is arguably what Brier ranking is *for* — and if it can't out-compete a manual blocklist, that's a signal Layer 2 is underpowered rather than that we need a blocklist.

Some alternatives worth discussing:
- Automatic quarantine triggered on measured behaviour (Brier threshold, calibration drift) rather than on name.
- Richer capability declarations the trader filters on (latency class, cost ceiling, …) so the filter is on *what a tool does*, not *what it's called*.
- Bandit algorithms that explore less on poor arms (Thompson / UCB) so bad tools naturally receive less traffic without a kill switch.

Questions I'd like the authors' take on:
1. Are there cases today where you think *only* a name blocklist can solve the problem, and measured-performance signals can't?
2. If we treated `irrelevant_tools` as deprecated-but-tolerated (empty by default, kept as an emergency override), what would break?
3. Is there appetite to make "replace each current `irrelevant_tools` entry with a measured-signal equivalent" a tracked workstream?

Not asking to block the PR on this — the spec can ship with `irrelevant_tools` described as a transitional mechanism. But I'd like us to land on a shared view of the *direction* before the gate gets enshrined as permanent.

---

## 3. How automated can we make adversarial robustness?

The spec is honest that the delivery gate doesn't catch a mech that *responds reliably but answers badly* (e.g. declares `category: prediction` and returns `p_yes = 0.5` for everything, or returns confident-but-miscalibrated outputs). The current story is: Layer 3 Brier catches it eventually, then a human adds it to `irrelevant_tools`. In an open marketplace that loop has a human in it, and I'd like us to discuss how much of it we can close automatically.

Some directions worth considering, in rough order of how far they go:
1. **Automatic per-`(tool, topic)` quarantine** — once a cell has `n ≥ N_min` and Brier crosses a threshold, quarantine it the same way `consecutive_failures` / `quarantine_duration` already quarantines in `EGreedyPolicy`. No human in the loop, same primitive.
2. **Calibration-aware signals** — Brier decomposes into reliability + resolution; a tool whose confidence is systematically uncorrelated with outcome is distinguishable from one that's just noisy. Useful for telling "bad tool" apart from "hard market."
3. **Bandit that stops exploring consistently-bad arms** — eps-greedy keeps sending traffic to known-bad tools at rate `eps`. Thompson sampling / UCB concentrate exploration where uncertainty is high, which also limits how much damage a newly-adversarial arm can do before the signal catches up.
4. **Mech-level contagion concern (already flagged OOS)** — an adversary can ship a malicious tool under an established mech's reputation halo, and the gate is mech-level. Worth at least documenting as a known open-marketplace vector even if the fix is deferred.

Questions for the authors:
1. What's the fastest honest-but-reasonable Brier signal you think we could trust for auto-quarantine? (i.e. what's `N_min` and the threshold you'd be comfortable flipping on?)
2. Is the attachment to epsilon-greedy load-bearing, or would you be open to evaluating UCB/Thompson in Layer 2?
3. How do we feel about tool-level (rather than mech-level) reputation as a follow-up — is the subgraph schema change the real blocker, or is there appetite?

Framing this as design conversation, not a block. But I'd like the spec to state a direction on automation even if v1 ships with humans in the loop, so we're not relying on manual kill switches indefinitely.

LOCKhart07:
Attack vector: Someone starting a mech with same tool name as top performing ones. Ideally accuracy mapped as tool+hash. But every tool update voids previous gains for that tool

bennyjo (changes requested):
We'll have another go at speccing tool selection properly. This draft PR and it's comments can serve as useful context. But this spec, as is, is not what we'll go ahead with.

PR #903, opened 2026-04-10
fix: gate Omen 'invalid' label on Reality.eth finalization (ZD#919)

LOCKhart07:
## Summary

The Omen status pipeline labelled any market whose subgraph `currentAnswer` equalled the `0xff..ff` invalid sentinel as `"invalid"`, with no check that Reality.eth's 24-hour dispute window had closed. Reality answers can flip during that window, so a temporary sentinel — even one later overwritten on-chain — would stick in the cached snapshot and surface to users as "invalid market" forever. The same pre-finalization terminal labelling also affected `won`/`lost` classification and accuracy aggregates.

This PR adds a Reality.eth finalization gate keyed off `answerFinalizedTimestamp`, plus an `isPendingArbitration` gate for markets escalated to Kleros. Any answer is treated as `pending` until the dispute window has closed (and arbitration is not active); only then can it become `invalid`, `won`, or `lost`.

**No funds affected** — purely the displayed status. The redeem path is unaffected: it already gates on `answerFinalizedTimestamp` via a separate query in `market_manager_abci`.

## Subgraph topology

`answerFinalizedTimestamp` and `isPendingArbitration` are **not exposed** by `olas_agents_subgraph` (the `predict-omen` proxy) — its fpmm entity is `FixedProductMarketMakerCreation`, which lacks both fields. Initially this PR tried to select `answerFinalizedTimestamp` from olas queries directly and would have 500'd in production (caught by @mariapiamo's review and live introspection — see https://github.com/valory-xyz/autonolas-subgraph-studio/issues/114).

The fix sources both fields from `omen_subgraph` (`https://omen.subgraph.autonolas.tech`, the Protofire-lineage Omen xDai subgraph proxied by Olas infra; Graph-Network id `9fUVQpFwzpdWS9bq5WkAnmKbNNcoBwatMR4yZq81pbbz`). One batched second-hop GraphQL query per refresh cycle, keyed by `id_in: [<fpmm_ids>]` and chunked at 1000 (the Graph hard cap). No new infrastructure — `omen_subgraph` is already declared at `trader_pearl/service.yaml:203-213`.

## What changed

1. **`graph_tooling/queries.py`** — stripped `answerFinalizedTimestamp` from `GET_TRADER_AGENT_BETS_QUERY`, `GET_PREDICTION_HISTORY_QUERY`, and `GET_SPECIFIC_MARKET_BETS_QUERY` (it doesn't exist on that schema). Added `GET_OMEN_FINALIZATION_QUERY` for the `omen_subgraph` enrichment hop.
2. **`graph_tooling/predictions_helper.py`**
   - `_fetch_finalization_by_fpmm_ids(ids)` — batched omen query with chunking; degrades silently on network/HTTP/JSON errors so missing enrichment falls through to "pending."
   - `_enrich_bets_with_finalization(bets)` — collects unique fpmm ids from the olas response, calls the fetcher, mutates each `bet["fixedProductMarketMaker"]` in place with both fields. Wired into `fetch_predictions` and `_fetch_bet_from_subgraph`.
   - Gate logic in `_get_prediction_status`, `_calculate_bet_net_profit`, and `_build_market_context` extended: `isPendingArbitration=True` returns `PENDING` / `(0.0, None)` regardless of finalization timestamp.
3. **`behaviours.py`** — `_get_prediction_accuracy` now invokes enrichment on the Omen path before computing accuracy. `_calculate_omen_accuracy` filters out `isPendingArbitration` bets.
4. **`models.py` + `skill.yaml`** — added `OmenSubgraph(Subgraph)` model and the `omen_subgraph` declaration block (mirrors the same model already used by `market_manager_abci`).

## Mitigations and known shortcomings

- **Cache stale-data caveat.** `shared_state.performance_summary.prediction_history` stores classified status strings. Currently-affected users won't see the fix until `FetchPerformanceDataRound` runs again (one FSM cycle for normally-running agents — minutes; indefinite for agents gridlocked at `MechRequestRound` due to low Safe balance, see Bug B).
- **Subgraph indexer lag** (10–60s typical) on either subgraph degrades to `pending`, the safer label.
- **Two-subgraph dependency.** Either subgraph being down → bets without enrichment default to `{answerFinalizedTimestamp: None, isPendingArbitration: False}` → status = `pending`. Strictly safer than terminal mis-classification.
- **Latency:** +1 sequential GraphQL POST per refresh cycle (~100–300ms typical). On a background round, invisible.
- **Long-term cleanup:** if `answerFinalizedTimestamp` is ever added natively to the predict subgraph (tracking issue: https://github.com/valory-xyz/autonolas-subgraph-studio/issues/114), the second hop can be removed.
- **CT `payoutDenominator` not checked.** Reality.eth finalization does not strictly imply ConditionalTokens has been told via `RealitioProxy.resolve()`. Brief window — typically one trader cycle — where the displayed status is terminal but the on-chain CT condition is still unresolved. Documented inline at `_get_prediction_status`. Subgraph treated as source of truth.
- **Polymarket helper untouched.** Its schema gates on `question.resolution` being non-null, which the subgraph only populates once the market is fully resolved.
- **Accuracy numbers may shift slightly** on the next snapshot. The buggy code excluded sentinel-bearing pending bets via the `INVALID_ANSWER_HEX` filter; the new code excludes them more uniformly (treating them as pending, not invalid). The new value is correct.
- **Defensive malformed-hex parse.** `parse_current_answer` returns `None` on malformed input; callers degrade to `pending` / skip.

## Visibility vs redeem

Bug A is purely a display fix. The user reporting ZD#919 is also affected by **Bug B** — an FSM gridlock where the agent's Safe dropped below the mech request price. Bug B is tracked separately and out of scope here.

## Workaround for currently-affected users

Top up the Safe with ~0.05 xDAI. One mech request succeeds, the FSM cycle proceeds to `FetchPerformanceDataRound`, the snapshot refreshes with correct labels, and `RedeemRound` runs to claim recoverable winnings.

## Test plan

- [x] All 6 tomte checks pass: black, isort, flake8, mypy, pylint, darglint
- [x] 966 tests pass in `agent_performance_summary_abci` (was 935 on main; +31 new)
- [x] 100% coverage on every modified production file (`behaviours.py`, `graph_tooling/predictions_helper.py`, `graph_tooling/queries.py`, `models.py`)
- [x] Hashes locked via `autonomy packages lock`
- [x] **Live wire test** against `omen.subgraph.autonolas.tech` using the exact `GET_OMEN_FINALIZATION_QUERY` body. Repeatable:
  ```bash
  curl -sS -X POST -H 'Content-Type: application/json' \
    -d '{"query":"query GetOmenFinalization($ids: [ID!]!) { fixedProductMarketMakers(where: { id_in: $ids }, first: 1000) { id answerFinalizedTimestamp isPendingArbitration } }","variables":{"ids":["0xc44a38419c33b70c904d7d51b20e47cfc2223def"]}}' \
    https://omen.subgraph.autonolas.tech
  ```
  Returns non-null `answerFinalizedTimestamp`, `isPendingArbitration: false`, `hasIndexingErrors: false`. Sample row has `answerFinalizedTimestamp − currentAnswerTimestamp = 86400`s — Reality.eth's 24h timeout applied at index time.
- [x] **Live agent run (Safe `0x5461768d1C2ce52E807Ff76c793584401300Fb80`, 148 historical Omen bets)**: `fetch_performance_data_round` completes with `Event.DONE`; zero `answerFinalizedTimestamp` subgraph errors; `_calculate_omen_accuracy` produces a value (no `"Omen accuracy unavailable"` warning, which would fire iff enrichment silently failed). End-to-end exercise of the olas→omen enrichment hop.
- [ ] Manual: confirm the Reality.eth dispute-window edge case by triggering a snapshot refresh on a market currently inside its 24-hour window — should appear as `pending` until `answerFinalizedTimestamp` passes
- [ ] Manual (new): if any historical bet sits on a market with active Kleros arbitration, confirm it appears as `pending` regardless of `currentAnswer`

## Falsifiability

Falsifiable wiring tests pin the enrichment integration in `TestEnrichmentWiredIntoFetchPaths` (predictions_helper) and `test_omen_platform_invokes_enrichment_before_accuracy` (behaviours): removing the `_enrich_bets_with_finalization` call from any of the three fetch sites causes `mock_enrich.assert_called_once()` to fail. Multi-day contested-market tests (`test_status_pending_during_multi_day_bond_war`, parametrized `test_status_pending_during_extended_arbitration` over [1, 7, 14, 30, 90] days) prove the gate respects future finalization regardless of duration. Arbitration-exclusion test on accuracy is bidirectionally falsifiable: 50% without the gate → 100% with.


DIvyaNautiyal07 (dismissed):

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -746,14 +779,24 @@ def _calculate_bet_net_profit(
         if current_answer is None:
             return 0.0, None
 
+        # Bug A (ZD#919): Reality.eth answers can flip during the dispute
+        # window, so any answer (including the invalid sentinel) must wait
+        # for finalization before becoming terminal.
+        if answer_finalized_ts is None or int(answer_finalized_ts) > self._now():
```
bennyjo: Two small robustness thoughts on this line (same pattern at line 841 and `behaviours.py:542`):

1. `_parse_current_answer` was added to harden the `currentAnswer` cast, but `int(answer_finalized_ts)` here has no equivalent guard. A malformed timestamp from the subgraph would raise `ValueError` and surface as a 500 on the predictions endpoint — the same failure mode the new helper defends against on the adjacent field. Would you consider a symmetric `_parse_timestamp` helper (or inline `try/except`) so both subgraph fields degrade to "pending" the same way?
2. Have you checked whether the Omen subgraph ever returns `"0"` for unset `answerFinalizedTimestamp` rather than `null`? If it does, a bet with a set `currentAnswer` + `answer_finalized_ts == "0"` would slip past this gate as `0 > now` is false, and get classified as terminal. An explicit `<= 0` check would close it — WDYT?

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -513,28 +515,51 @@ def _get_prediction_accuracy(self) -> Generator[None, None, Optional[float]]:
...
+        # Reality.eth answers can flip during the dispute window, so a bet
+        # whose market has currentAnswer set but answerFinalizedTimestamp
+        # in the future is still provisional and must be excluded.
+        bets_on_finalized_markets = []
+        for bet in bets:
+            fpmm = bet.get("fixedProductMarketMaker", {})
+            if fpmm.get("currentAnswer") is None:
+                continue
+            finalized_ts = fpmm.get("answerFinalizedTimestamp")
+            if finalized_ts is None or int(finalized_ts) > now:
```
bennyjo: The new filter silently drops every bet if `answerFinalizedTimestamp` is ever missing from the response (indexer skew, partial response, future schema change) — `bets_on_finalized_markets` goes empty and accuracy returns `None` with no log signal. Would it be worth a `logger.warning` when `len(bets) > 0` but `len(bets_on_finalized_markets) == 0`, so a future subgraph regression shows up as a log line rather than accuracy silently vanishing from the UI?

Tangentially related: the PR description notes `shared_state.performance_summary.prediction_history` is not invalidated, so historical cached entries won't carry the field either and will degrade to pending until the next `FetchPerformanceDataRound`. Is a schema-version bump on the cache key something you'd want to fold in here, or keep strictly out of scope for a follow-up?

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -48,6 +49,26 @@
 DEFAULT_CURRENCY = "USD"
 
 
+def _parse_current_answer(value: Optional[str]) -> Optional[int]:
```
bennyjo: Small naming question: leading underscore signals module-private, but `behaviours.py` imports this symbol directly. Would you prefer to drop the underscore (it's effectively public across the package now) or move it to a shared util module? Current form can trip linters configured to flag cross-module private-name imports.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -513,28 +515,51 @@ def _get_prediction_accuracy(self) -> Generator[None, None, Optional[float]]:
         else:
             return self._calculate_omen_accuracy(agent_bets_data)
 
+    def _now(self) -> int:
```
bennyjo, line 526: `PredictionsFetcher._now` at `predictions_helper.py:119` is the same body/docstring/rationale. Two copies of a time source can drift subtly (e.g. one gets patched in tests to return ms, the other seconds). Would you consider extracting a single module-level `_now()` in a shared util that both classes call, or is the duplication deliberate to keep each class independently patchable?

bennyjo (commented):
Inline comments on brittleness points and the accuracy regression — suggestions, not blockers.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Both accepted — fixed in c5714cb0.

1. Added a symmetric `parse_timestamp` helper in `predictions_helper.py` that handles the same failure modes as `parse_current_answer`: `None`, malformed input, and — per your second point — the `"0"` / `0` / negative "unset" sentinel. Returns `Optional[int]`; callers degrade to pending instead of crashing or silently slipping past the gate.

2. You're right that the `"0"` case would slip through — verified by the now-passing test `TestFinalizationGateHardening::test_status_zero_finalization_returns_pending`. I'd rather not speculate on whether the Omen subgraph emits `"0"` vs `null` for unset fields today (the current schema uses `null`, but that's not a stable contract across deploys/versions), so I pinned it defensively.

All three gate sites (`predictions_helper.py:785`, `:841`, `behaviours.py:542`) now route through `parse_timestamp` and share the same failure semantics.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
LOCKhart07: Warning: accepted in c5714cb0. `_calculate_omen_accuracy` now logs a warning when bets exist *and* have a resolved `currentAnswer` but none are finalized — that's the actual regression signal (indexer skew / schema drift). All-unresolved is the normal pre-resolution state so it intentionally stays silent. Covered by `test_warns_when_resolved_bets_have_no_finalization` and `test_no_warning_when_all_bets_unresolved`.

Cache invalidation: deliberately not doing this — the concern doesn't apply here. `prediction_history` isn't an incremental cache with retained stale entries; `_fetch_prediction_history` (`behaviours.py:967`) does a fresh subgraph fetch each `FetchPerformanceDataRound` and `_save_agent_performance_summary` (line 1756+) overwrites the field wholesale. There's no key to bump because the snapshot is fully replaced every cycle. The only lag is "until the next round runs", which the PR description already calls out as the expected user-visible behaviour.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Accepted — renamed in c5714cb0. `_parse_current_answer` is now `parse_current_answer` (no underscore), matching its actual cross-module visibility. Callers in `behaviours.py` and both test files updated. Added `TestParseCurrentAnswerPublicName` to pin the exported name.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
LOCKhart07: Accepted — deduplicated in c5714cb0. Added a module-level `now_ts()` in `predictions_helper.py` as the single real time source. Both `PredictionsFetcher._now` and `FetchPerformanceSummaryBehaviour._now` are now thin delegators that call `now_ts()` — so they can't drift (seconds vs ms, etc.). Kept them as instance methods so existing `patch.object(instance, "_now", ...)` test patterns still work. New test `TestNowTsModuleLevel::test_fetcher_now_delegates_to_module` patches the module-level symbol to prove the delegation is real.

mariapiamo:
P1 regression: queries.py (line 393) omits answerFinalizedTimestamp, but predictions_helper.py (line 840) now requires it and predictions_helper.py (line 608) uses that status path for specific bet details.
P2 test gap: test_queries.py (line 167) doesn’t cover GET_SPECIFIC_MARKET_BETS_QUERY, so this regression is not caught.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -513,28 +516,68 @@ def _get_prediction_accuracy(self) -> Generator[None, None, Optional[float]]:
...
         won_bets = 0
         total_bets = 0
 
-        for bet in bets_on_resolved_markets:
+        for bet in bets_on_finalized_markets:
             market_answer = bet["fixedProductMarketMaker"]["currentAnswer"]
             bet_answer = bet.get("outcomeIndex")
             if market_answer == INVALID_ANSWER_HEX or bet_answer is None:
                 continue
+            correct = parse_current_answer(market_answer)
```
bennyjo, line 587: Optional nit: since `parse_current_answer` returns `None` for `INVALID_ANSWER_HEX`, the `market_answer == INVALID_ANSWER_HEX` check on line 573 is now redundant with the `correct is None` guard right below — both paths `continue` without incrementing `total_bets`. Dropping the sentinel check (and letting `parse_current_answer` handle sentinel + malformed uniformly) would make the intent clearer. Happy to leave as-is if you prefer the explicit sentinel signal at the top of the loop.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -746,14 +823,26 @@ def _calculate_bet_net_profit(
...
                 return 0.0, None
             refund_share = total_payout * (bet_amount / total_traded)
             return refund_share - bet_amount, refund_share
 
-        correct_answer = int(current_answer, 0)
+        correct_answer = parse_current_answer(current_answer)
+        if correct_answer is None:
+            # Malformed currentAnswer (not None, not sentinel, not parseable).
+            # Treat as unresolved rather than crash.
+            return 0.0, None
```
bennyjo, line 967: Optional nit: `_get_prediction_status` logs a warning on the same malformed-`currentAnswer` condition (see line 899), but this branch degrades silently. Asymmetric observability — a subgraph data-quality regression would surface in one helper and not the other. Would you consider a symmetric `self.logger.warning(...)` here so the failure mode is diagnosable from logs?

bennyjo (dismissed):
Two optional nits — approving.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
LOCKhart07: Accepted — dropped the sentinel check in 0df4a147. The `bet_answer is None` guard had to stay (no equivalent downstream), but the `market_answer == INVALID_ANSWER_HEX` half is redundant because `parse_current_answer(INVALID_ANSWER_HEX)` returns `None` (predictions_helper.py:83), so the `correct is None: continue` branch right below already excludes sentinel bets. Inline comment updated to explain that the `None` branch intentionally covers both sentinel and malformed. Existing `test_invalid_answer_hex_excluded_from_denominator` and `test_all_invalid_returns_none` pin the invariant — both still pass.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Accepted — added symmetric warning in 0df4a147. `_calculate_bet_net_profit` now logs `Malformed currentAnswer for bet <id>: <value>` on the same condition as `_get_prediction_status`. Pinned by new falsifiable test `test_malformed_current_answer_logs_warning` (red without the warning, green with). A subgraph data-quality regression will now surface in both helpers rather than only one.

DIvyaNautiyal07 (commented):
Thorough fix with solid test coverage (+31 tests, 100% on modified files). The core finalization gate, arbitration handling, and degradation path are all correct. Verified: all three entry points are enriched, no stale `INVALID_ANSWER_HEX` references outside the gated paths, boundary condition (`finalized_ts == now`) is consistent between `_calculate_omen_accuracy` and `_get_prediction_status`, and `GET_OMEN_FINALIZATION_QUERY` `first: 1000` aligns with `OMEN_ID_IN_CHUNK`. No blocking issues — one minor P2 inline.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -510,31 +513,86 @@ def _get_prediction_accuracy(self) -> Generator[None, None, Optional[float]]:
...
-        else:
-            return self._calculate_omen_accuracy(agent_bets_data)
+
+        # ZD#919: enrich Omen bets with Reality.eth finalization data from
+        # omen_subgraph before computing accuracy. The olas_agents subgraph
+        # does not expose answerFinalizedTimestamp; without enrichment
+        # every bet would be filtered out as unfinalized and accuracy
+        # would silently disappear from the UI.
+        fetcher = PredictionsFetcher(self.context, self.context.logger)
+        fetcher._enrich_bets_with_finalization(agent_bets_data["bets"])
```
DIvyaNautiyal07, line 523: P2 (consider): This creates a full `PredictionsFetcher` (with `predict_url`, `mech_url`, `omen_url`) just to call one method. Consider making `_enrich_bets_with_finalization` a public module-level function or `@staticmethod` that accepts `(omen_url, logger, bets)`, which would also resolve the private-method cross-module call (`fetcher._enrich_bets_with_finalization`). Not urgent — once-per-cycle cost is negligible.

DIvyaNautiyal07 (dismissed):

bennyjo (dismissed):
Approving. The major rewire to source `answerFinalizedTimestamp` and `isPendingArbitration` from `omen_subgraph` is sound — the olas_agents subgraph does not expose those fields on its `FixedProductMarketMakerCreation` entity, so the pre-rewire query would have 500'd in prod (unit tests passed only because fixtures fabricated the field). Implementation covers it: new `GET_OMEN_FINALIZATION_QUERY` chunked at 1000, `_enrich_bets_with_finalization` wired into all three olas fetch paths, `isPendingArbitration` gate propagated through status/profit/accuracy and `_build_market_context`, defensive degradation to `pending`. Tests cover it: falsifiable wiring tests at each call site, chunking (1500→1000+500), error paths, parametrized arbitration durations, bidirectionally falsifiable accuracy test, schema pin on the new query. Two non-blocking observations for follow-up: (1) `OmenSubgraph` ApiSpecs model is declared but `_fetch_finalization_by_fpmm_ids` uses raw `requests.post`, bypassing the declared `retries: 5`; (2) mocked-fixture test pattern that let the original bug through is unchanged — a live introspection/smoke check would make it self-correcting.

bennyjo (approved):
Re-approving. Previous approval was auto-dismissed by the mechanical hash/yaml bump in c8cfadd5 (merge + autonomy packages lock); no code changes to the fix since. All review rounds addressed, CI green across the full matrix.

DIvyaNautiyal07 (approved):

packages/valory/skills/agent_performance_summary_abci/behaviours.py
LOCKhart07: Not fixing as benefits negligible

PR #83, opened 2023-09-19
fix: optionally call `claimWinnings`

0xArdi:
This PR adds support for only calling `claimWinnings` when the `history_hash` of the question is not null. The `history_hash` of a question can be null in cases when `claimWinnings` has been called by somebody else, and all the answer history has been provided.

PR #904, opened 2026-04-10
fix: always redeem winnings before mech requests or new positions

LOCKhart07:
## Summary

- **Restructure the trader FSM so every cycle attempts redemption before any mech request or bet placement**, eliminating the death-spiral where a low-balance safe with unredeemed winning positions can never reach the redeem flow that would refund itself.
- **Add `PostBetUpdateRound`**, a small dedicated round that runs after Omen `BetPlacementRound` / `SellOutcomeTokensRound` tx settlement to restore the post-bet local-state bookkeeping (`update_bet_transaction_information`) that was previously a side effect of the post-bet `RedeemBehaviour.async_act` and is no longer reachable under the new routing.
- **Revert a 2.5-year-old vestigial `redeeming_batch_size=1` service override** that was a band-aid for a Reality.eth bug fixed within a week of the band-aid landing — without this revert, the new always-redeem-first flow would only drain 1 unredeemed position per cycle on Pearl Quickstart, taking N cycles to recover from a backlog of N positions.
- **Polishing**: drop a stale "sell outcome tokens" mention from the new round's user-facing description; document the `PostBetUpdateBehaviour` divergence from the legacy hook; add a defensive `did_transact` guard.

> Targets `fix/zd919-invalid-market-finalization-gate` so it stacks on top of the in-flight ZD#919 fix.

## Background — the death spiral

A trader agent got stuck:

- **~3.85 xDAI** of unredeemed winning positions on-chain (5 markets, oldest resolved 4+ days ago).
- **0.005 xDAI** liquid in the safe — below the 0.01 xDAI mech fee.
- The FSM only reached the redeem flow **after** completing a mech request, so `MechRequestRound` looped forever (`"balance not enough"` warning every 15 min, 204 occurrences in the incident logs) and never got to the redeem step that would have refunded itself.

David's directive: *"Make redeeming always happen first before anything else. All available redemptions should be made before any new positions are taken."*

## Commits

| Commit | What |
|---|---|
| `f2d0a2c9` | The FSM restructure itself — 10 chain-level transition rewrites in `composition.py` plus a new `FinishedPolymarketBetPlacementRound` final state to cover Polymarket's off-chain bet path. |
| `fa1e0c69` | Adds `PostBetUpdateRound` + behaviour + tests to restore post-bet bookkeeping for Omen `BetPlacementRound` / `SellOutcomeTokensRound`, which the FSM restructure had inadvertently broken. |
| `18bee280` | Description-only fix on the new round to drop the stale sell-outcome-tokens mention. |
| `5564b9a7` | Audits and documents the `PostBetUpdateBehaviour` dispatch divergence from the legacy `RedeemBehaviour` hook (review finding 1). |
| `ac8da0ab` | Adds a defensive `did_transact` guard to `PostBetUpdateBehaviour` (review finding 2) plus a TDD test for the stale-state scenario. |
| `0747ff6c` | Reverts the 2.5-year-old vestigial `REDEEMING_BATCH_SIZE:int:1` service override on `trader_pearl` and `polymarket_trader` back to the skill default of `5`, since the underlying Reality.eth claim bug it was a band-aid for was fixed in PR #83 four days after the band-aid landed. |

## Target FSM (after the change)

```
MarketManager -> RedeemRouter (early redeem)
                  -> Redeem / PolymarketRedeem
                      -> CheckStopTrading
                  -> [trade decision -> Mech -> BetPlacement]
                      -> PostBetUpdateRound (Omen on-chain wrap-up)
                  -> CallCheckpoint -> ResetPause -> loop
```

After the change, `RedeemRouterRound` has **exactly two incoming edges** in the flattened FSM (`UpdateBetsRound DONE` and `PolymarketFetchMarketRound DONE`) — both intentional cycle entries. The death-spiral scenario is unreachable because redemption now runs before the mech is ever consulted.

## Coverage — both Omenstrat and Polystrat

| Path | Routing |
|---|---|
| Omen cycle entry | `UpdateBetsRound DONE` -> `RedeemRouterRound` |
| Polymarket cycle entry | `PolymarketFetchMarketRound DONE` -> `RedeemRouterRound` |
| Omen redeem with tx | `RedeemRound DONE` -> tx settlement -> `REDEEMING_DONE` -> `FinishedRedeemingTxRound` -> `CheckStopTradingRound` |
| Polymarket on-chain redeem | `PolymarketRedeemRound PREPARE_TX` -> tx settlement -> `REDEEMING_DONE` -> same exit (multiplexer routes by tx_submitter) |
| Polymarket off-chain redeem (DONE/MOCK_TX) | `PolymarketRedeemRound DONE` -> `FinishedPolymarketRedeemRound` -> `CheckStopTradingRound` |
| No positions to redeem | `Redeem*Round NO_REDEEMING` -> `FinishedWithoutRedeemingRound` -> `CheckStopTradingRound` |
| Omen post-bet wrap-up | `BetPlacement` -> tx -> `BET_PLACEMENT_DONE` -> `PostBetUpdateRound` -> `CallCheckpointRound` |
| Polymarket post-bet wrap-up | `PolymarketBetPlacementRound DONE` -> `FinishedPolymarketBetPlacementRound` -> `CallCheckpointRound` (Polymarket bookkeeping is inline, no PostBetUpdate needed) |

Internal `BlacklistingRound DONE` was rerouted to `FinishedWithoutDecisionRound` to preserve the legacy "end the cycle on blacklist" semantics under the new routing.

## Why the post-bet bookkeeping needed `PostBetUpdateRound`

`RedeemBehaviour.async_act` (`reedem.py:952-964`) was the post-tx-settlement hook that called `update_bet_transaction_information` (or `update_sell_transaction_information`) when `tx_submitter` was `BetPlacementRound` or `SellOutcomeTokensRound`. With the new routing, `RedeemBehaviour` no longer runs after a successful Omen bet/sell tx. `tx_submitter` / `did_transact` / `sampled_bet_index` are not in `cross_period_persisted_keys`, so they're cleared at period reset before the next cycle's early redeem can react.

Without the bookkeeping:
- The bet's `queue_status.next_status()` is never called -> the same bet may be re-sampled
- `processed_timestamp` and `bet.strategy` are never set
- `invested_amount` tracking diverges from the actual on-chain investment

Polymarket bets are unaffected — `PolymarketBetPlacementBehaviour` does the bookkeeping inline (`polymarket_bet_placement.py:148,156`).

`PostBetUpdateRound` is a small dedicated round that runs immediately after the multiplexer's `BET_PLACEMENT_DONE` / `SELL_OUTCOME_TOKENS_DONE` exits. Its behaviour reads `tx_submitter`, dispatches to the appropriate bookkeeping helper, and emits a `PostBetUpdatePayload` to advance the round so the cycle can wrap up via `CallCheckpointRound`. The existing `RedeemBehaviour` hook is intentionally left in place as a defensive no-op.

The new `PostBetUpdateBehaviour` dispatches to `update_bet_transaction_information` for `BetPlacementRound` and `update_sell_transaction_information` for `SellOutcomeTokensRound` — this is an intentional correction of a legacy outlier (the legacy `RedeemBehaviour` hook collapsed both submitters to the bet helper). Audit confirms 3 of 4 existing inline call sites already dispatch by tx type; only the legacy `RedeemBehaviour` post-tx hook lazily reused the bet helper for sells. See commit `5564b9a7` for the full audit. The behaviour also includes a `did_transact` guard mirroring the legacy pattern, to defend against tendermint replays / restarts that could land mid-round in `PostBetUpdateRound` without a fresh successful tx (commit `ac8da0ab`).

## Why revert the `redeeming_batch_size=1` service override

The skill default for `redeeming_batch_size` has always been `5` (introduced 2023-08-30 by [PR #79's predecessor](https://github.com/valory-xyz/trader/commit/21ecd957) with the documented rationale *"increasing this number equals fewer fees but more chances for the transaction to fail"*).

Two weeks later, [PR #79](https://github.com/valory-xyz/trader/pull/79) ("Temporary fix for redemption", 2023-09-15) overrode the **service-level default** from `5` -> `1` as a band-aid for a different stuck-state: the agent was reverting on Reality.eth questions where a third party had already provided the full answer history. PR #79's body explicitly says it is *"a temporary fix... before a final fix is implemented in an upcoming PR"*.

The "final fix" landed **4 days later** in [PR #83](https://github.com/valory-xyz/trader/pull/83) ("fix: optionally call `claimWinnings`", commit `906e71ce`), which added the `if not self.is_history_hash_null` gate around the `claimWinnings` call. That logic still lives at `reedem.py:744`. The Reality.eth bug has been fixed for 2.5 years.

But nobody ever raised the temporary service override back to `5`. The Pearl Quickstart and `polymarket_trader` services have been redeeming **one Omen position per cycle** ever since.

This matters concretely under the always-redeem-first FSM: an agent with N unredeemed winning positions needs N cycles to fully drain its backlog. With the vestigial `1` cap, the original Slack incident (5 unredeemed positions) would take 5 separate cycles to self-heal under the new flow. With the cap restored to the skill default of `5`, the same backlog drains in 1 cycle.

The underlying tunable is preserved — operators who want the old behavior can still set `REDEEMING_BATCH_SIZE=1` via env var.

## Polymarket has no equivalent cap (and that is structurally fine)

The Omen `redeeming_batch_size` cap exists because each Omen winning position can require **1-3 inner txs** in the multisend (`resolve` on Realitio.eth + `claimWinnings` on Realitio.eth + `redeemPositions` on ConditionalTokens), so a multisend of 5 positions can be up to 15 inner contract calls.

Polymarket redemption is **1 inner tx per position** — either `redeemPositions` on the CTF contract (standard markets) or `redeemPositions` on the NegRiskAdapter (negative-risk markets). There is no Reality.eth equivalent on Polymarket, so there is no resolve+claim+redeem pipeline. `polymarket_reedem.py:_prepare_redeem_tx` walks the entire `redeemable_positions` list with no batch counter, which is structurally defensible.

## Out of scope (intentionally — flagged in the Slack thread for follow-up)

- `MechRequestRound` infinite-loop on insufficient balance (latent bug — unreachable now under always-redeem-first, but worth filing)
- Systematic Yes-bet bias (Kelly strategy `min_oracle_prob: 0.5` filter)
- 30x bet sizing escalation on negative edge (test -> live phase scaling)
- Funding requirements for higher Omenstrat staking contracts
- Document the Polymarket vs Omen redemption-helper divergence in skill code (so a future contributor doesn't unify them by mistake)

## Test plan

- [x] `uv run tomte check-code` — all 6 checks pass (black, isort, flake8, mypy, pylint, darglint)
- [x] `uv run pytest packages/valory/skills/trader_abci/tests/` — 8/8 pass, including the bumped `EXPECTED_TRANSITION_MAPPING_LENGTH = 46`
- [x] `uv run pytest packages/valory/skills/decision_maker_abci/tests/test_rounds.py` — 13/13 pass, including new `test_polymarket_bet_placement_round_transition`, `test_post_bet_update_round_transition`, and updated `test_blacklisting_round_transition` / `test_final_states`
- [x] `uv run pytest packages/valory/skills/decision_maker_abci/tests/behaviours/test_post_bet_update.py` — 4/4 pass (TDD-style; includes `test_async_act_skips_bookkeeping_when_did_not_transact` for the `did_transact` guard)
- [x] `uv run pytest` for all impacted skills (`trader_abci`, `decision_maker_abci`, `check_stop_trading_abci`, `staking_abci`, `market_manager_abci`, `tx_settlement_multiplexer_abci`, `mech_interact_abci`) — 2337/2337 pass
- [x] `uv run autonomy analyse fsm-specs --app-class TraderAbciApp --package packages/valory/skills/trader_abci/` — Check successful
- [x] `uv run autonomy analyse fsm-specs --app-class DecisionMakerAbciApp --package packages/valory/skills/decision_maker_abci/` — Check successful
- [x] `uv run autonomy analyse docstrings --update` — `DecisionMakerAbciApp` docstring auto-regenerated to reflect new transitions
- [x] `uv run autonomy packages lock` — hashes updated for `decision_maker_abci`, `tx_settlement_multiplexer_abci`, `trader_abci`, the `trader` agent, and `polymarket_trader` + `trader_pearl` services
- [x] Programmatic audit: `RedeemRouterRound` now has exactly 2 incoming edges in the flattened FSM, both intentional cycle entries
- [x] Programmatic audit: `ROUNDS_INFO` covers all 48 source rounds in the chained FSM (no `KeyError` risk on `load_rounds_info_with_transitions`)
- [ ] Manual scenario: spin up a local agent with safe balance < `mech_max_delivery_rate` and at least one resolved winning position; confirm the agent reaches `RedeemRouterRound` from `MarketManager` without ever invoking the mech, redeems the winnings, and proceeds normally on the next cycle
- [ ] Confirm the originally-stuck agent in the Slack incident self-recovers within a single cycle on this build (with `redeeming_batch_size=5` restored, a 5-position backlog should drain in 1 cycle)


DIvyaNautiyal07 (dismissed):
The approach is sound and the implementation is correct.

Non-blocking note (benchmarking mode only):
The pre-existing RedeemRound.MOCK_TX → SamplingRound transition now skips CheckStopTradingRound under the new ordering. In the old flow this was fine because CheckStopTrading ran before Redeem; in the new flow it runs after, so the MOCK_TX shortcut bypasses it entirely.
This only affects benchmarking mode (not production), and the transition itself is unchanged by this PR, just something to be aware of if benchmarking behaviour ever needs to respect stop-trading conditions.

mariapiamo:
Non-idempotent post-bet mutation on retries: post_bet_update.py (line 97), rounds.py (line 456), base.py (line 406).
Regression-safety tests are too weak for the new graph rewiring: test_composition.py (line 35).

packages/valory/skills/decision_maker_abci/behaviours/post_bet_update.py
```diff
@@ -0,0 +1,126 @@
...
+            # `tx_settlement_multiplexer_abci`) into it. That means a
+            # tendermint replay or agent restart could land here without a
+            # fresh successful tx in the current period. The
+            # `did_transact` guard mirrors the legacy `RedeemBehaviour`
+            # post-tx hook (`reedem.py:956-957`) and prevents
+            # re-mutating `queue_status` / `processed_timestamp` /
+            # `invested_amount` based on stale `tx_submitter` state.
+            did_transact = self.synchronized_data.did_transact
+            tx_submitter = self.synchronized_data.tx_submitter if did_transact else None
+            if did_transact and tx_submitter == BetPlacementRound.auto_round_id():
```
bennyjo: Optional/cosmetic: the `did_transact and` on lines 99 and 104-106 is redundant — line 98 already sets `tx_submitter = None` when `did_transact` is False, so neither branch can fire in the stale-state case. Could simplify to `if tx_submitter == BetPlacementRound.auto_round_id():` / `elif tx_submitter == SellOutcomeTokensRound.auto_round_id():`. Not a blocker.

bennyjo (dismissed):
LGTM — one optional cosmetic note inline.

packages/valory/skills/decision_maker_abci/behaviours/post_bet_update.py
bennyjo: Current logic is correct.

LOCKhart07:
Addressed both items in c9ce9d99 (merge of base) + fbfed9bb.

**1. Non-idempotent post-bet mutation on retries** — `PostBetUpdateRound` self-loops on `NO_MAJORITY` / `ROUND_TIMEOUT` / `NONE`, so on retry `async_act` re-runs against unchanged `synchronized_data` and would double-advance `queue_status`, double-bump `invested_amount`, and rewrite `processed_timestamp`. Fixed by adding `post_bet_update_applied_tx_hash: Optional[str]` to `SharedState` (`models.py`) and guarding the dispatch in `PostBetUpdateBehaviour.async_act` on `synchronized_data.final_tx_hash`: retries within the same period skip bookkeeping, a fresh settled tx naturally re-applies. Three new tests in `test_post_bet_update.py` cover retry-skip (bet and sell) and new-tx re-apply.

**2. Weak regression tests for graph rewiring** — `test_composition.py` kept the length tripwire (`EXPECTED_TRANSITION_MAPPING_LENGTH = 46`) and added parametrised per-edge assertions for every transition this PR introduces or retargets (`FinishedMarketManagerRound` / `FinishedPolymarketFetchMarketRound → RedeemRouterRound`, `FinishedRedeemingTxRound → CheckStopTradingRound`, `FinishedBetPlacementTxRound` / `FinishedSellOutcomeTokensTxRound → PostBetUpdateRound`, `FinishedPostBetUpdateRound → CallCheckpointRound`), plus a `test_only_expected_edges_enter_post_bet_update` exactness check so a stray future route into `PostBetUpdateRound` trips the test.

All 1198 skill tests pass; `tomte check-code` clean; `autonomy packages lock` run.

LOCKhart07:
@DIvyaNautiyal07 noted on the `RedeemRound.MOCK_TX → SamplingRound` shortcut skipping `CheckStopTradingRound` under the new ordering — leaving it as-is since it only affects benchmarking mode and the transition itself is unchanged by this PR; worth a separate follow-up if benchmarking ever needs to honour stop-trading.

@bennyjo re: the `did_transact and …` redundancy — agreed with your self-follow-up; leaving as-is since the defensive doubling documents intent alongside the new `final_tx_hash` guard and doesn't change behaviour.

mariapiamo (commented):
Thanks for the retry guard and the new transition assertions, this is a solid improvement.

I still see one residual edge case in `post_bet_update.py`:

The idempotency guard only triggers when `final_tx_hash` is non-null and matches the in-memory marker (`post_bet_update_applied_tx_hash`). If `final_tx_hash` is missing (which `PostTxSettlementRound` currently allows for bet/sell done paths) or if the agent restarts, bookkeeping can still be re-applied for the same settled tx, causing double mutation of local bet state (`queue_status`, `invested_amount`, `processed_timestamp`).

Would you consider hardening this by either:
1. Enforcing `final_tx_hash` for bet/sell settlement exits before entering `PostBetUpdateRound`, or
2. Persisting the “already applied” marker in synchronized state (restart-safe), keyed by tx identity.

bennyjo (dismissed):
Ready to merge. All 22 CI checks pass; idempotency guard and composition-test tripwire address the prior review concerns.

LOCKhart07:
Thanks Pia — dug into both points carefully. Reply is long because I want to show the reasoning, not just the conclusion.

**On point 2 (restart safety of the in-memory marker).** I tried to break this under the service's actual restart model and couldn't, because the window closes before the marker matters. On agent crash/restart, the FSM re-enters at `AgentRegistrationAbciApp` rather than resuming mid-round, and `final_tx_hash` isn't in `DecisionMakerAbciApp.cross_period_persisted_keys` (rounds.py:488-499 — the set is `available_mech_tools`, `policy`, `utilized_tools`, `redeemed_condition_ids`, `payout_so_far`, `mech_price`, `mocking_mode`, `next_mock_data_row`, `agreement_id`), so any post-restart flow arrives at `PostBetUpdateRound` with `final_tx_hash` already cleared — bookkeeping is correctly skipped via the `did_transact` / null-hash branches, no double mutation. The only replay window that exists is the in-period self-loop (`NO_MAJORITY` / `ROUND_TIMEOUT` / `NONE`), which is entirely within a single agent lifetime, so `SharedState.post_bet_update_applied_tx_hash` has matching scope. Persisting the marker in synchronized state would be correct but strictly broader than the risk that actually exists here.

Worth noting: the legacy `RedeemBehaviour.async_act` hook (`reedem.py:940-972` on main) was strictly weaker — it had the `did_transact` + `tx_submitter` guard but no tx-hash idempotency and no marker of any kind, so a plain `NO_MAJORITY` retry (no restart needed) would re-apply bookkeeping. This PR narrows that window rather than widening it.

**On point 1 (null `final_tx_hash` reaching `PostBetUpdateRound`).** Real and worth closing. `PostTxSettlementRound.end_block` logs a warning and still emits `BET_PLACEMENT_DONE` / `SELL_OUTCOME_TOKENS_DONE` when `final_tx_hash is None`, so the round is reachable with a null hash, and the prior guard would set the marker to `None` — comparing `None is not None` fails on retry, re-applying bookkeeping. Tightening the upstream multiplexer invariant is the structurally right fix, but it's a cross-cutting change to a shared round with other consumers, so I've handled it locally in `PostBetUpdateBehaviour` instead:

```python
elif did_transact and settled_tx_hash is None:
    self.context.logger.warning(
        "PostBetUpdateRound reached with did_transact=True but "
        "final_tx_hash is None; skipping bookkeeping to avoid "
        "non-idempotent mutations."
    )
```

Pushed in 3b194941, with `test_async_act_skips_bookkeeping_when_final_tx_hash_missing` asserting the skip and that the marker is never set to `None`. Hardening `PostTxSettlementRound` to refuse emitting the *_DONE events without a hash is worth a separate ticket — happy to file one.

`tomte check-code` green, `autonomy packages lock` run, post_bet_update at 100% branch coverage (31/31 stmts, 10/10 branches).

mariapiamo (approved):

DIvyaNautiyal07 (approved):

Issue #310, opened 2024-08-09
Bug when claiming bonds in the case of multiple answers on reality


kongzii:
[According to Realitio docs](https://realitio.github.io/docs/html/contract_explanation.html#answer-history-entries), only the latest history hash is stored in the question struct:

```
Empty  Bytes32 + Answer + Bond + Answerer + False -> History Hash 1
History Hash 1 + Answer + Bond + Answerer + False -> History Hash 2
History Hash 2 + Answer + Bond + Answerer + False -> History Hash 3 <- stored in Questions struct
```

If multiple answers were submitted, and we query them on subgraph, we can see it:

![Screenshot by Dropbox Capture](https://github.com/user-attachments/assets/b0288a02-7ce7-4635-881c-10d7a6dfa3fa)

All the returned history hashes are the same. [(it's this question)](https://reality.eth.limo/app/#!/question/0x79e32ae03fb27b07c89c0c568f80287c01ca2e57-0x7e6d8446fd70f8b20de6a9925135bee8298bc15cdc5a8632d17bf09307cf1972/token/XDAI)

And that's why the following code https://github.com/kongzii/trader/blob/main/packages/valory/skills/market_manager_abci/graph_tooling/requests.py#L297:

```
answers = [
            {
                "args": {
                    "answer": bytes.fromhex(answer["answer"][2:]),
                    "question_id": bytes.fromhex(answer["question"]["questionId"][2:]),
                    "history_hash": bytes.fromhex(
                        answer["question"]["historyHash"][2:]
                    ),
                    "user": Web3.to_checksum_address(answer["question"]["user"]),
                    "bond": int(answer["bondAggregate"]),
                    "timestamp": int(answer["timestamp"]),
                    "is_commitment": False,
                }
            }
            for answer in raw_answers
]
```

submits invalid hashes if there are multiple different answers and claiming of bonds fails.

gabrielfior:
Here's what I know
-> Subgraph only serves the latest history hash (not all previous history hashes)
-> If there's a second answer, what happens?
a. Second answer is correct - market creator will try to claim bond, unsuccessful (OK, as its answer was wrong)
b. Second answer is incorrect - market creator will try to claim bond, unsucessful (NOT OK, because its answer was correct and claim should have been possible, but history hash will not match the one from the answer)
In conclusion, fetching history_hash from the subgraph is not working for this.

kongzii:
[...]

PR #905, opened 2026-04-11
Add methods to Realitio contract

jmoreira-valory:
Add methods needed for market-resolver to Realitio contract.
Non-breaking change, backwards compatible.


Adamantios (approved):

LOCKhart07:
Does this solve https://github.com/valory-xyz/trader/issues/310 ?

LOCKhart07 (approved):

jmoreira-valory:
> Does this solve #310 ?

@LOCKhart07 The changes in the contract itself don't solve this unless further work is done on the trader. However,

For the trader, this is indeed mostly irrelevant because the trader's primary goal is redeemPositions (get winning bet tokens back as wxDAI) - this only needs resolve(), not claimWinnings (the market-creator/resolver needs claimWinnings, but they already call the method). Moreover if an error occurs on the trader's claimWinnings, it could trigger an exception or something blocking its "useful" business logic.

On this line, we have created recently three self-contained skills on https://github.com/valory-xyz/market-creator/pull/183  (perhaps these 3 skills could be moved to this repository?) to recover (1) LP funds, (2) CT Wining token funds, and (3) Realitio bonds. Any service could simply integrate any of these skills to recover its funds in the Omen ecosystem (for the case of the trader it would only need  (2), as it's only placing bets and recovering CT wining positions - this could be considered a future refactor).

PR #909, opened 2026-04-15
fix(realitio): chunked eth_getLogs in get_claim_params

jmoreira-valory:
## Summary

`RealitioContract.get_claim_params` previously called `eth_getLogs` with the full `[from_block, to_block]` range in a single request. When callers that do **not** chunk themselves pass a wide range, the request exceeds the RPC provider's `eth_getLogs` cap (typically 5k–10k blocks) and the response is either empty or raises an exception that the contract wrapper did not catch — resulting in silent "no events found" and thus unclaimable bonds.

This PR:
- Adds an optional `chunk_size: Optional[int] = None` parameter to `get_claim_params`. When set to a positive integer, the block range is split into windows of at most `chunk_size` blocks, each sent as a separate `eth_getLogs` request, and the results are concatenated.
- Broadens the exception handler to catch `Web3Exception` and `ValueError` (covers both `web3>=7` `Web3RPCError` and legacy-provider RPC errors), returning a diagnostic string via the existing `error` return path instead of letting it escape or degenerate into empty results.
- Keeps the legacy single-call behaviour when `chunk_size` is `None` or `<=0`, so existing callers that chunk themselves (e.g. `decision_maker_abci` — see its `_get_claim_params_via_events` in [reedem.py](https://github.com/valory-xyz/trader/blob/main/packages/valory/skills/decision_maker_abci/behaviours/reedem.py)) are unaffected.
- Passes a fresh `topics` list per chunk, since `get_entries` mutates it by prepending the event signature at index 0 (would duplicate across chunks otherwise).

## Why it's needed

Downstream consumers of this contract that don't chunk at the skill level — specifically `market-resolver`'s `omen_realitio_withdraw_bonds_abci` — call `get_claim_params` with the full Realitio `[createdBlock - 1, updatedBlock + 2]` window. For a question whose bond-escalation + finalization-timeout spans ~93k blocks (typical for 24h+ Realitio timeouts on Gnosis), that blows past the RPC cap and the claim silently fails every cycle. Six markets are currently stuck unclaimable in the market-resolver safe on Gnosis because of this.

With this fix, a consumer can pass e.g. `chunk_size=5000` and the contract iterates the window transparently.

## Test plan

- [x] Added 4 new tests in `test_contract.py`:
  - `test_get_claim_params_chunked` — verifies range is correctly split into windows and entries concatenated
  - `test_get_claim_params_chunk_size_zero_acts_as_unchunked` — verifies `chunk_size <= 0` falls back to single call (backward compat)
  - `test_get_claim_params_rpc_rejection_web3_exception` — verifies `Web3Exception` from eth_getLogs is caught and returned as error
  - `test_get_claim_params_rpc_rejection_value_error` — same for legacy `ValueError`
- [x] All 34 existing contract tests still pass
- [x] Lint pipeline passes: `isort`, `black`, `pylint`, `mypy`, `flake8`, `darglint`, `check-packages`
- [x] `autonomy packages lock` run to refresh hashes (cascades to `decision_maker_abci`, `trader_abci`, agent, services — since they transitively depend on the realitio contract hash)


Adamantios (dismissed):

LOCKhart07 (dismissed):
LGTM — fix is correct, tests cover chunking math, the `None`/`<=0` fallback, and both error paths. Fresh `[question_id]` per chunk correctly avoids `get_entries`'s `topics.insert(0, ...)` mutation.

A few non-blocking follow-ups worth considering:

- **`except (Web3Exception, ValueError)` is very broad** (`contract.py:203`). `ValueError` will also swallow event-decode errors from `get_event_data`, programmer bugs, or malformed inputs and surface them as "eth_getLogs failed", which will be misleading during triage. Consider narrowing the catch or at least logging the original exception with traceback before returning the sanitized string.
- **Error message uses the outer range, not the failing chunk** (`contract.py:204-207`). If e.g. the 11th chunk fails, the message still quotes `from_block` → `to_block` (the caller-supplied range) instead of the chunk window that actually tripped, which hurts debugging. Worth tracking and reporting `(chunk_start, chunk_end)`.
- **Docstring not updated** (`contract.py:174`). Still only says "Filters the `LogNewAnswer` event…"; please mention that the call is chunked by default and returns `{"error": ...}` on RPC failure.

Low-priority PR description nits (bundled):
- The claim that callers like `decision_maker_abci` are unaffected is literally false given the default `chunk_size=5000` — in practice they're fine because the skill-side batch is also ≤5000, but the wording is misleading.
- Lists 4 new tests; the diff actually adds 5 (missing `test_get_claim_params_default_chunk_size_splits_wide_range`).
- Noticeable unrelated reformatting churn in `tests/test_contract.py` (assertion-style flips, context-manager rewrite, import re-order) — presumably pre-commit artefacts, but adds review surface.

LOCKhart07 (approved):

Adamantios (approved):

PR #908, opened 2026-04-14
chore: bump mech interact to solve malformed mech issue

LOCKhart07:
[mech-interact@v0.26.0](https://github.com/valory-xyz/mech-interact/releases/tag/v0.26.0)

```
[2026-04-14 19:32:09,071][INFO] [agent] Entered in the 'mech_information_behaviour' behaviour
[2026-04-14 19:32:12,047] [INFO] [agent] Retrieved mechs' information: [MechInfo(id='1722', address='0x26928d3cf08ec37456d428c0cadb498968570e53', service=Service(metadata=[{'metadata': '0x157d3b106831e2713b86af1b52af76a3ef28c52ae0853e9638180902ebee41d4'}], deliveries=[]), karma=-2208, received_requests=2386, self_delivered=36, max_delivery_rate=10000, relevant_tools=set()), MechInfo(id='2135', address='0xbead38e4c4777341bb3fd44e8cd4d1ba1a7ad9d7', service=Service(metadata=[{'metadata': '0x157d3b106831e2713b86af1b52af76a3ef28c52ae0853e9638180902ebee41d4'}], deliveries=[]), karma=132, received_requests=688, self_delivered=387, max_delivery_rate=10000000000000000, relevant_tools=set()), MechInfo(id='2150', address='0xfacaa9dd513af6b5a79b73353daff041925d0101', service=Service(metadata=[{'metadata': '0x157d3b106831e2713b86af1b52af76a3ef28c52ae0853e9638180902ebee41d4'}], deliveries=[]), karma=49, received_requests=386, self_delivered=164, max_delivery_rate=10000000000000000, relevant_tools=set()), MechInfo(id='2182', address='0xc05e7412439bd7e91730a6880e18d5d587.
[...]

PR #910, opened 2026-04-16
chore: revert mech interact bump (#908)

LOCKhart07:
## Summary
Reverts #908 (commit 61fabcdb), which bumped `mech_interact_abci` to mech-interact [v0.26.0](https://github.com/valory-xyz/mech-interact/releases/tag/v0.26.0).

## Version mapping
- **Before #908 / after this revert**: `mech_interact_abci` → `bafybeied5l5xlygjyo3uju5bw52pidjvbkgdahhuzvahh6qis7kd3ht6ta` (mech-interact [v0.25.0](https://github.com/valory-xyz/mech-interact/releases/tag/v0.25.0))
- **After #908**: `mech_interact_abci` → `bafybeifnwgjflfhagoomgqtnmlymlal5r23qcyesdyrhu5hyw7h6iqhqda` (mech-interact [v0.26.0](https://github.com/valory-xyz/mech-interact/releases/tag/v0.26.0))

This PR takes us back to v0.25.0 (the tagged release that preceded the bump).

## Why revert
#908 was merged to address the Pearl Omen trader silent hang in `mech_information_round`. Subsequent investigation indicates the mech-interact bump is **not** the right place to fix the hang:

- v0.26.0 introduced `_fetch_http_parallel`, per-mech quarantine, and a permanent-error classifier (via valory-xyz/mech-interact#69 / valory-xyz/mech-interact#68 / valory-xyz/mech-interact#66). The Omen/Gnosis hang still reproduces with that code. The 20 s wall-clock deadline inside `_fetch_http_parallel` never fires during the wedge — i.e. the behaviour generator simply stops being ticked, which points at a scheduling issue upstream of mech-interact rather than inside it.
- **Leading hunch (not a confirmed root cause):** a scheduling deadlock between Pearl's `/funds-status` poll and the `mech_information` behaviour's outgoing IPFS fetches, mediated by `funds_manager/behaviours.py:59` issuing a `Web3.HTTPProvider(rpc_url)` with **no timeout kwarg** on the AEA event loop thread. Under any RPC stall, that call would block the loop indefinitely and starve the behaviour tick loop. Evidence is circumstantial — I have not yet proven this is the trigger for the current hang, only that it is a real latent bug and the only unbounded synchronous network call reachable from any Pearl-polled GET handler in the trader.
- **Speculative fix to explore separately** (one-liner in `funds_manager`, not in mech-interact; brings it in line with the existing convention at `trader_abci/handlers.py:688`):
  ```diff
  - w3 = Web3(Web3.HTTPProvider(rpc_url))
  + w3 = Web3(Web3.HTTPProvider(rpc_url, request_kwargs={"timeout": 10}))
  ```
  **No follow-up PR opened yet** — the `funds_manager` change lives in a different repo and this is still a hunch that needs validation (e.g. instrumented Pearl repro, or a temporary `skill_exception_policy: just_log` debug run to surface any swallowed exceptions) before it gets shipped.

Reverting #908 narrows the surface area while we continue diagnosing.

## Test plan
- [ ] CI green
- [ ] `packages.json` and all dependent skill/agent/service hashes match mech-interact v0.25.0 state (verified locally via `autonomy packages sync --update-packages` — no further changes produced)


bennyjo (approved):
Approving — clean symmetric revert (23+/23-, all 8 hashes confirmed against pre-#908 state, dependent skill/agent/service hashes coherent per `autonomy packages sync`, CI green). Two non-blocking asks before merge:

1. The PR description references `MECH_INFORMATION_HANG_REPORT.md` "on the branch" — the file isn't present anywhere in the `chore/revert-bump-mech-interact` tree. Either commit the report or inline the conclusion and drop the citation.
2. The "real fix" (`funds_manager` `Web3.HTTPProvider` timeout) has to land in the upstream `funds_manager` repo — only its hash lives in this tree. Please link the follow-up PR (or open one) so main isn't left in "reverted, no replacement scheduled" limbo.

DIvyaNautiyal07 (approved):

PR #912, opened 2026-04-21
feat: disable betting on Polymarket markets by tag

LOCKhart07:
## Summary

Stops polystrat from placing new bets on Polymarket markets whose event tags match a configured disable list. Derived from `plans/classification/PNL_BY_TAG_REPORT.md` §7.1 — 36 slugs covering `hide-from-new`, narrow-band `rewards-*`, foreign elections, Trump-geopolitics, and US primaries. Per that report's backtest, disabling these would have flipped the 3-month ROI from −12.5% to +4.6%.

## Expected flow (fetch → filter → sample)

Two-knob design: a coarse **include** list (hardcoded for now) controls the fetch surface; a fine-grained **disable** list (config, this PR) is applied at sampling time as a policy filter.

1. **Fetch (include list, unchanged):** iterate the 9 hardcoded slugs in `POLYMARKET_CATEGORY_TAGS` (`business`, `politics`, `science`, `technology`, `health`, `entertainment`, `weather`, `finance`, `international`) and hit `/events?tag_slug=<slug>` for each.
2. **Flatten:** each event's full `tags[]` list is attached to every child market as `_poly_tags`. A market fetched via `politics` might carry e.g. `['politics', 'trump-iran', 'middle-east']`.
3. **Connection-side filter pipeline (data correctness only, no policy):**
   - `_filter_markets_by_created_at` — drop markets older than the subgraph cutoff.
   - `_filter_yes_no_markets` — binary-outcome only.
   - `_filter_tradeable_markets` — `active=true` + non-empty `outcomePrices`/`clobTokenIds`. Restores the server-side filter that `/markets?tag_id=X` used to do implicitly. WARN fires if this drops 100% of a non-empty input (upstream schema-shift signal).
4. **Bet construction:** `_process_chunk` creates new bets (`poly_tags` populated from `_poly_tags`) and updates existing ones via `Bet.update_market_info`, which **copies `poly_tags` over** — this is what refreshes tags on legacy pre-PR bets that deserialized with `poly_tags=[]`.
5. **Sampling-time policy filter (the disable list):** `decision_maker_abci/behaviours/sampling.py::_sample` iterates candidates and, for each, checks `set(bet.poly_tags) & disabled_polymarket_tags` with both sides lowercased. Matching bets are skipped on the same cycle. Nothing is blacklisted — the bet stays in its natural state, only sampling ignores it.

**Worked example.** A market tagged `['politics', 'trump-iran', 'middle-east']` gets fetched via `politics`, flattened with all three tags attached, passes created-at/yes-no/tradeable, becomes a Bet with `poly_tags=['politics', 'trump-iran', 'middle-east']`. When `_sample` gets to it, `{'trump-iran'} & {'trump-iran', 'middle-east', …36 slugs}` is non-empty → skipped with `INFO Sampled bet X has disabled tag(s) ['trump-iran'], skipping`. Sampling continues with the next candidate.

**Orphan legacy-bet case** (pre-PR bet with `poly_tags=[]`): market is still returned by the connection (the connection no longer filters by tag); `_process_chunk` runs `update_market_info` which copies the fresh `poly_tags` onto the stored bet; sampling then sees the refreshed tags and skips. Fixed by construction — no blacklist path, no persistent state change.

## How it works (minimal blast radius)

- **`connection.py`:** `/markets?tag_id=X` → `/events?tag_slug=X` so tag data rides with every market (REPORT §5: strictly fewer HTTP rounds, richer data). New `_filter_tradeable_markets` restores the server-side `active=true` + populated-fields gate the old endpoint did implicitly. No policy logic — `disabled_tags` is not a connection concern.
- **`bets.py`:** additive `poly_tags: List[str] = []` on `Bet`. Legacy `multi_bets.json` deserializes unchanged. `update_market_info` now copies `poly_tags` from the incoming bet so legacy `[]` tags get refreshed on every re-fetch.
- **`polymarket_fetch_market.py`:** populates `bet_dict.poly_tags` from `_poly_tags`. Otherwise unchanged.
- **`sampling.py`:** new filter block inside `_sample`'s while loop, alongside the existing neg_risk / liquidity / outcome-side-threshold checks. Lower-cased set intersection against `self.params.disabled_polymarket_tags`. `DecisionMakerParams` already inherits from `MarketManagerParams`, so no new params plumbing.
- **Config:** `disabled_polymarket_tags` on `MarketManagerParams`, env-var overridable. Empty list = feature off. Declared in `market_manager_abci/skill.yaml`, and duplicated into `decision_maker_abci/skill.yaml` and `trader_abci/skill.yaml` because Open Autonomy composition forces each composing skill to re-declare the nested skill's `args:`. Drift between the three is guarded by an invariant test.

No FSM, composition, or other sampling-layer changes. Sampling already respects `QueueStatus.EXPIRED`.

Also bundles a one-line `Makefile` fix (`aea-helpers` → `uv run aea-helpers`) so `make run-agent` works in a fresh clone.

## Validations

**Unit tests (2875 green in scope):**
- 11 tests for `_fetch_markets_by_tag_slug` (events flatten, pagination, params, tag attach, empty events/tags).
- 8 tests for `_filter_tradeable_markets` (drops on missing `outcomePrices` / `clobTokenIds` / `active=false` / malformed JSON; keeps populated+active; logs DEBUG on malformed JSON).
- 1 integration test for `_fetch_markets` + tradeable filter + 100% drop WARN.
- 2 `Bet` round-trip tests for `poly_tags` (legacy-without, new-with).
- 1 direct unit test on `Bet.update_market_info` → copies `poly_tags`.
- 2 `MarketManagerParams` tests (reads list, raises on missing key).
- 1 behaviour test that `bet_dict.poly_tags` flows from `market["_poly_tags"]`.
- 4 new `_sample` tests: banned bet skipped, case-insensitive match, empty list no-op, **orphan legacy bet with refreshed tags skipped** (regression guard for J1).
- 1 drift-risk invariant test loading all 3 `skill.yaml` files and asserting `disabled_polymarket_tags` are identical.
- Removed 21 tests for deleted `_fetch_tag_id` / `_fetch_markets_by_tag` / `_load_cache_file` / `_save_cache_file` (round 1) and 5 `_blacklist_disabled_tag_bets`-related tests (round 2, method no longer exists).

**Scope run:** `market_manager_abci`, `decision_maker_abci`, `trader_abci`, `agent_performance_summary_abci`, `polymarket_client` → 2875 pass.

**Lint:** `tomte check-code` green (black, isort, flake8, mypy, pylint, darglint).

**Hashes:** `autonomy packages lock` regenerated; 7 skills + connection + agent + 2 services updated.

**API feasibility spot-checks (live against `gamma-api.polymarket.com`):**
- `/events` vs `/markets` field parity: all 13 downstream fields (`conditionId`, `outcomes`, `outcomePrices`, `clobTokenIds`, `endDate`, `negRisk`, `liquidity`, `spread`, `createdAt`, `question`, `id`, `closed`, `submitted_by`) present, with identical values for shared `conditionId`.
- Tag coverage: 100% of sampled events had `tags[]`, 5/5 events produced slug lists on first try.
- Throughput: 5 events → 75 markets (~15 markets/event), matching REPORT §5.
- Tradeable-filter feasibility: tested `enableOrderBook=true` query param — **not viable** (all 26 problem markets already had it `true`). Tested `active=true` on `/events` — **not viable** (flag applies at event level, doesn't filter nested markets). Client-side filter confirmed as the only path and implemented.

**Live agent-log validation (`logs/agent_log_latest.log`):**
- **Sampling filter firing correctly:** 60 `Sampled bet X has disabled tag(s), skipping` lines on a single cycle. Examples: `['hide-from-new']`, `['hide-from-new', 'rewards-automation-50-4pt5-50']` (multi-slug intersection works).
- **Tradeable filter intact:** politics dropped 26 inactive/unpriced markets; other categories dropped 0 (matches their API profile).
- **0 missing-field warnings** (`Skipping market ... Missing required fields` count: 0 — tradeable filter gates them out client-side).
- **0 old connection-filter log lines** (`Dropped N markets matching disabled tags`) — confirms the connection filter is gone.
- **Sampling still reaches valid bets:** one successful sample on a Trump "AI during WHCA Dinner" market with `poly_tags=['mention-markets', 'politics', 'trump']` (none of which are in the disable list — `trump` alone is kept per REPORT §7.3).

## Known caveat

`trump-presidency` is in the default disable list per §7.1, but §7.3 flags it as shared with profitable Trump-domestic markets. Leaving as-is — operator can drop the slug from the YAML (or env-var override) if domestic Trump markets start showing up as false positives. Disable list is config, not code, and filter is read-time (no blacklist), so the change takes effect on the very next sample.

## Things intentionally left out of scope (per `REPORT.md`)

- The broader two-level config (`enabled_categories` / `category_groups` / `disabled_subcategories`). Include list (`POLYMARKET_CATEGORY_TAGS`) is still hardcoded in `connection.py` — the symmetric config move is a separate follow-up.
- Regex classifier supplement for Polymarket's 3% under-tagged ladder gap.
- `/events` coverage expansion (`sports`, `crypto`, `pop-culture`, `world`).
- Rewards-income audit for `rewards-*` categories (§8 item 1) — list may need pruning once that lands.

## Review comment status

Round 1 (`DIvyaNautiyal07`, 4 comments):
- Blocking: `update_market_info` missing `poly_tags` copy → fixed. (This fix remains load-bearing for the sampling-layer approach — without it, orphan legacy bets never get their tags refreshed.)
- Nit: dead `cache_file_path` kwarg → dropped.
- P2: silent JSON error → DEBUG log added.
- Test quality: MagicMock wiring test → replaced with real behaviour test (later migrated to sampling in round 2).

Round 2 (`jmoreira-valory` × 3, `bennyjo` × 2):
- **J1 (orphan legacy bets):** filter moved from connection to sampling layer. Orphan case fixed by construction — banned markets now stay in the fetch response, `update_market_info` refreshes legacy `poly_tags`, sampling reads them on the same cycle. Live-log shows the filter firing correctly on real bets.
- **J2 (case sensitivity):** `.lower()`-normalized on both sides of the intersection in `_sample`.
- **J3 (no log trace):** DEBUG on `_sample` entry + INFO per skip; 60 hits observed in the latest run.
- **B1 (silent 100% tradeable drop):** WARN fires on `yes_no_markets and not tradeable_markets`.
- **B2 (drift risk across 3 skill.yamls):** invariant test in `test_disabled_tags_invariant.py` asserts all three match.

## Test plan

- [ ] CI green on the new branch
- [ ] Deploy to a test agent, confirm `Sampled bet X has disabled tag(s), skipping` log lines appear and `Skipping market ... Missing required fields` stays at 0
- [ ] Over one trading cycle, confirm no sampled bet has a tag in `disabled_polymarket_tags`
- [ ] Confirm a legacy bet whose tag is in the disable list gets skipped at sample time (look for the bet ID in the "has disabled tag(s)" log)
- [ ] Verify toggling `disabled_polymarket_tags` at runtime via env var takes effect on the next sampling cycle (no restart needed beyond picking up the env change)
- [ ] After one week, check P&L by tag to verify no regression on non-disabled tags

## Diff breakdown

Of 1476 lines changed (+1011 / −465) across 24 files:

| Category | Lines | Nature |
|---|---:|---|
| **Actual new logic** | **~80** | `_fetch_markets_by_tag_slug` (~30), `_filter_tradeable_markets` (~15), `Bet.update_market_info` + `.poly_tags` (~2), `_sample` filter block (~15), B1 WARN (~5), J3 logs (~4), Makefile (1) |
| **Dead-code removal** | **~155** | Round 1: old `_fetch_tag_id`, `_fetch_markets_by_tag`, `_load_cache_file`, `_save_cache_file` in `connection.py` (~130). Round 2: `_blacklist_disabled_tag_bets` method + call + payload entry (~25) |
| **Disable-list config data** | **~111** | 36 slugs × 3 `skill.yaml` files — composition duplicates `MarketManagerParams.args`, invariant test guards against drift |
| **Hash noise** | **~70** | Auto-generated by `autonomy packages lock` across `packages.json`, `aea-config.yaml`, 6 service/skill yamls, plus fingerprint lines inside the 3 config yamls |
| **Tests** | **~1060** | Round 1: 32 new, 21 legacy deleted. Round 2: 4 new sampling tests, 1 invariant test, 5 stale `_blacklist_disabled_tag_bets` tests deleted, 4 connection-filter tests deleted |

TL;DR: ~**80 lines of actual new behaviour**, ~**155 lines of dead-code cleanup** (net improvement), ~**180 lines of config + hash bookkeeping**, ~**1060 lines of test churn** (majority additive — new sampling tests, invariant test, tradeable-filter tests).


packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -175,6 +175,7 @@ class Bet:
     strategy_vote: Optional[int] = None
     market_spread: Optional[float] = None
     neg_risk: bool = False
+    poly_tags: List[str] = dataclasses.field(default_factory=list)
```
DIvyaNautiyal07, line 178: **Blocking — companion fix required in `update_market_info`.** Adding `poly_tags` to the `Bet` dataclass is fine, but `Bet.update_market_info` (below, around line 381) does not copy `poly_tags` from the incoming bet. That kills the PR's stated legacy-blacklist behaviour:

1. Pre-PR bets deserialize from `multi_bets.json` with the dataclass default `poly_tags=[]`.
2. While the tag is NOT yet in `disabled_polymarket_tags`, the market is re-fetched, but `_process_chunk` takes the `update_market_info` branch for existing bets — `poly_tags` stays `[]`.
3. User later adds the slug to the disable list. The connection-side filter now drops that market from the response entirely, so the bet is never re-observed with its real tags.
4. `_blacklist_disabled_tag_bets` intersects `set([]) & disabled` → empty → never blacklists.

Net result: "blacklists legacy in-pool bets matching those tags" (per the PR description) is dead code for any bet that existed before this PR merged.

Fix: in `update_market_info`, add `self.poly_tags = list(bet.poly_tags)`. Plus a test that preloads `self.bets` with a bet whose `poly_tags=[]`, runs `_update_bets` with a fresh chunk carrying `_poly_tags=["x"]`, and asserts the existing bet ends up with `poly_tags == ["x"]`.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -609,40 +570,30 @@ def _remove_duplicate_markets(self, markets: list) -> list:
...
-        Resolved/over markets (extreme outcome prices) are blacklisted in the
-        PolymarketFetchMarketBehaviour._blacklist_expired_bets logic.
+        Fetches events per category from /events?tag_slug=X, flattens to markets
+        with per-market _poly_tags attached, and filters for Yes/No outcomes.
+        Markets whose _poly_tags intersect `disabled_tags` are dropped. Resolved
+        markets (extreme outcome prices) are blacklisted downstream by
+        PolymarketFetchMarketBehaviour._blacklist_expired_bets.
 
-        :param cache_file_path: Optional path to persistent cache file for tag IDs
+        :param cache_file_path: Unused; kept for backwards-compatible payload shape
```
DIvyaNautiyal07: Nit: `cache_file_path` is documented as "Unused; kept for backwards-compatible payload shape", but the behaviour at `polymarket_fetch_market.py` still builds `cache_file_path = str(self.params.store_path / "polymarket.json")` and ships it in every FETCH_MARKETS payload. Since behaviour + connection live in the same repo and are released together (both hashes bumped in this PR), cleaner to drop the kwarg on both sides plus the `test_cache_file_path_kwarg_accepted_but_ignored` test fixture rather than carrying a dead param forever.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -592,6 +529,30 @@ def _filter_yes_no_markets(self, markets: list) -> list:
...
+        :return: Markets with non-empty outcomePrices and clobTokenIds, active=True
+        """
+        tradeable = []
+        for market in markets:
+            if not market.get("active"):
+                continue
+            try:
+                outcome_prices = json.loads(market.get("outcomePrices") or "[]")
+                clob_token_ids = json.loads(market.get("clobTokenIds") or "[]")
+            except (json.JSONDecodeError, TypeError):
```
DIvyaNautiyal07: P2: silent `continue` on `json.JSONDecodeError` / `TypeError`. If Polymarket changes how `outcomePrices` / `clobTokenIds` are encoded (e.g. to a native list rather than a JSON string), this filter drops 100% of markets with no log signal. A `self.logger.debug(f"Dropped market {market.get('id')}: malformed JSON in price/token fields")` on the except path would give early warning during a schema shift. Cheap.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
```diff
@@ -1855,6 +1856,129 @@ def test_successful_update(self) -> None:
...
+        behaviour._blacklist_disabled_tag_bets = MagicMock()  # type: ignore[method-assign]
+        behaviour._blacklist_expired_bets = MagicMock()  # type: ignore[method-assign]
+        behaviour.context.params.opening_margin = 1000
+        behaviour.context.params.disabled_polymarket_tags = ["hide-from-new"]
+        type(behaviour).synced_time = PropertyMock(return_value=5000)  # type: ignore[method-assign]
+
+        gen = behaviour._update_bets()
+        _exhaust_gen(gen)  # type: ignore[arg-type]
+
+        behaviour._blacklist_disabled_tag_bets.assert_called_once()  # type: ignore[attr-defined]
```
DIvyaNautiyal07: Test quality: this is a wiring test (MagicMock + `assert_called_once`), not a behaviour test — `_blacklist_disabled_tag_bets` and `_process_chunk` never run. This is precisely why the `update_market_info` / `poly_tags` bug flagged on `bets.py` isn't caught.

A stronger variant: preload `self.bets` with a bet whose `poly_tags=[]` (simulating a pre-PR legacy bet), return a fresh market chunk whose `_poly_tags=["hide-from-new"]`, run `_update_bets` with that tag in `disabled_polymarket_tags`, and assert the existing bet ends up `QueueStatus.EXPIRED`. That test would fail today.

DIvyaNautiyal07 (changes requested):

packages/valory/skills/market_manager_abci/bets.py
LOCKhart07: Good catch — fixed in d0706ce9. `update_market_info` now copies `poly_tags` from the incoming bet. Added the direct unit test on `Bet.update_market_info` in `test_bets.py` and replaced the wiring test below with a real end-to-end behaviour test (preload legacy bet, feed chunk with banned tags, assert bet ends up EXPIRED) — both tests were verified to fail against HEAD before the fix.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: Agreed — dropped in d0706ce9. Removed `cache_file_path` from `_fetch_markets` signature, from the behaviour payload, and deleted the corresponding test.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: Done in d0706ce9. Added `logger.debug` with the market ID and the exception message on the malformed-JSON except path. DEBUG so routine runs stay quiet; surfaces when log level is bumped or grepped for `malformed JSON`.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
LOCKhart07: Valid criticism — replaced in d0706ce9 with the end-to-end scenario you suggested. Preloads `self.bets` with a legacy bet (`poly_tags=[]`), returns a fresh chunk with `_poly_tags=["hide-from-new", "politics"]`, sets `disabled_polymarket_tags=["hide-from-new"]`, runs `_update_bets`, and asserts both `legacy_bet.poly_tags == ["hide-from-new", "politics"]` AND `legacy_bet.queue_status == QueueStatus.EXPIRED`. That test did fail against HEAD before the `update_market_info` fix (AssertionError on the poly_tags assertion), confirming it catches the regression.

DIvyaNautiyal07 (dismissed):
Verified `d0706ce`. All four points addressed:

1. **Blocking / `update_market_info` poly_tags copy** — fixed at `bets.py:382` (`self.poly_tags = list(bet.poly_tags)`, defensive copy). New end-to-end test `test_update_bets_blacklists_legacy_bet_when_its_tag_becomes_disabled` preloads a legacy bet with `poly_tags=[]`, feeds a chunk with banned tags, and asserts both that the tags propagated AND that `queue_status == EXPIRED` — mutation-test solid on both the copy and the blacklist call. The direct unit test in `test_bets.py` also guards the copy itself.
2. **Dead `cache_file_path` kwarg** — fully removed from the connection signature, docstring, behaviour payload, and test fixture. `grep cache_file_path packages/valory` comes back empty.
3. **Silent JSON swallow in `_filter_tradeable_markets`** — DEBUG log added with the market ID and exception message.
4. **Wiring-only test** — replaced with the end-to-end version in point #1.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -592,6 +529,34 @@ def _filter_yes_no_markets(self, markets: list) -> list:
...
+                clob_token_ids = json.loads(market.get("clobTokenIds") or "[]")
+            except (json.JSONDecodeError, TypeError) as e:
+                self.logger.debug(
+                    f"Dropped market {market.get('id')}: "
+                    f"malformed JSON in price/token fields ({e})"
+                )
+                continue
+            if outcome_prices and clob_token_ids:
+                tradeable.append(market)
+        return tradeable
```
bennyjo, line 558: Minor brittleness: if this filter ever drops 100% of a non-empty input, the only signal is the existing per-category INFO line "Filtered to 0 tradeable markets" — no explicit "something is wrong" signal.

Failure modes worth considering:
- `outcomePrices` / `clobTokenIds` changing from JSON-encoded string to a native list (`json.loads` raises TypeError → every market dropped)
- a new lifecycle state like `active="pending"` or `active=null` — `not market.get("active")` treats both as False
- transient upstream state returning empty arrays on every market

Would you consider a WARN when this drops 100% of a non-empty input? One post-loop branch, converts a silent collapse into a log signal. Skip if you're actively monitoring the existing INFO lines.

packages/valory/skills/market_manager_abci/skill.yaml
```diff
@@ -166,6 +166,43 @@ models:
       opening_margin: 300
       languages:
       - en_US
+      disabled_polymarket_tags:
```
bennyjo, line 170: Flagging a drift risk. The same 36-slug list is declared here and at `decision_maker_abci/skill.yaml:266` and `trader_abci/skill.yaml:194` — three copies, because Open Autonomy composition requires each composing skill to re-declare the nested skill's `args:`.

There's no CI check or runtime assertion that the three lists agree. An update to one without the others silently ships with the old list in whichever yaml the deployed service loads through — no exception, tests pass, `autonomy packages lock` regenerates hashes cleanly for each file independently. The only visible symptom would be "the new slug didn't take effect in production."

The PR body acknowledges the duplication ("composition duplicates `MarketManagerParams.args`") but doesn't close the drift risk. Worth explicit discussion before merge — the composition layout makes this class of bug easy to trip over.

bennyjo (commented):
Two non-blocking observations on brittleness and config drift.

jmoreira-valory (commented):
Three follow-up items — one real correctness gap around legacy bets whose markets are dropped upstream, two smaller (case sensitivity, logging). Not blocking; DIvya's approval covers the rest of the surface. Tests pass, CI green, framework invariants untouched.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -154,6 +154,26 @@ def _blacklist_expired_bets(self) -> None:
...
+        if not disabled:
+            return
+        for bet in self.bets:
+            if bet.queue_status.is_expired():
+                continue
+            hit = set(bet.poly_tags) & disabled
+            if hit:
+                self.context.logger.info(
+                    f"Blacklisting bet {bet.id!r} — disabled tags: {sorted(hit)}"
+                )
```
jmoreira-valory: The legacy-bet leak isn't fully closed. `_blacklist_disabled_tag_bets` relies on `bet.poly_tags` being populated. For legacy bets from `multi_bets.json` (pre-PR, so `poly_tags=[]`), the re-tagging only happens when `_process_chunk` runs `update_market_info` — which only runs if the market is still in the fetch response. The connection-side `disabled_tags` filter drops exactly those markets, so the behaviour never sees them, `poly_tags` stays `[]`, `set(bet.poly_tags) & disabled` is empty, and the bet stays active forever.

The new `test_end_to_end_legacy_bet_gets_refreshed_and_blacklisted` covers the happy path (legacy bet still in the chunk); the orphan case — legacy bet whose market is no longer returned because upstream now drops it — isn't tested.

Fix options:
- (a) Issue a pre-filter fetch (`disabled_tags=[]`) once per cycle specifically for the refresh-tags pass, so legacy bets get their tags populated before the blacklist loop runs.
- (b) Have the connection return dropped markets tagged but flagged (e.g. `_dropped_reason="disabled_tag"`), and let the behaviour do the final drop — so `update_market_info` still runs and populates `poly_tags`.
- (c) Match legacy bets by `market.id` against the list of IDs dropped upstream.

Non-hard-blocker because the feature is opt-in, but the PR body's "step 5 catches it on the next cycle" guarantee doesn't hold for legacy bets in this orphan case.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -405,35 +405,6 @@ def _place_bet(
             response = {"error": error_msg, "signed_order_json": signed_order_json}
```
jmoreira-valory, line 405: Case-sensitive tag intersection. YAML config ships lowercase slugs (`hide-from-new`, `trump-iran`, …). Polymarket's `/events` API may return tag slugs in mixed case depending on the event (this isn't documented as guaranteed lowercase). A single casing mismatch → zero matches, silently, with no log.

Suggestion: normalize at the intersection point on both sides — `{t.lower() for t in m.get("_poly_tags", [])} & {t.lower() for t in disabled_set}`. Same change in `polymarket_fetch_market.py::_blacklist_disabled_tag_bets` (`set(bet.poly_tags) & disabled`).

Cheap defence against a silent-failure mode that's easy to introduce the day Polymarket adds e.g. `"Trade-War"` to an event payload.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -154,6 +154,26 @@ def _blacklist_expired_bets(self) -> None:
                 ):
                     bet.blacklist_forever()
 
+    def _blacklist_disabled_tag_bets(self) -> None:
```
jmoreira-valory: No trace when the filter runs but matches nothing. Current logging only fires inside the `if hit:` branch. An operator asking "did my newly-added disabled slug take effect?" has no way to tell from the logs if zero bets matched — indistinguishable from "loop didn't run" or "tag typo". Combined with @bennyjo's related note on `_filter_tradeable_markets` logging, consider a single DEBUG at loop entry:

```python
self.context.logger.debug(
    f"Checking {len(self.bets)} bets against {len(disabled)} disabled tags"
)
```

Or INFO once per cycle — doesn't have to be verbose, just enough signal to grep for `disabled tags` and see evidence the loop ran.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
LOCKhart07: Fixed in 87435d4e by moving the filter to the sampling layer entirely.

Connection no longer filters on `disabled_tags` — it returns every market with its `_poly_tags` attached. `update_market_info` now runs on banned markets too, which refreshes `poly_tags` on legacy bets whose markets are still in the fetch window. `sampling.py::_sample` then does the set intersection on the refreshed tags and skips matching bets.

Orphan case concretely: legacy bet with `poly_tags=[]` + banned market still in fetch → market now stays in the chunk → `_process_chunk` runs `update_market_info` → bet's `poly_tags` refreshed → sampling sees the real tags on the next iteration and skips.

New `test_sample_orphan_legacy_bet_with_refreshed_tags_skipped` in `test_sampling.py` is the regression guard. Paired with `test_update_market_info_copies_poly_tags` in `test_bets.py`, the two together cover the full path.

Side benefits of this refactor:
- Policy lives in the policy skill (decision_maker_abci), not split between connection and market_manager.
- Reversible: operator removing a slug from the disable list immediately re-allows those markets next sample (no un-blacklist needed — nothing was ever blacklisted).
- Safer for redeem/position-tracking: banned bets stay in their natural state, only sampling skips them. Other consumers of `self.bets` (investments, eventually redeem) see them normally.

Live-log validation: agent_log shows 60 `Sampled bet X has disabled tag(s), skipping` lines on the new code path, 0 occurrences of the old connection-side "Dropped N markets matching disabled tags" log.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: Fixed in 87435d4e. Both sides of the intersection now `.lower()`-normalize at the one remaining site (`sampling.py::_sample`):

```python
disabled_tags = {t.lower() for t in self.params.disabled_polymarket_tags}
...
bet_tags = {t.lower() for t in sampled_bet.poly_tags}
hit = disabled_tags & bet_tags
```

Regression test: `test_sample_skips_bet_case_insensitive` — bet with `poly_tags=["Hide-From-New", "Politics"]` against disabled `["hide-from-new"]` → skipped.

Note: since J1 moved the filter to sampling, there's only one intersection site now. No duplicated normalization logic.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
LOCKhart07: Fixed in 87435d4e. Two log signals now:

1. **Entry-point DEBUG** (once per `_sample` call when disable list is non-empty):
   ```
   Sampling: filtering candidates against 36 disabled tags
   ```
2. **Per-skip INFO** (what was already there, now always fires since the filter lives in `_sample`):
   ```
   Sampled bet 2043032 has disabled tag(s) ['hide-from-new'], skipping
   ```

So an operator can grep for `disabled tag` to see evidence the loop ran (DEBUG if they want the loop-entry signal; INFO for actual hits). Together with no connection-side filter anymore, there's a single place to look.

Live-log check on the latest run: 60 per-skip INFO lines on the new code path.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: Added in 87435d4e. Per-category WARN fires when `yes_no_markets` is non-empty but `tradeable_markets` is empty:

```python
if yes_no_markets and not tradeable_markets:
    self.logger.warning(
        f"Tradeable filter dropped 100% of {len(yes_no_markets)} markets "
        f"for category '{category}' — possible upstream schema change"
    )
```

Asymmetric check is intentional — `yes_no_markets` being empty to begin with is normal upstream variation (e.g. `weather` returning zero markets some periods). We only want to signal the "had data, dropped everything" case.

Catches the three failure modes you listed:
1. `outcomePrices` / `clobTokenIds` encoding change → `json.loads` raises `TypeError` on every market → 100% drop.
2. New lifecycle state like `active=null` → `not market.get("active")` is True for all → 100% drop.
3. Transient upstream returning empty arrays on every market.

New `test_fetch_markets_warns_on_full_tradeable_drop` in `test_connection.py` asserts the WARN fires with "dropped 100%" in the message.

packages/valory/skills/market_manager_abci/skill.yaml
LOCKhart07: Closed in 87435d4e with an invariant test: `packages/valory/skills/market_manager_abci/tests/test_disabled_tags_invariant.py`.

It loads the `disabled_polymarket_tags` list from the three composed skills' `skill.yaml` files and asserts they're identical. Passes today (lists currently agree), fails the moment anyone edits one yaml without the others — catches the silent drift before it ships.

```python
def test_disabled_polymarket_tags_consistent_across_skills() -> None:
    mm = _load_disabled_tags("valory/skills/market_manager_abci")
    dm = _load_disabled_tags("valory/skills/decision_maker_abci")
    trader = _load_disabled_tags("valory/skills/trader_abci")
    assert mm == dm == trader, (
        "disabled_polymarket_tags drift detected:\n"
        f"  market_manager_abci ({len(mm)} slugs): {mm}\n"
        f"  decision_maker_abci ({len(dm)} slugs): {dm}\n"
        f"  trader_abci ({len(trader)} slugs): {trader}"
    )
```

The structural issue is pre-existing to Open Autonomy composition — we can't eliminate the duplication, only catch divergence. This test runs in the standard scope sweep so CI will fail immediately on a mismatch.

jmoreira-valory (dismissed):
Approving after 87435d4e. All three of my comments addressed — filter moved to sampling layer (closes the legacy-bet orphan correctness gap with a direct regression test), case-insensitive intersection with a case-variant test, and the entry-point DEBUG + per-skip INFO logging now surface loop activity. @bennyjo's two concerns also addressed (zero-tradeable WARN, cross-skill drift invariant test). LGTM.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -254,6 +254,16 @@ def _sample(self) -> Optional[int]:
             if not idx:
```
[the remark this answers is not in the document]
LOCKhart07: Benchmarking mode is not a concern rn. Needs to be reworked for Polystrat anyways

bennyjo (dismissed):
Re-reviewed after the round-2 restructure (filter moved from connection to `_sample`). LGTM.

All four prior findings resolved:
- J1 (orphan legacy bets) — closed by construction: market stays in fetch chunk, `update_market_info` refreshes `poly_tags`, `_sample` reads them on the same cycle. No blacklist path, no persistent state change.
- J2 (case sensitivity) — `.lower()` on both sides of the intersection.
- J3 (log trace) — DEBUG on filter setup, INFO per-skip.
- B1 (silent 100% tradeable drop) — WARN fires on non-empty input collapsing to empty.
- B2 (yaml drift across 3 skills) — `test_disabled_tags_invariant.py` guards parity with strict `==`.

Architecturally cleaner — skill boundaries are now right (market_manager = data, decision_maker = policy). Filter is a pure read-time pattern match, so toggling the YAML takes effect on the next sample with no migration step. Omen bets unaffected (poly_tags stays `[]`, empty intersection). Test coverage is discriminating (each new `_sample` test fails against pre-fix code).

One minor observation, not blocking: the "No valid bets found after liquidity validation!" WARN at the end of `_sample` also fires when tag/neg_risk/outcome-threshold exhaust candidates — message is misleading but pre-existing.

mariapiamo (commented):
[P2] Disabled-tag policy can be bypassed, and decisions are not reproducible, because tags are mutable
- Assumptions: upstream /events metadata can be temporarily incomplete; policy is expected to be stable/auditable across runs.
- Evidence: tags are taken directly from event payload and copied into bet state, then used for skip decisions: connection.py (line 477), connection.py (line 481), bets.py (line 382), sampling.py (line 304).
Fix:
- Do not clear existing poly_tags on empty incoming tags (or require N>=2 consecutive empty reads before clearing).
- Add append-only tag_history (ts, old_tags, new_tags, source).
- Add immutable decision snapshot fields (tags_at_decision, disabled_tags_at_decision, policy_version, decision_ts) persisted in multi_bets.json.

[P3] Disabled bets are skipped but re-evaluated every cycle
Assumptions: non-trivial number of disabled-tag bets is expected in production.
Evidence: skip happens in sampling loop with no state transition/caching: sampling.py (line 304).
Fix: introduce a persistent blocked state (e.g., BLOCKED_BY_TAG) or cached skip marker, and re-open only when tag/config changes.

[P3] Tag normalization/validation is weak at policy boundary
Tags are consumed directly from upstream and matched via lowercase only (no trim/canonicalization/provenance confidence), in connection.py and sampling logic.
This can create edge-case bypasses/noise with formatting drift ("stock " vs "stock", mixed payload quality).
Low severity, but easy hardening win.

[P3] Operational config gap for disabled_polymarket_tags
Assumptions: operators need runtime tuning without package edits.
Evidence: list is declared in skill configs but not exposed in service/agent runtime templates: service.yaml, service.yaml, aea-config.yaml.
Fix: expose disabled_polymarket_tags via env-driven override in service/agent templates (list/json parsing + validation).

[P4] Sampling optimization (non-blocking)
If policy is strict (disabled tag => always skip), run tag filtering before other per-bet checks and preferably pre-filter available_bets (or cache/block disabled bets) so they are not reconsidered every cycle.
This reduces local CPU/log churn only; it does not reduce external network/API calls in _sample.

LOCKhart07:
Thanks for the detailed review @mariapiamo. Summary of what's addressed in 856ec7c7 and what's explicitly out of scope.

**Addressed (landed in 856ec7c7):**

- **P3 tag normalization** — `.strip().lower()` on both sides of the intersection in `_sample`. Regression test `test_sample_strips_whitespace_around_tags` with padded + mixed-case input.
- **P3 config gap** — `disabled_polymarket_tags` now exposed via `${DISABLED_POLYMARKET_TAGS:list:[...]}` env override in `polymarket_trader/service.yaml`, `trader_pearl/service.yaml`, and `trader/aea-config.yaml`. Operators can tune at deploy time without editing packages.
- **P4 sampling optimization** — pre-filter `available_bets` before the while-loop instead of per-iteration. Regression-guarded by a test that wraps `_sampled_bet_idx` and asserts it's called once when 5 banned + 1 ok bets are passed in (previously 6 times).

**Explicitly not addressing:**

- **"Preserve `poly_tags` on empty incoming tags"** — pushed back. Polymarket is the source of truth for tags; adding heuristics for "maybe the API is transiently lying" introduces ambiguity about when the filter applies and breaks the read-time discipline that the sampling-layer pattern is built on. If tags are unreliable upstream, that's a data-correctness problem to address at the source, not to paper over with N-consecutive-read logic here.

- **Append-only `tag_history` + immutable decision-snapshot fields** — deferred pending production evidence that Polymarket actually returns empty tags for active markets. The audit/recovery machinery only becomes useful if the wipe scenario is real; absent data showing that, building it is premature. Also follows directly from the above pushback — if the API is the source of truth, we don't need history to guard against it being wrong. Happy to revisit if log data ever shows otherwise.

- **P3 "blocked state" (BLOCKED_BY_TAG queue status with policy versioning)** — deferred. Meaningful optimization but adds persistent state + policy-version tracking + re-evaluation paths on config change. Not justified by the current load profile; the P4 pre-filter already gets us the main CPU win without the persistence complexity.


- The above points are claude fluff, but my  reasoning is "If Polymarket provided tags are unreliable, this whole feature stops making sense"

mariapiamo (approved):

DIvyaNautiyal07 (approved):

bennyjo (approved):
Re-reviewed after `856ec7c7` (round 3). LGTM — merge.

All three of mariapiamo's non-deferred findings addressed cleanly:
- **P3 tag normalization** — `.strip().lower()` on both sides of the intersection (`sampling.py:266,272`), regression-guarded by `test_sample_strips_whitespace_around_tags` with `"  Hide-From-New  "` on the bet side.
- **P3 config gap** — `disabled_polymarket_tags` now env-overridable via `${DISABLED_POLYMARKET_TAGS:list:[...]}` in `polymarket_trader/service.yaml`, `trader_pearl/service.yaml`, and `trader/aea-config.yaml`.
- **P4 pre-filter** — banned bets dropped from `available_bets` once before the while-loop; discriminating test asserts `_sampled_bet_idx` is called 1× (not 6×) for 5-banned + 1-ok.

Pre-filter semantics verified: `_sampled_bet_idx` returns indices into `self.bets`, not `available_bets`, so re-assigning the filtered list doesn't misindex (test expects `result == 5` at `self.bets[5]`). Empty-after-filter branch correctly returns `None` with a WARN.

Two minor non-blocking observations for a follow-up, not holding this PR:
- `test_disabled_tags_invariant.py` covers the 3 `skill.yaml` files but not the 3 new deployment-config occurrences added in round 3 (`polymarket_trader/service.yaml`, `trader_pearl/service.yaml`, `agents/trader/aea-config.yaml`). Worth extending since service.yaml overrides skill.yaml at deploy time.
- Per-bet `INFO "Sampled bet X has disabled tag(s)..."` was replaced by an aggregate DEBUG. Fine in steady state; slightly reduces per-bet observability when operators need to trace a specific market.

CI green, `mergeable: clean`, no security or attacker-analysis concerns on this diff.

jmoreira-valory (approved):

OjusWiZard (approved):

PR #914, opened 2026-04-23
chore(ci): add aggregate `All checks passed` gate

DavidMinarsch:
## Summary
- Adds a single `all_checks_passed` meta job in `common_checks.yaml` whose `needs` enumerate every mandatory job (`lock_check`, `copyright_and_dependencies_check`, `linter_checks`, `scan`, `test`, `coverage`).
- The job uses `if: always()` and fails if any dependency `failure` or `cancelled`, so it accurately reflects the aggregate status.
- Lets branch protection require a single stable context (`All checks passed`) instead of the full matrix. Adding/removing mandatory jobs is then a code change — update the `needs` list — rather than a branch-protection edit.

Once this lands, branch protection on `main` will be updated to require the `All checks passed` context with `strict: true`.

## Test plan
- [ ] CI runs all jobs as before
- [ ] `All checks passed` reports success when all upstream jobs pass
- [ ] `All checks passed` reports failure if any upstream job fails or is cancelled

### DIFF
--- a/.github/workflows/common_checks.yaml
+++ b/.github/workflows/common_checks.yaml
@@ -252,3 +252,19 @@ jobs:
           files: ./coverage.xml
           flags: unittests
           fail_ci_if_error: false
+
+  all_checks_passed:
+    name: All checks passed
+    if: always()
+    needs:
+      - lock_check
+      - copyright_and_dependencies_check
+      - linter_checks
+      - scan
+      - test
+      - coverage
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Fail if any required job failed or was cancelled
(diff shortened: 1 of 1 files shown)


dev-vektr-ai:
Review summary: this change looks correct to me.

The new `all_checks_passed` meta job cleanly aggregates the required CI jobs, and using `if: always()` together with the explicit failure condition on `failure`/`cancelled` preserves the gate semantics while giving branch protection a single stable status context (`All checks passed`).

I attempted to submit this as a formal approval review, but GitHub returned an organization-level OAuth app restriction (403), so I’m leaving the review outcome here as a PR comment instead.

PR #915, opened 2026-04-24
fix(ci): append .exe suffix to agent_runner_bin path on Windows

LOCKhart07:
## Summary

- Windows release flow currently fails at `make check-agent-runner` because `./dist/agent_runner_bin` doesn't exist — PyInstaller emits `./dist/agent_runner_bin.exe` on Windows.
- Introduce a platform-aware `EXE_SUFFIX` Make variable (`.exe` when `$(OS)=Windows_NT`, empty otherwise) and apply it to all three `./dist/agent_runner_bin` call sites in the Makefile.
- On Linux/macOS behaviour is unchanged; on Windows the paths now correctly target the `.exe` PyInstaller produced.

## Background

Release Flow failure on windows-2025 runners:

```
Error: Invalid value for 'BINARY_PATH':
Path './dist/agent_runner_bin' does not exist.
make: *** [Makefile:198: check-agent-runner] Error 2
```

`signCommand command for file dist/agent_runner_bin.exe was SUCCESSFUL` earlier in the same job confirms the binary exists with the `.exe` extension — the check step just looks at the wrong path.

Git Bash's PATHEXT lookup had been masking the issue for the `./dist/agent_runner_bin --version` sanity check at the end of the build targets (`Makefile:148, 167`), but `aea-helpers check-binary` is a Python tool using `Path.exists()`, which doesn't auto-resolve extensions — so the check-binary step always fails on Windows.

Example of the failure in Release Flow: https://github.com/valory-xyz/trader/actions/runs/24893174943/job/72890799087

## Test plan

- [x] Verified `make -n check-agent-runner` on Linux still renders `uv run aea-helpers check-binary ./dist/agent_runner_bin ./agent` (unchanged behaviour).
- [x] Release Flow `build-agent-runner (windows-2025)` job passes.
- [x] Release Flow `build-agent-runner (ubuntu-latest)` and `build-agent-runner-mac` jobs remain green.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -201,8 +201,12 @@ check-agent-runner:
 	# for the skill's store_path, so a single STORE_PATH override drives it.
 	# Path-based env vars like SKILL_..._STORE_PATH are the fallback when the
 	# template lacks an explicit var name and are silently ignored here.
+	# Use a repo-local scratch dir under ./dist/ rather than ``/tmp`` — the
+	# latter doesn't exist on Windows and trips staking_abci's store-path
+	# ``isdir`` check during the binary's boot sequence.
+	mkdir -p ./dist/.runner-check-store
 	uv run aea-helpers check-binary ./dist/agent_runner_bin$(EXE_SUFFIX) ./agent \
-	--env-var STORE_PATH=/tmp
+	--env-var STORE_PATH=./dist/.runner-check-store
 
 .PHONY: ci-linter-checks
 ci-linter-checks:
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,13 @@ MECH_INTERACT_VERSION := $(shell git ls-remote --tags --sort="v:refname" https:/
 # keeps the path references consistent across platforms.
 EXE_SUFFIX := $(if $(filter Windows_NT,$(OS)),.exe,)
 
+# Dummy STORE_PATH for the ``check-agent-runner`` binary smoke test.
+# ``aea-helpers check-binary`` chdirs into ``./agent`` before spawning the
+# binary, so a relative path wouldn't resolve from the subprocess cwd. Use
+# ``/tmp`` on Linux/macOS and ``%TEMP%`` (inherited as ``$(TEMP)``) on
+# Windows — both are absolute, writable, and guaranteed to exist.
+STORE_PATH_VALUE := $(if $(filter Windows_NT,$(OS)),$(TEMP),/tmp)
+
 .PHONY: sync-packages
 sync-packages:
 	@echo "Syncing packages with versions:"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -195,8 +208,8 @@ check-agent-runner:
 	# for the skill's store_path, so a single STORE_PATH override drives it.
 	# Path-based env vars like SKILL_..._STORE_PATH are the fallback when the
 	# template lacks an explicit var name and are silently ignored here.
-	uv run aea-helpers check-binary ./dist/agent_runner_bin ./agent \
-	--env-var STORE_PATH=/tmp
+	uv run aea-helpers check-binary ./dist/agent_runner_bin$(EXE_SUFFIX) ./agent \
+	--env-var STORE_PATH=$(STORE_PATH_VALUE)
```
DIvyaNautiyal07, line 212: On Windows the release workflow runs this recipe under `defaults: shell: bash` (Git Bash). `$(TEMP)` on a windows-2025 runner is a backslash path (e.g. `D:\a\_temp` or `C:\Users\RUNNER~1\AppData\Local\Temp`). After Make substitutes the value, bash receives the token `STORE_PATH=D:\a\_temp` **unquoted** and strips the backslashes:

```
$ bash -c 'printf "%s\n" --env-var STORE_PATH=D:\a\_temp'
--env-var
STORE_PATH=D:a_temp
```

So `aea-helpers check-binary` will hit the same `ValueError: The store path '...' is not a directory` failure that the previous `/tmp` and `./dist/.runner-check-store` attempts produced.

Either quote the assignment so the backslashes survive shell tokenization:

```make
--env-var "STORE_PATH=$(STORE_PATH_VALUE)"
```

or normalize to forward slashes at the variable level:

```make
STORE_PATH_VALUE := $(if $(filter Windows_NT,$(OS)),$(subst \,/,$(TEMP)),/tmp)
```

Worth re-running the windows-2025 release-flow job (or a workflow_dispatch variant) to confirm before relying on this — the test plan's Windows checkbox is still unchecked and the release workflow only triggers on tag push, so PR CI doesn't exercise this path.

LOCKhart07: I've ran the Release Flow on this commit (`cc2b78a8`, the same diff tagged as v0.35.0-rc1) and the windows-2025 `build-agent-runner` job passed end-to-end, including the `make check-agent-runner` step that was failing before:

- Release Flow run (all 4 `build-agent-runner` matrix legs green): https://github.com/valory-xyz/trader/actions/runs/24895867946
- windows-2025 job specifically: https://github.com/valory-xyz/trader/actions/runs/24895867946/job/72900609789
- Resulting release: https://github.com/valory-xyz/trader/releases/tag/v0.35.0-rc1

So in practice `STORE_PATH=$(TEMP)` survives Git Bash tokenization on the windows-2025 runner — `%TEMP%` there is `C:\Users\RUNNER~1\AppData\Local\Temp` and `aea-helpers check-binary` accepts it without the backslash-stripping you reproduced locally. Happy to add explicit quoting / forward-slash normalization as a belt-and-braces follow-up if you'd still prefer it, but the windows test plan box is empirically green now.

DIvyaNautiyal07 (approved):

Makefile
DIvyaNautiyal07: Thanks for the rc tag run, that settles it. The check step is green on windows-2025 in that release run, so empirically the path survives whatever Git Bash + MSYS2 make does with the recipe on the GitHub-hosted runner. My local repro must've been missing something about how MSYS2 make hands the recipe off to sh, or how TEMP comes through in that environment. Either way, the actual CI is the source of truth.

No need to add the quoting just for hardening, current code does the job. If a future runner image ships with a TEMP shape that does get mangled, the fix is a one-liner. LGTM as-is.

DhairyaPatel7 (approved):

PR #918, opened 2026-04-27
feat(perf-summary): migrate to predict-* subgraph PR #103

LOCKhart07:
## Summary

Minimal trader-side migration to the new predict-omen / predict-polymarket schemas from [autonolas-subgraph-studio PR #103](https://github.com/valory-xyz/autonolas-subgraph-studio/pull/103). Stacked on top of \`polystrat-clob-v2\`.

The PR keeps every displayed metric semantically identical:

- **TraderAgent**: \`totalPayout\` (which now means *actually redeemed*) is replaced by \`totalExpectedPayout\` (which preserves the old *expected at settlement* semantics) at the 3 query selections + 4 \`behaviours.py\` reads where the trader uses it for ROI / lifetime profit / locked funds calculations.
- **Polymarket \`Bet\`**: \`amount\` and \`shares\` are now signed (negative on sells). Two prediction-history queries get \`where: { isBuy: true }\` filters so the existing arithmetic stays positive. The trader has no Polymarket sell path today (\`polymarket_client/connection.py\` hardcodes \`side=BUY\`, no SELL request type, redeem is post-resolution only), so this is a no-op against current data and a guard against future inputs.
- **MarketParticipant** and **DailyProfitStatistic** \`totalPayout\` selections were intentionally left unchanged: their semantics are unchanged across PR #103, and the existing redemption-status check (\`if total_payout == 0\` → \"won but not redeemed → pending\" at \`predictions_helper.py:1043\` and \`polymarket_predictions_helper.py:294\`) relies on the *redeemed* semantics that those fields have always had.
- **Endpoints**: 5 YAMLs swapped from staging / legacy hosts (\`api.subgraph.staging.autonolas.tech\`, \`predict-agents.subgraph.autonolas.tech\`, \`predict-polymarket-agents.subgraph.autonolas.tech\`) to production: \`https://api.subgraph.autonolas.tech/api/proxy/predict-omen\` and \`.../predict-polymarket\`.

## Out of scope

- \`Bet.timestamp\` → \`Bet.blockTimestamp\` rename (the deprecated field still works, no code change required to keep working).
- Cache invalidation for \`agent_performance.json\` (the daily-profit attribution-day shift produces a one-day boundary artifact that we accepted; cumulative profit is unchanged, the bulk of historical days stay where they are).
- New entities (\`PayoutRedemption\`, \`MarketParticipant.expectedPayout\`, \`outcomeShares*\`/\`outcomeTokenBalance*\`, \`settled\`, \`Bet.isBuy\` field selection beyond the where-filter) — additive, ignored for this minimum-change cut.
- No subgraph-side changes (those are upstream in \`autonolas-subgraph-studio\`).

## Test plan

- [x] \`uv run pytest packages/valory/skills/agent_performance_summary_abci/tests/\` — 967 passed
- [x] \`tomte check-code\` — black, isort, flake8, mypy, pylint, darglint all green
- [x] \`autonomy packages lock\` — fingerprints regenerated and committed
- [ ] Manual: spot-curl the 9 affected queries against the new endpoints with a known agent address; confirm \`totalExpectedPayout\` is present and positive for an agent with at least one settled win


OjusWiZard (approved):

bennyjo (approved):
LGTM. Migration is mechanically clean and the new fields verify against the live schema (predict-omen `TraderAgent.totalExpectedPayout` and predict-polymarket `Bet.isBuy` both confirmed via introspection).

Worth a release-note line on the two user-visible behavior shifts so users don't read them as bugs:
- ROI and all-time profit step **up** to include settled-but-unredeemed winnings (was previously redeemed-only). Direction is always up or flat. Magnitude depends on how much the agent leaves unclaimed; agents that redeem promptly will see near-zero change.
- Daily profit history re-attributes from redemption date to settlement date (per upstream PR #103's accounting move). Cumulative profit unchanged.

Minor follow-up candidates (non-blocking): `GET_POLYMARKET_TRADER_AGENT_BETS_QUERY` is the only Polymarket bet query without the `where: { isBuy: true }` filter — fine today since it doesn't read amount/shares and there's no sell path; flagging only as a future-proofing asymmetry. Also, `tests/graph_tooling/test_queries.py` checks operation names and non-empty strings only, so a future revert of the rename or filter wouldn't fail any test (the ROI behaviour tests do indirectly cover the rename, though).

jmoreira-valory (approved):

Tanya-atatakai:
**Review notes (non-blocking, follow-ups):**

The migration is correct and minimal — `totalPayout` → `totalExpectedPayout` rewires preserve the "estimated at settlement" semantics, and `totalTradedSettled` / `totalFeesSettled` now flow through automatically because the subgraph itself moved them off claim-time. A few items worth a follow-up:

1. **Extend `isBuy: true` to the remaining Polymarket bet selections.** The commit frames the filter as defensive against the new signed `amount`/`shares` schema. Two sibling queries omit it:
   - `GetPolymarketTraderAgentBets` (`queries.py:347–366`) feeds `_calculate_polymarket_accuracy` (`behaviours.py:606`). Even though it only reads `id`/`outcomeIndex`/`winningIndex`, a user-side sell would create a sibling `Bet` row in the same resolved market and inflate the win-rate denominator. The agent has no sell path, but a user can sell on Polymarket directly.
   - `GetPolymarketDailyProfitStatistics` (`queries.py:423–459`) — `profitParticipants[].bets` only reads `blockTimestamp` / `metadata.title`. Lower impact, but inconsistent.

2. **`DailyProfitStatistic.totalPayout` is selected but never consumed.** Lines 211 (Omen) and 439 (Polymarket) still pull `totalPayout`; the chart pipeline only reads `dailyProfit` (`behaviours.py:1322`, `1567`). Either drop the selections or comment why they stay.

3. **`MarketParticipant.totalPayout` correctly stays on claim-time semantics** — the WON/PENDING and invalid-market proration gates (`predictions_helper.py:1041`, `polymarket_predictions_helper.py:249, 294, 302`) depend on it.

4. **Endpoint cutover is consistent** across `aea-config.yaml`, both `service.yaml` files, and the three skill yamls.

5. **Optional**: the new schema also exposes `MarketParticipant.expectedPayout` and `outcomeShares0/1` (Polymarket), `outcomeTokenBalance0/1` and `Bet.outcomeTokenAmount` (Omen). Not used today; available if you ever want per-market estimated-at-settlement values in the UI.

DIvyaNautiyal07 (approved):
## Verdict: Approve

Small, well-scoped migration to the new predict-omen / predict-polymarket schemas. Every concept that changed in upstream PR #103 is traced through cleanly, and the rationale for what was *not* changed is documented and correct.

### Reasoning

**TraderAgent.totalPayout -> totalExpectedPayout migration is complete.** 3 query selections (`GET_TRADER_AGENT_QUERY`, `GET_TRADER_AGENT_PERFORMANCE_QUERY`, `GET_POLYMARKET_TRADER_AGENT_PERFORMANCE_QUERY`) and 4 reads in `behaviours.py` (lines 370, 409, 415, 725) all switched. Test fixtures and the renamed `test_trader_agent_missing_totalExpectedPayout` are consistent with the new field name.

**`MarketParticipant.totalPayout` and `DailyProfitStatistic.totalPayout` retentions are correct.** The redemption-status check at `predictions_helper.py:1043` and `polymarket_predictions_helper.py:294` (`if total_payout == 0` -> 'won but not redeemed -> pending') relies on the *redeemed* semantics those fields carry, which the upstream PR did not change for these entities.

**Polymarket Bet `isBuy: true` filter is applied where it matters.** Only the two queries that read the now-signed `amount`/`shares` (`GET_POLYMARKET_PREDICTION_HISTORY_QUERY`, `GET_POLYMARKET_SPECIFIC_BET_QUERY`) got the filter; queries that only read `outcomeIndex` were left alone, which matches the stated rationale. Verified the 'no Polymarket sell path today' claim: `polymarket_client/connection.py:497` hardcodes `side=BUY`, so this is a no-op against current data.

**Endpoint cutover is consistent.** All 5 YAMLs swap from staging / legacy hosts (`api.subgraph.staging.autonolas.tech`, `predict-agents.subgraph.autonolas.tech`, `predict-polymarket-agents.subgraph.autonolas.tech`) to the production proxy endpoints. Grepped for any leftover references; none remain.

**Package hashes are regenerated end-to-end** (skill.yaml fingerprints, agent aea-config, service yamls, packages.json), and the parent skills (`trader_abci`, `decision_maker_abci`, `chatui_abci`, `tx_settlement_multiplexer_abci`) all reference the updated child hashes.

Manual spot-curl of the 9 affected queries against the new endpoints is the one remaining unchecked item in the test plan, which is the right way to validate an endpoint cutover. No blocking concerns from static review.

mariapiamo (commented):
This migration preserves the current buy-only accounting model by filtering Polymarket bets with `isBuy: true`. That avoids signed sell rows breaking the existing positive-amount arithmetic, but it also means manual sells are not represented in prediction history or per-bet PnL.

The predict-polymarket subgraph does expose sells as `Bet` rows with `isBuy: false`, negative `amount`, and negative `shares`, so a manually sold/reduced position is detectable at the data layer. With this PR, those rows are intentionally excluded. As a result, a manually sold position may still be displayed/accounted from the original buy and final resolution/redeem state, without showing sale proceeds, reduced exposure, or a sold status.

I think this is acceptable for this minimal migration if we explicitly keep the current assumption: Polymarket performance summary is buy-only/redeem-only. But it would be good to document that manual sells are not yet modeled, or open a follow-up to add signed trade accounting for `isBuy: false` rows.

PR #921, opened 2026-04-28
fix(ci): restore check-agent-runner with path-based STORE_PATH override

LOCKhart07:
## Summary

Re-enables the `make check-agent-runner` smoke test that was muted in 308e7294. Surfaced as a follow-up commitment on PR #913 ([thread](https://github.com/valory-xyz/trader/pull/913#discussion_r3150192690)).

### Root cause

The break came from the anonymous env-var template flip in #919. With:

```yaml
# packages/valory/agents/trader/aea-config.yaml:232
store_path: ${str:/data/}
```

open-aea's `replace_with_env_var` only consults the **path-based** fallback (`SKILL_TRADER_ABCI_MODELS_PARAMS_ARGS_STORE_PATH=...`) for anonymous templates. The previous `STORE_PATH=/tmp` override bound to nothing, the template fell through to its YAML default `/data/`, and `staking_abci/get_store_path` raised `ValueError: The store path '/data/' is not a directory or is not accessible.`

The anonymous form has to stay — a named template (`${STORE_PATH:str:/data/}`) suppresses the path-based fallback and breaks Pearl's runtime injection. See valory-xyz/olas-operate-middleware#424 for the full story.

### Fix

- `check-agent-runner`: switch the override to the path-based key `SKILL_TRADER_ABCI_MODELS_PARAMS_ARGS_STORE_PATH` so it actually binds to the anonymous template.
- `build-agent-runner{,-mac}`: add `./dist/agent_runner_bin --help 1>/dev/null` before `--version` so a PyInstaller import-graph regression fails the build instead of slipping through to runtime. This is the "extra test that exercises the bundled import graph" that PR #913 reviewers asked us to keep.
- 3-line comment block in the recipe pointing at the Pearl PR so the next person doesn't "fix" the path-based key back to a named template.

### Targeting `polystrat-clob-v2`

Branched off `polystrat-clob-v2` and targets `polystrat-clob-v2` because:
- The anonymous template flip (#919) only lives on this branch — branching off `main` would either duplicate that change or apply the fix where the template is still named (so the path-based key wouldn't bind).
- Lands together with #913 once that PR merges.

## Test plan

- [x] `make -n check-agent-runner` renders the recipe correctly (no make-time `${...}` mangling in the comment block).
- [x] CI: `Test (build-agent-runner)` and `Test (check-agent-runner)` jobs pass on Linux + macOS + Windows.
- [ ] Local: `make build-agent-runner && make check-agent-runner` boots through to `Starting AEA` on at least one platform.


LOCKhart07:
Proof of a successfull build https://github.com/valory-xyz/trader/actions/runs/25041947749 https://github.com/valory-xyz/trader/releases/tag/0.35.0-test

OjusWiZard (dismissed):

bennyjo (dismissed):
LGTM — root cause analysis matches the YAML at aea-config.yaml:232 and the path-based key is the right fix. `--help` smoke addition is a reasonable guard against PyInstaller import-graph regressions.

OjusWiZard (approved):

bennyjo (approved):

PR #922, opened 2026-04-28
fix: switch to main clob url

LOCKhart07:
https://docs.polymarket.com/v2-migration

> Test against V2 before go-live. Point your client at https://clob-v2.polymarket.com to start integrating now. On April 28th (~11:00 UTC), V2 takes over the production URL — https://clob.polymarket.com — so you don’t need to change the base URL again after the cutover.


OjusWiZard (approved):

bennyjo (approved):
Approving — small mechanical host swap (`clob-v2.polymarket.com` → `clob.polymarket.com`) plus the cascading IPFS hash bumps. Verified no remaining `clob-v2` references in the repo and that `trader_pearl/service.yaml` has no host override (so the agent-hash bump alone is correct there). One non-blocking coordination note for the deployer: the migration doc says V2 takes over `clob.polymarket.com` at ~11:00 UTC on April 28, so make sure deployment lands after the cutover.

PR #929, opened 2026-04-29
feat(polymarket): route redeem through CtfCollateralAdapter for direct pUSD payout

LOCKhart07:
## Summary

The agent currently redeems Polymarket positions by calling the raw CTF / NegRiskAdapter with `pUSD` as `collateralToken`. The CTF stores positions under USDC.e, so `getPositionId(pUSD, …) ≠` the asset id the Safe actually holds — redeem silently no-ops, emits `PayoutRedemption` with `payout=0`, burns gas, doesn't burn tokens.

**This PR switches the redeem target to `CtfCollateralAdapter` (`0xADa1…9718`) for standard markets and `NegRiskCtfCollateralAdapter` (`0xAdA20…c6F1`) for neg-risk markets.** **Standard markets** call `redeemPositions(address,bytes32,bytes32,uint256[])` (selector `0x01b7037c`) on `CtfCollateralAdapter`. **Neg-risk markets** call the 2-arg `redeemPositions(bytes32,uint256[])` (selector `0xdbeccb23`) on `NegRiskCtfCollateralAdapter` — the same selector the old `NegRiskAdapter` used (preserves the per-outcome `uint256[] amounts` semantics that neg-risk needs and the 4-arg form cannot express). Bytecode probe confirms BOTH selectors are present on `0xAdA20…c6F1`, so the swap is target-address-only — calldata shape per branch is unchanged. The adapters unwrap USDC.e → pUSD inside the same tx, so the Safe receives **pUSD directly** with no downstream wrap step.

## Scope

What this PR redeems:

- Standard Polymarket positions whose CTF `positionId` is derived from **USDC.e** (`0x2791Bca1…84174`) — routed through `CtfCollateralAdapter`.
- Neg-risk Polymarket positions whose CTF `positionId` is derived from the legacy neg-risk wrapped collateral — routed through `NegRiskCtfCollateralAdapter`.

What this PR does **not** redeem:

- Positions whose CTF `positionId` was derived from **pUSD** (`0xC011a7E1…2DFB`). `CtfCollateralAdapter` hardcodes `USDCE` for `getPositionId` derivation and the underlying CTF `redeemPositions` call ([source L161-L164, L190-L198](https://github.com/Polymarket/ctf-exchange-v2/blob/main/src/adapters/CtfCollateralAdapter.sol#L161-L198)) and ignores any externally-passed collateral. A pUSD-derived `positionId` would not match anything either adapter can redeem.

Why pUSD-derived positions don't currently arise:

CTFExchange v2 (`0xE1111800…996B`) is constructed with two separate collateral parameters — `collateral` (= pUSD, the user-facing settlement token) and `ctfCollateral` (= USDC.e, used for CTF positionId derivation). The exchange unwraps pUSD → USDC.e before calling `splitPosition` on the CTF, so every CTF position created through v2 is USDC.e-derived as a contract-level invariant. Empirically verified on a representative agent trade ([`0xb5ca0dfb…46a7`](https://polygonscan.com/tx/0xb5ca0dfbc66e910fcc086f433694cf005e7c2188c23d22c22084d9078ee646a7)): the Safe acquired YES tokenId `0x6f159b…0659`, which matches `getPositionId(USDC.e, …)` exactly; `getPositionId(pUSD, …)` produces a completely different id that never appears in the tx.

If Polymarket ever ships a v3 exchange with a different `ctfCollateral` invariant, or a flow that splits pUSD-derived positions outside the exchange, this PR's adapter-only redeem path will miss those positions and namespace-detection-and-routing logic will need to be added.

## Empirical evidence motivating this change

Safe `0x9070A951…` redeemed Mar-a-Lago (Yes side, 8.88 tokens, market resolved Yes=$1):
- tx `0xcf86dfd4f30c371d784cee0ccf2030028331ff3fc05313399a644c59ee8f8911`
- status `0x1` (mined successfully)
- `PayoutRedemption.payout = 0` for both positions in the bundled redeem
- balance check after tx: Safe **still holds 8,880,000 of asset id `0xed53d02f…`** — tokens were never burned

The asset's positionId derivation only matches when computed against `USDC.e`, not pUSD. Discussion with `830Lincoln` confirmed: redeeming via raw CTF/NRA requires USDC.e; redeeming via the collateral adapters takes pUSD and returns pUSD.

## Approval expansion

Extends the existing `PolymarketSetApprovalRound` multisend from 6 → 8 entries:
| # | Token | Spender / Operator | Purpose |
|---|---|---|---|
| 7 | CTF `setApprovalForAll` | CtfCollateralAdapter | **redeem-critical** |
| 8 | CTF `setApprovalForAll` | NegRiskCtfCollateralAdapter | **redeem-critical** |

Only ERC-1155 operator rights are granted — the adapters' redeem path doesn't pull ERC-20 from the Safe, so no pUSD allowance is needed. (Earlier draft of this PR mirrored the Polymarket UI's split/merge batch and added two pUSD allowances; those were dropped in #930 — already merged into this branch.)

## Migration trigger

Bumps `POLYMARKET_ALLOWANCES_FILE_CLOB_VERSION` from `"v2"` to `"v3"` (`states/check_benchmarking.py:41`). Every running Safe sees its persisted allowances file as stale on first wake post-deploy and re-issues the expanded 8-tx approval set before any redeem runs. No manual ops needed per Safe.

## FSM ordering

`CheckBenchmarkingMode` (reads allowances file)
  → `SET_APPROVAL` if file is stale or missing → `PolymarketSetApprovalRound` (8-tx multisend) → settle → `PolymarketPostSetApprovalRound` (on-chain `_check_approval` over 5 ERC-1155 + 3 ERC-20 fields)
  → `WrapCollateral` → ... → `PolymarketRedeemRound` (now targeting the adapters, paying pUSD back to the Safe)

## Discarded alternatives

| Option | Why rejected |
|---|---|
| **A** — fire two redeems per position (one with pUSD, one with USDC.e), batched in same multisend | Doubles per-position gas with a guaranteed no-op call. |
| **B** — positionId-resolver: compute `getPositionId(candidate, …)` locally and pick the matching collateral | Still routes through raw CTF → payouts come back as USDC.e → Safe still depends on `PolymarketWrapCollateralRound` to wrap before next bet → one-cycle fund lag, dust-threshold strands small wins under `polymarket_usdc_e_wrap_dust_threshold` (`skill.yaml:376`). |
| **C** — pass USDC.e to raw CTF directly (one-line collateral swap) | Same wrap-round dependency as Option B; rejected for the same fund-lag and dust-threshold reasons. |
| **D — chosen** | One-line target swap + 2 new approvals, atomic, no FSM dependency on wrap-collateral for redeem outflow. |

## Test plan

- [x] All 1,254 existing decision_maker_abci + connection tests pass
- [x] `tomte check-code` clean (black, isort, flake8, mypy, pylint, darglint)
- [x] `autonomy packages lock` re-locks downstream skill/service hashes
- [x] New tests:
  - `test_prepare_approval_tx_success` — expects 8 batches (was 6)
  - `test_prepare_approval_tx_includes_ctf_collateral_adapter_approval` — asserts the redeem-critical `setApprovalForAll` entries
  - `test_prepare_approval_tx_excludes_collateral_allowance_for_adapters` — guards against re-introducing the dropped pUSD allowances (added in #930)
  - `test_standard_position_targets_ctf_collateral_adapter` — assert standard-market redeem target
  - `test_neg_risk_position_targets_neg_risk_ctf_collateral_adapter` — assert neg-risk redeem target
  - `test_end_block_polymarket_stale_v2_allowances_file_reapproves` — v2→v3 migration trigger
  - `test_clob_version_constant_is_v3` — pin the bumped constant
  - `TestSetApproval.test_includes_collateral_adapter_setapprovalforall` (connection-side)
  - `TestRedeemPositions.test_standard_market_targets_ctf_collateral_adapter` (connection-side)
  - `TestRedeemPositions.test_neg_risk_targets_neg_risk_ctf_collateral_adapter` (connection-side)
  - `TestCheckApproval` updated for 5 ERC-1155 + 3 ERC-20 checks (the original 3 collateral allowances; collateral adapters get only ERC-1155 operator rights)
- [x] **Pre-merge — standard market**: verified end-to-end on Polygon mainnet against Safe `0x9070A95134A06290141e94B6Dd453c9c88dc4BA4` — see [Live mainnet validation](#live-mainnet-validation) below. Redeem tx [`0xa476c971…296c0b`](https://polygonscan.com/tx/0xa476c971bea85be04a60bdf89de7a0d314c9635d10231179588a12962d296c0b): `PayoutRedemption.payout = 1,061,568` (1.061568 pUSD), YES position tokens burned, Safe received pUSD directly.
- [ ] **Pre-merge — neg-risk market** (raised in [review](https://github.com/valory-xyz/trader/pull/929#discussion_r3163380006)): the neg-risk path on `NegRiskCtfCollateralAdapter` is bytecode-validated (selector `0xdbeccb23` present) but transactionally untested in this PR. Run a redeem against a known neg-risk position to confirm the 2-arg path on the new adapter unwraps to pUSD as expected, before merging.

## Live mainnet validation

End-to-end run on Polygon mainnet, 2026-04-30, against Safe `0x9070A95134A06290141e94B6Dd453c9c88dc4BA4`. The agent picked up a real won position, prepared the redeem through the new adapter target, settled it via Safe, and the on-chain effects confirm the redesign works.

**Position picked up:**
- Market: *LoL: Gen.G vs Kiwoom DRX (BO3) - LCK Rounds 1-2*
- conditionId: `0xc0daba444380d10c9accbe87cd29ff31b7f8fac2bc52ab1d739692ad8331d0d0`
- size: 1.0615 YES tokens, market resolved YES=$1, `redeemable: true`, `negativeRisk: false` → standard CTF path

**Agent log excerpts** (`agent_log_latest.log`, 2026-04-30 19:49 UTC):

```
[19:49:27] Fetched 1 positions for 0x9070A95134A06290141e94B6Dd453c9c88dc4BA4
[19:49:27] Fetched 1 redeemable positions
[19:49:27] Polymarket builder program disabled - preparing redemption transaction...
[19:49:27] Preparing redeem tx for standard position:
           0xc0daba444380d10c9accbe87cd29ff31b7f8fac2bc52ab1d739692ad8331d0d0
           - Gen.G (size: 1.0615)
[19:49:29] 'polymarket_redeem_round' round is done with event: Event.PREPARE_TX
...
[19:49:35] Transaction sent! Received transaction digest:
           0xa476c971bea85be04a60bdf89de7a0d314c9635d10231179588a12962d296c0b
[19:49:49] Verified result: True
[19:49:49] Finalized with transaction hash:
           0xa476c971bea85be04a60bdf89de7a0d314c9635d10231179588a12962d296c0b
```

**Safe-tx calldata** confirms the redesign — target is `CtfCollateralAdapter` (not raw CTF), selector is the 4-arg `redeemPositions`:

```
target  = 0xada100874d00e3331d00f2007a9c336a65009718  ← CtfCollateralAdapter
selector= 0x01b7037c                                   ← redeemPositions(address,bytes32,bytes32,uint256[])
arg0    = 0xc011a7e12a19f7b1f670d46f03b03f3342e82dfb  ← pUSD (collateral parameter)
arg1    = 0x0...0                                      ← parentCollectionId
arg2    = 0xc0daba444380d10c9accbe87cd29ff31b7f8fac2bc52ab1d739692ad8331d0d0  ← conditionId
arg3    = [1, 2]                                       ← outcome index set
```

**On-chain receipt** ([Polygonscan](https://polygonscan.com/tx/0xa476c971bea85be04a60bdf89de7a0d314c9635d10231179588a12962d296c0b)) — block 86,219,367, status `0x1`, `gasUsed = 265,936`:

| log | event | meaning |
|---|---|---|
| 2345 | CTF `PayoutRedemption` | redeemer = adapter `0xada1…9718`, condition = `0xc0daba…d0d0`, **payout = `0x1032c0` = 1,061,568** (≈1.0615 pUSD at 6dp) |
| 2346 | CTF `TransferSingle` (ERC-1155) | YES positionId `0x6f159b…0659` burned (to `0x0`), amount `0x1032c0` |
| 2347 | USDC.e `Transfer` | CTF → adapter, `0x1032c0` (CTF pays adapter in USDC.e, as expected per the namespace invariant) |
| 2349 | USDC.e `Transfer` | adapter → pUSD contract, `0x1032c0` (adapter unwraps in-tx) |
| 2350 | pUSD `Transfer` | `0x0` → **Safe `0x9070…4BA4`**, `0x1032c0` (mint) |

**Post-redeem on-chain balance reads** (latest block 86,219,674, ~5 minutes after the tx):

| Read | Result | Verdict |
|---|---|---|
| `ConditionalTokens.balanceOf(safe, 0x6f159b…0659)` | `0` | YES position tokens fully burned ✓ |
| `pUSD.balanceOf(safe)` | `100,292,209` (= 100.292209 pUSD) | includes the 1.061568 from this redeem ✓ |

So: tokens burned, payout > 0, Safe received pUSD directly, no downstream wrap step needed. The Polymarket UI may still show a "redeem" button transiently — `data-api.polymarket.com/positions` is indexer-backed and lags Polygon by minutes; the on-chain truth is what's tabulated above.

## Lines of code breakdown

For reviewer triage — most of the diff is tests and lock-file noise, not logic:

| Category | Insertions | Deletions | Notes |
|---|---|---|---|
| **Real impl (production code)** | **171** | **30** | 7 files |
| Tests | 414 | 21 | 5 files |
| Hash propagation (autonomy packages lock) | 28 | 28 | 9 cascaded skill/service YAML hash swaps |
| **Total** | 613 | 79 | 21 files |

> **Note:** the totals above were captured before #930 merged into this branch. After the trim, the impl insertions for the dropped pUSD allowance entries (~55 lines on the two callsites and their `_check_approval` reads) come back out, and the test-side equivalents flip from positive-assertion to exclusion-guard. The post-#930 shape is what the diff currently shows; the breakdown stays a useful reviewer-triage map (impl vs tests vs hash propagation), even if the precise line counts are now smaller.

Post-#930 novel-logic surface area:
- 2 new approval entries × 2 callsites (`polymarket_set_approval.py:_prepare_approval_tx` and `connection.py:_set_approval`) — CTF `setApprovalForAll` for both collateral adapters
- 2 new params plumbed through skill.yaml + models.py + connection.yaml + connection.py `__init__`
- 2 new `_check_erc1155_approval` reads + dict entries in `connection.py:_check_approval`
- `target_address` swap in `polymarket_reedem.py` and `connection.py:_redeem_positions`
- `POLYMARKET_ALLOWANCES_FILE_CLOB_VERSION = "v3"`

So the *novel-logic* surface area is small — roughly 30 lines of new behavioural code, plus the address-list expansions that are mechanical. Reviewer effort should focus on `polymarket_reedem.py` (target swap) and `polymarket_set_approval.py:_prepare_approval_tx` (new approval entries).


packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
```diff
@@ -450,16 +450,21 @@ def _prepare_redeem_tx(
                     condition_id=condition_id,
                     redeem_amounts=redeem_amounts,
                 )
-                target_address = self.params.polymarket_neg_risk_adapter_address
+                target_address = (
```
jmoreira-valory, line 453: **Critical — neg-risk calldata vs PR description mismatch; untested empirically.**

The PR description says *"Both adapters expose the same `redeemPositions(address,bytes32,bytes32,uint256[])` ABI (selector `0x01b7037c`, verified by bytecode probe)"*. But the neg-risk branch here calls `_build_redeem_neg_risk_data`, which still emits the OLD 2-arg ABI:

```python
function_signature = "0xdbeccb23"  # keccak256("redeemPositions(bytes32,uint256[])")[:4]
```

…the same selector used against the previous `NegRiskAdapter`. The target address swapped from `NegRiskAdapter` to `NegRiskCtfCollateralAdapter`, but the calldata shape did not. (Same observation applies in `connection.py:_redeem_positions` neg-risk branch.)

Either:
- (a) The description is wrong and `NegRiskCtfCollateralAdapter` keeps the 2-arg signature like its predecessor — code is fine, description is misleading; or
- (b) The description is right and the neg-risk redeem will revert (function-not-found / silent revert) when the new adapter doesn't expose the 2-arg form.

The empirical evidence in the PR body (`0xcf86dfd4…` redeeming Mar-a-Lago) is a **standard** market, not neg-risk. The neg-risk path on the new adapter is **untested** in this PR.

Strong recommendation: bytecode-probe both `redeemPositions` selectors on `0xAdA20…c6F1` (or run a test redeem on a neg-risk market with the new adapter) **before merge**. If the adapter only exposes the 4-arg form, neg-risk redeems will silently revert and the agent will keep trying to redeem positions that can never burn.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -1321,6 +1384,8 @@ def _check_approval(self) -> Tuple[Any, Any]:
                         ctf_ctf_exchange_approved,
```
jmoreira-valory, line 1369: **Set/check asymmetry — pUSD allowances on the new adapters are issued but never verified.**

`_set_approval` now issues 4 new approvals (entries 7–10): pUSD allowance + CTF `setApprovalForAll` for each of `CtfCollateralAdapter` and `NegRiskCtfCollateralAdapter`. `_check_approval` adds the 2 new ERC-1155 checks (`ctf_collateral_adapter_approved`, `ctf_neg_risk_collateral_adapter_approved`) but does **not** add the matching ERC-20 allowance checks for the 2 new adapter pUSD approvals.

Net effect: after a successful set, `all_approvals_set = True` even if the pUSD allowances silently failed (RPC blip, partial revert, etc.). The cache file gets stamped as "v3 approvals OK" and the agent never re-issues them.

Either:
- add the 2 missing `_check_erc20_allowance` calls — matching the existing pattern for the 3 exchanges (CTF Exchange, NegRisk CTF Exchange, NegRisk Adapter); or
- document explicitly that the pUSD allowances are intentionally not required for redeem (per the existing comment, they're for split flows the agent doesn't perform) and explain why they're issued anyway / why they're safe to skip from the check.

The current state — issue, don't verify — is the worst of both: it commits gas to approvals nobody validates.

jmoreira-valory:
**PR description — small counting / body inconsistencies for reviewer triage.**

Two small things in the description that don't match the code (non-blocking, but worth fixing so future readers don't have to reconstruct the truth from the diff):

- *"`PolymarketPostSetApprovalRound` (on-chain `_check_approval` over 8 ERC-1155 + 5 ERC-20 fields)"* — actually it's **5 ERC-1155** (CTF Exchange, NegRisk CTF Exchange, NegRisk Adapter, CtfCollateralAdapter, NegRiskCtfCollateralAdapter) + **3 ERC-20** (the original 3 collateral allowances; the 2 new adapter pUSD allowances aren't checked — see [the inline comment on `_check_approval`](https://github.com/valory-xyz/trader/pull/929#discussion_r3163380006)).
- The "Approval expansion" table says all 4 new entries are needed, but only the `setApprovalForAll` rows (#8 and #10) are redeem-critical. The pUSD allowance rows (#7, #9) are split-flow future-proofing per the connection.py comment, and the agent doesn't perform splits. Worth marking those as *"future-use, mirrors UI batch"* in the table so a future reader doesn't burn cycles wondering why the agent approves something it doesn't use.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
LOCKhart07: Good catch — the description was misleading; the calldata on the neg-risk branch is unchanged from the previous `NegRiskAdapter` path (selector `0xdbeccb23` / 2-arg `redeemPositions(bytes32,uint256[])`). Bytecode probe at `0xAdA20…c6F1` confirms BOTH selectors are present:

```
0x01b7037c   YES   redeemPositions(address,bytes32,bytes32,uint256[])
0xdbeccb23   YES   redeemPositions(bytes32,uint256[])
```

The 2-arg form is the right choice for neg-risk because it lets us specify per-outcome `redeem_amounts = [yes_amount, no_amount]`, which the 4-arg `indexSets` form can't express. The new adapter's wrapping logic should apply to either selector since they're both entrypoints into the same redemption flow, but you're right that we don't have transaction-level proof of that yet — only that the function exists.

**Action taken:**
1. Updated PR description to explicitly call out per-branch ABI: standard = 4-arg on `CtfCollateralAdapter`; neg-risk = 2-arg on `NegRiskCtfCollateralAdapter`.
2. Added a separate pre-merge action item for empirical neg-risk verification — bytecode-validated isn't enough; we need a real neg-risk redeem against the new adapter to confirm the 2-arg path actually unwraps to pUSD.

Will not merge until that neg-risk verification lands.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: Real bug, fixed in c238b70b.

Added the two missing `_check_erc20_allowance` reads for the adapter pUSD allowances and included them in the `all_approvals_set` `all()` chain. New regression test `test_collateral_adapter_pusd_allowance_zero_marks_all_approvals_unset` ensures any future drift between the set side and the check side gets caught.

Resulting verify-side surface area: 5 ERC-1155 + 5 ERC-20 checks, matching the 10 entries `_set_approval` issues. PR description corrected to reflect this (was wrongly stated as 8 ERC-1155 + 5 ERC-20 in the FSM-ordering section).

LOCKhart07:
Both points addressed:

1. **Field count corrected.** PR description now says `5 ERC-1155 + 5 ERC-20`. The asymmetry you noticed (ERC-20 adapter checks missing) was a real bug; fixed in c238b70b — `_check_approval` now verifies all 10 entries `_set_approval` issues, with a regression test guarding the parity.

2. **Approval table re-marked.** Entries #7 and #9 (pUSD allowances to the collateral adapters) are now labeled `future-use (split/merge); mirrors UI batch 0xb8da243b…`. Only #8 and #10 (`setApprovalForAll`) are flagged as redeem-critical. Reasoning kept minimal: we mirror the empirical UI batch rather than skip them, because skipping risks a future split-flow change quietly failing on a dropped allowance — and the cost of an unlimited approval is one tx slot at deploy.

Thanks for the close read.

mariapiamo (commented):
I think the two new pUSD `approve(MAX_UINT256)` entries should be removed from this PR unless there is an immediate split/merge flow that uses them.

For the redeem bug being fixed here, the required permission is CTF ERC-1155 operator approval:

- `CTF.setApprovalForAll(CtfCollateralAdapter, true)`
- `CTF.setApprovalForAll(NegRiskCtfCollateralAdapter, true)`

That is needed because the adapter pulls the Safe’s CTF position tokens during redeem.

The pUSD approvals are different: they are only needed for adapter split/merge flows, where the adapter pulls pUSD from the Safe. The agent does not currently perform those flows in this PR. Mirroring the Polymarket UI batch explains why the entries are present, but it also expands the Safe’s standing fund permission surface beyond the redeem fix.

Recommendation: keep the two redeem-critical `setApprovalForAll` entries, remove the two pUSD adapter allowances and their matching checks, and add pUSD approval later if/when split/merge is implemented, ideally scoped to the amount or bundled with that flow.

mariapiamo (commented):
If this PR fully migrates neg-risk redeem away from NegRiskAdapter, why do we still require pUSD allowance and CTF operator approval for NegRiskAdapter in the approval gate?

mariapiamo (changes requested):
I think we should be precise about what this PR fixes and what it does not make universal.

The PR changes all Polymarket redeem paths to use the collateral adapters:

- standard markets -> `CtfCollateralAdapter`
- neg-risk markets -> `NegRiskCtfCollateralAdapter`


This is correct for the observed bug, where the Safe held legacy/USDC.e-derived CTF positions but the old code redeemed raw CTF with pUSD. The adapter fixes that because it internally uses the legacy collateral namespace and then returns pUSD.

However, this does **not** create a universal redeem path for every possible collateral namespace.

For standard markets, `CtfCollateralAdapter.redeemPositions(...)` ignores the externally passed collateral address: https://github.com/Polymarket/ctf-exchange-v2/blob/main/src/adapters/CtfCollateralAdapter.sol#L117-L126

Then it computes position IDs with `USDCE`: https://github.com/Polymarket/ctf-exchange-v2/blob/main/src/adapters/CtfCollateralAdapter.sol#L161-L164

And the actual raw CTF redeem also uses `USDCE`: https://github.com/Polymarket/ctf-exchange-v2/blob/main/src/adapters/CtfCollateralAdapter.sol#L190-L198

For neg-risk markets, the collateral adapter similarly uses the legacy neg-risk wrapped collateral, not an arbitrary externally selected collateral namespace: https://github.com/Polymarket/ctf-exchange-v2/blob/main/src/adapters/NegRiskCtfCollateralAdapter.sol#L139-L162

So the current PR can redeem:

- standard Polymarket positions whose CTF position IDs are derived from USDC.e
- neg-risk Polymarket positions whose IDs are derived from the legacy neg-risk wrapped collateral
- and returns the payout to the Safe as pUSD

But it cannot redeem hypothetical/direct raw CTF positions whose IDs were actually derived from pUSD: positionId(pUSD, condition, outcome)

If we need a universal redeem flow, the agent should first check which position ID the Safe actually holds, then choose the redeem route:

if balanceOf(Safe, positionId(USDC.e, ...)) > 0:
    redeem via CtfCollateralAdapter

if balanceOf(Safe, positionId(pUSD, ...)) > 0:
    redeem raw CTF with pUSD

if neg-risk legacy wrapped position exists:
    redeem via NegRiskCtfCollateralAdapter

LOCKhart07:
Agreed — the pUSD allowances aren't redeem-critical and shouldn't sit on the Safe just to mirror the UI batch.

Dropped both pUSD entries (and their corresponding `_check_approval` reads / `all_approvals_set` chain entries) in **#930**, stacked on this branch. Approval batch shrinks from 10 → 8 entries; the two `CTF.setApprovalForAll(adapter, true)` entries (the only redeem-critical part) stay.

No `POLYMARKET_ALLOWANCES_FILE_CLOB_VERSION` bump there — already-deployed Safes carry harmless leftover pUSD allowances rather than burning gas through the relayer to revoke them. If you'd prefer an active revoke flow for those Safes, happy to add it as a separate change.

#930 also adds a regression-guard test (`test_prepare_approval_tx_excludes_collateral_allowance_for_adapters`) to keep the entries from creeping back in.

LOCKhart07:
`NegRiskAdapter` is kept in the approval gate because it's still used for **trading** neg-risk markets, not just redeem. Order placement / cancellation against neg-risk markets goes through `NegRiskAdapter` — the v1 ERC-1155 setApprovalForAll and pUSD allowance for that contract are how the Safe authorizes the adapter to move tokens during fills.

What this PR migrated was specifically the **redeem** target: from the raw CTF / `NegRiskAdapter.redeemPositions(...)` to the `*CtfCollateralAdapter` pair, which auto-unwraps to pUSD. Trading still flows through `NegRiskAdapter`, so its approvals stay.

If you'd like, I can rename the gate fields in `_check_approval` to make the trading-vs-redeem split clearer (e.g. `neg_risk_adapter_trading` vs `neg_risk_ctf_collateral_adapter_redeem`) — happy to do that as a separate cleanup if it'd help future readers.

LOCKhart07:
Agreed on the precision point — "this PR fixes the bug" is a stronger claim than "this PR is a universal CTF-redeem layer." Your reading of the adapter source is correct: `CtfCollateralAdapter` ignores the externally-passed collateral arg and hardcodes `USDCE` for both `getPositionId` derivation and the underlying CTF redeem; `NegRiskCtfCollateralAdapter` hardcodes the legacy neg-risk wrapped collateral. So the adapter only redeems positions whose `positionId` was derived from those specific namespaces.

Where I want to push back is on whether the pUSD-derived positionId case is something the agent can actually encounter today.

## Empirical receipt

For a representative agent trade — Safe `0x9070A951…dc4ba4` BUY-YES via `CTFExchange.matchOrders(...)` in tx [`0xb5ca0dfb…46a7`](https://polygonscan.com/tx/0xb5ca0dfbc66e910fcc086f433694cf005e7c2188c23d22c22084d9078ee646a7):

- Observed YES tokenId acquired by the Safe: `0x6f159b13bd521bb1adb40859b8c4809b46edb4fec5b80958f4928a47e5840659`
- Direct CTF call, `getPositionId(USDC.e, …)` → `0x6f159b…0659` ✅ exact match
- Direct CTF call, `getPositionId(pUSD, …)` → `0xebf102…7843` ❌ doesn't appear anywhere in the tx

The `PositionSplit` event on the CTF in this tx names `collateralToken = USDC.e (0x2791Bca1…84174)` directly. A pUSD-derived positionId would have been a completely different token id, and that token id never shows up.

## Why this generalizes (under v2 invariants)

This isn't a coincidence of one trade. The new `CTFExchange` v2 (`0xE1111800…996B`, verified source on Polygonscan) takes **two** collateral parameters at construction time:

```solidity
address collateral;     // settlement currency users pay/receive in  -> pUSD
address ctfCollateral;  // collateral used for CTF positionId derivation
                        // "May differ from `collateral` when using adapters
                        //  (e.g. USDC.e vs CollateralToken)"
```

The exchange itself owns the pUSD ↔ USDC.e adapter logic at the protocol level: users settle in pUSD, the exchange unwraps to USDC.e before calling `splitPosition` on the CTF, then wraps back to pUSD on the merge/redeem side. Every CTF position created through this exchange is USDC.e-derived as a contract-level invariant — and incidentally, the unwrap path the exchange itself routes through is `0xada100…9718`, which is the same `CtfCollateralAdapter` we use for redeem here.

A pUSD-derived `positionId` would only arise if someone called `ConditionalTokens.splitPosition(pUSD, …)` outside the exchange (i.e. directly against the CTF). The agent doesn't do that; no Polymarket flow we know of does either.

## Honest framing

"Universal across all theoretical CTF collateral namespaces" — no, the PR isn't that, and shouldn't claim to be.

"Universal across all positions any agent acquires through Polymarket as of v2" — yes, by virtue of the v2 exchange's `ctfCollateral=USDC.e` invariant. If Polymarket ever ships a v3 exchange with a different `ctfCollateral`, or starts doing pUSD-direct CTF splits outside the exchange, the adapter-only redeem path will miss those positions and we'd need the namespace-detection logic you sketched.

My proposal:

- **Don't** add the namespace detection in this PR (would be code for a path that doesn't exist today; the "don't design for hypothetical future requirements" preference).
- **Do** add an explicit "Scope" section to the PR description spelling out: this PR redeems positions in the USDC.e namespace (CtfCollateralAdapter) and the legacy neg-risk wrapped namespace (NegRiskCtfCollateralAdapter); pUSD-derived CTF positions are out of scope and don't currently arise from any Polymarket flow.
- **If** Polymarket ships a v3 exchange that breaks the invariant, file a follow-up to add namespace detection then.

Updating the PR description now. If you'd rather have the detection logic land here despite the path being currently unreachable, say the word and I'll spec it out.

mariapiamo:
> `NegRiskAdapter` is kept in the approval gate because it's still used for **trading** neg-risk markets, not just redeem. Order placement / cancellation against neg-risk markets goes through `NegRiskAdapter` — the v1 ERC-1155 setApprovalForAll and pUSD allowance for that contract are how the Safe authorizes the adapter to move tokens during fills.
> 
> What this PR migrated was specifically the **redeem** target: from the raw CTF / `NegRiskAdapter.redeemPositions(...)` to the `*CtfCollateralAdapter` pair, which auto-unwraps to pUSD. Trading still flows through `NegRiskAdapter`, so its approvals stay.
> 
> If you'd like, I can rename the gate fields in `_check_approval` to make the trading-vs-redeem split clearer (e.g. `neg_risk_adapter_trading` vs `neg_risk_ctf_collateral_adapter_redeem`) — happy to do that as a separate cleanup if it'd help future readers

Great! No need as of now

mariapiamo:
> Agreed on the precision point — "this PR fixes the bug" is a stronger claim than "this PR is a universal CTF-redeem layer." Your reading of the adapter source is correct: `CtfCollateralAdapter` ignores the externally-passed collateral arg and hardcodes `USDCE` for both `getPositionId` derivation and the underlying CTF redeem; `NegRiskCtfCollateralAdapter` hardcodes the legacy neg-risk wrapped collateral. So the adapter only redeems positions whose `positionId` was derived from those specific namespaces.
> 
> Where I want to push back is on whether the pUSD-derived positionId case is something the agent can actually encounter today.
> 
> ## Empirical receipt
> For a representative agent trade — Safe `0x9070A951…dc4ba4` BUY-YES via `CTFExchange.matchOrders(...)` in tx [`0xb5ca0dfb…46a7`](https://polygonscan.com/tx/0xb5ca0dfbc66e910fcc086f433694cf005e7c2188c23d22c22084d9078ee646a7):
> 
> * Observed YES tokenId acquired by the Safe: `0x6f159b13bd521bb1adb40859b8c4809b46edb4fec5b80958f4928a47e5840659`
> * Direct CTF call, `getPositionId(USDC.e, …)` → `0x6f159b…0659` ✅ exact match
> * Direct CTF call, `getPositionId(pUSD, …)` → `0xebf102…7843` ❌ doesn't appear anywhere in the tx
> 
> The `PositionSplit` event on the CTF in this tx names `collateralToken = USDC.e (0x2791Bca1…84174)` directly. A pUSD-derived positionId would have been a completely different token id, and that token id never shows up.
> 
> ## Why this generalizes (under v2 invariants)
> This isn't a coincidence of one trade. The new `CTFExchange` v2 (`0xE1111800…996B`, verified source on Polygonscan) takes **two** collateral parameters at construction time:
> 
> ```solidity
> address collateral;     // settlement currency users pay/receive in  -> pUSD
> address ctfCollateral;  // collateral used for CTF positionId derivation
>                         // "May differ from `collateral` when using adapters
>                         //  (e.g. USDC.e vs CollateralToken)"
> ```
> 
> The exchange itself owns the pUSD ↔ USDC.e adapter logic at the protocol level: users settle in pUSD, the exchange unwraps to USDC.e before calling `splitPosition` on the CTF, then wraps back to pUSD on the merge/redeem side. Every CTF position created through this exchange is USDC.e-derived as a contract-level invariant — and incidentally, the unwrap path the exchange itself routes through is `0xada100…9718`, which is the same `CtfCollateralAdapter` we use for redeem here.
> 
> A pUSD-derived `positionId` would only arise if someone called `ConditionalTokens.splitPosition(pUSD, …)` outside the exchange (i.e. directly against the CTF). The agent doesn't do that; no Polymarket flow we know of does either.
> 
> ## Honest framing
> "Universal across all theoretical CTF collateral namespaces" — no, the PR isn't that, and shouldn't claim to be.
> 
> "Universal across all positions any agent acquires through Polymarket as of v2" — yes, by virtue of the v2 exchange's `ctfCollateral=USDC.e` invariant. If Polymarket ever ships a v3 exchange with a different `ctfCollateral`, or starts doing pUSD-direct CTF splits outside the exchange, the adapter-only redeem path will miss those positions and we'd need the namespace-detection logic you sketched.
> 
> My proposal:
> 
> * **Don't** add the namespace detection in this PR (would be code for a path that doesn't exist today; the "don't design for hypothetical future requirements" preference).
> * **Do** add an explicit "Scope" section to the PR description spelling out: this PR redeems positions in the USDC.e namespace (CtfCollateralAdapter) and the legacy neg-risk wrapped namespace (NegRiskCtfCollateralAdapter); pUSD-derived CTF positions are out of scope and don't currently arise from any Polymarket flow.
> * **If** Polymarket ships a v3 exchange that breaks the invariant, file a follow-up to add namespace detection then.
> 
> Updating the PR description now. If you'd rather have the detection logic land here despite the path being currently unreachable, say the word and I'll spec it out.

Good! I agree with your proposal

OjusWiZard (dismissed):

mariapiamo (dismissed):
Approved from my side.

I re-checked the latest head and the implementation now grants the minimal new approval set needed for the redeem fix: only the two additional CTF `setApprovalForAll` approvals for `CtfCollateralAdapter` and `NegRiskCtfCollateralAdapter`. The previous pUSD `approve(MAX_UINT256)` entries to the collateral adapters are no longer present, which addresses the standing allowance concern.

The configured adapter addresses also match the official Polymarket contracts, and the redeem target changes are consistent with the v2 CLOB invariant: user-facing pUSD, but USDC.e / legacy wrapped collateral used for CTF position accounting.

Two non-blocking points before merge:

1. The PR body still has stale references to a 10-entry approval batch and pUSD adapter approvals. Current code is 8 entries, so the description should be updated to avoid future reviewer/operator confusion.

2. The neg-risk path should still get transaction-level validation. I would test it on a Tenderly fork using a real neg-risk position. Bytecode/source checks show the selector exists, but a real forked redeem is the stronger confirmation that the 2-arg neg-risk path unwraps and pays back pUSD as expected.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_set_approval.py
```diff
@@ -150,10 +150,18 @@ def _prepare_approval_tx(self) -> Generator[None, None, Optional[str]]:
...
+            self.params.polymarket_neg_risk_ctf_collateral_adapter_address
+        )
 
         # Build approval transactions and add to multisend_batches (must match
-        # polymarket_client _check_approval: 3 collateral allowances + 3 CTF
-        # setApprovalForAll).
+        # polymarket_client _check_approval: 3 collateral allowances + 5 CTF
+        # setApprovalForAll; 8 entries total). The collateral adapters
+        # intentionally receive only ERC-1155 operator rights — their redeem
+        # path doesn't pull ERC-20 from the Safe, so no pUSD allowance.
```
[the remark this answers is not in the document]
LOCKhart07: Addressed. The in-code comment at this line was already updated in #930 (now reads `8 entries total`); the matching staleness in the PR description (`Approval expansion` table claiming `6 → 10 entries` with 4 new rows) is now corrected to `6 → 8 entries` with only the 2 redeem-critical `setApprovalForAll` rows. The two pUSD allowances were dropped — the adapters' redeem path doesn't pull ERC-20 from the Safe.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_set_approval.py
```diff
@@ -364,8 +366,143 @@ def mock_build_multisend_safe_tx_hash() -> None:  # type: ignore[no-untyped-def,
...
 
         assert result == "0xfinalHash"
-        # Should have 6 batches: 3 USDC approves + 3 CTF setApprovalForAll
-        assert len(behaviour.multisend_batches) == 6
+        # 8 batches: 3 collateral approves for the v1 CLOB Exchange + NegRisk
+        # Exchange + NegRisk Adapter, 3 CTF setApprovalForAll for the same,
+        # plus 2 CTF setApprovalForAll for CtfCollateralAdapter and
+        # NegRiskCtfCollateralAdapter. The collateral adapters intentionally
+        # receive no pUSD allowance — their redeem path doesn't pull
+        # ERC-20 from the Safe.
```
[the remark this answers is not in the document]
LOCKhart07: Fixed in f6f591df — the comment now correctly says `v2 CTF Exchange` (the address mocked here is `polymarket_ctf_exchange_address`, which in production is `0xE111…996B`, the v2 CTFExchange — the agent uses `py_clob_client_v2`).

OjusWiZard (approved):

bennyjo (approved):

PR #931, opened 2026-04-30
fix(polymarket): migrate /events to keyset pagination

LOCKhart07:
## Summary
- Replace offset-based pagination on `GET /events` with cursor-based `GET /events/keyset` in `_fetch_markets_by_tag_slug` (the only paginated gamma-api caller in this repo).
- Polymarket removes legacy offset endpoints on **2026-05-01** (per the [Apr 10, 2026 changelog](https://docs.polymarket.com/changelog#apr-10-2026)). After that date, sending `offset` returns 422 and breaks the market sweep.

## What changed
- `connection.py` — send `after_cursor` instead of `offset`, hit `/events/keyset`, unwrap `{"events": [...], "next_cursor": "..."}`, stop when `next_cursor` is omitted (documented end-of-pages signal).
- `EVENTS_LIMIT=200` retained — well under the new endpoint's 500 cap.
- Tests updated to the wrapped response shape; one renamed to `test_paginates_until_next_cursor_absent` and now also asserts cursor forwarding on the second call.
- Singleton lookups (`/markets/slug/{slug}`, `/markets/{id}`) are unaffected — they aren't paginated.

## Test plan
- [x] `uv run pytest packages/valory/connections/polymarket_client/tests/test_connection.py` — 160 passed
- [x] 100% coverage on `polymarket_client/connection.py`
- [x] `uv run tomte check-code` — black, isort, flake8, mypy, pylint, darglint all green
- [x] `uv run autonomy packages lock` — hashes regenerated
- [x] Live agent run against the new endpoint: 9 categories iterated cleanly, 4 returned data (politics 309, science 28, weather 1331, finance 1126 markets), 5 returned 0 events (no errors); 2725 markets fetched / 2697 unique after dedup; downstream FSM, position fetches, and BUY orders all proceeded normally.

## Caveat
Cursor-forwarding (multi-page walk) is exercised by tests but the live run hit only single-page categories. The first sweep that exceeds 200 events in one tag will exercise it for real.


OjusWiZard (approved):

LOCKhart07:
## Local verification — keyset pagination + date filter

Follow-up to the [Caveat](https://github.com/valory-xyz/trader/pull/931#issue-description) that multi-page cursor-walking wasn't exercised in the live run. Tested locally on `2026-05-01`:

### 1. Wire format — raw HTTP to `/events/keyset`
Confirmed the response shape matches the parser. With `tag_slug=weather&limit=5`, walked 4 pages, each returning `{"events": [...], "next_cursor": "..."}` with cursors forwarded correctly.

### 2. Multi-page walk via `_fetch_markets_by_tag_slug`
Bypassed `__init__` and called the method directly with `EVENTS_LIMIT = 5` to force pagination on a small page size:

```
  Fetched 5 events → 34 markets (total: 34)
  Fetched 5 events → 13 markets (total: 47)
  Fetched 5 events → 24 markets (total: 71)
  ...
  Fetched 5 events → 55 markets (total: 14462)
```

Walked 300+ pages cleanly with monotonically climbing totals — the cursor branch is unambiguously triggered page after page, no errors. **Resolves the caveat.**

### 3. Date filter is honored on `/events/keyset`
Concern: if `end_date_min`/`end_date_max` were silently ignored under keyset, every sweep would page through the entire corpus.

| Window | Result |
|---|---|
| No filter | 10 events, all from 2024 (oldest-first ordering) |
| `2026-04-01..2026-06-01` | 10 events, every `endDate` falls inside the window |
| `2099-01-01..2099-12-31` | 0 events, no `next_cursor` (clean empty response) |

Filter is applied correctly. The production `weather` log showing only 180 events with `EVENTS_LIMIT = 200` is consistent — the date window narrows the corpus before pagination, and that category genuinely fits in one page.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -570,27 +569,27 @@ def _fetch_markets_by_tag_slug(
...
+            response, error = self._request_with_retries(
+                f"{GAMMA_API_BASE_URL}/events/keyset", params=params
             )
 
             if error:
                 return None, error
 
-            if not events_data:
-                break
+            events_data = response.get("events") or []
```
bennyjo, line 592: Two robustness-vs-upstream concerns worth a follow-up PR — same shape (less defensive against the server misbehaving than the old offset-based loop):

1. **`response.get("events")` here can `AttributeError` if `_request_with_retries` returns a non-dict.** That helper is typed `Tuple[Any, str]` and just returns `response.json()` — a JSON `null`, a stray list, or a string from a misbehaving edge proxy would crash here and bubble up to `fetch_markets`'s blanket `except Exception`, dropping the whole category. The old code's `if not events_data: break` degraded gracefully on the same inputs. Cheap fix: `if not isinstance(response, dict): break` (or `(response or {}).get("events")`).

2. **No client-side termination bound on the cursor loop** (visible at L609–611). The old code had `len(events_data) < EVENTS_LIMIT: break` as a natural ceiling; the new code relies entirely on the server eventually returning a falsy `next_cursor`. If the upstream ever returns the same cursor in a row (server bug, cache poisoning, partial outage), `all_markets` grows unbounded until OOM. Worth a `for _ in range(MAX_PAGES)` outer with a logged warning on cap — the agent runs unattended.

Both are follow-up-eligible — flagging because the migration changes the failure mode here from "stop paginating gracefully" to "raise / loop forever."

bennyjo (approved):
Approving — migration is correct and the deadline argues for shipping. Two robustness items inline that I'd suggest carrying into a follow-up PR rather than blocking this one.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: https://github.com/valory-xyz/trader/issues/936

PR #933, opened 2026-04-30
fix(perf-summary): rebuild profit_over_time on schema cutover

LOCKhart07:
## Summary

Adds a one-shot rebuild gate for `agent_performance.json`'s `profit_over_time` series when the on-disk schema version is below the current value.

## Why

After the predict-* subgraph cutover (#918), pre-cutover days written under the legacy/staging endpoints' redemption-date attribution don't reconcile with new days written under prod's settlement-date attribution. Result: the cached cumulative profit series goes negative for users with unredeemed wins, even though top-down ROI (recomputed every cycle) is correct. Verified empirically — Dima's agent showed ROI = +15.51% / +\$4.79 USD while the cached lifetime series ended at −\$5.87 USD; deleting the cache and letting it backfill fixed it.

Verified via the prod endpoint that `DailyProfitStatistic.dailyProfit` summed across all days equals \`totalExpectedPayout − totalTradedSettled\` to the cent — the field is correct, only the historical mix of two endpoint regimes is broken.

## What

- New module-level constant `PROFIT_OVER_TIME_SCHEMA_VERSION = 1` in `models.py`. Bump to invalidate again on future cutovers.
- New field `schema_version: int = 0` on `ProfitOverTimeData`. Default 0 = legacy/missing → triggers rebuild on first run.
- Fourth gate in `_build_profit_over_time_data` mirroring the existing `settled_mech_request_count` and `includes_unplaced_mech_fees` gates: `stored_version < CURRENT` → `_perform_initial_backfill`.
- Three `ProfitOverTimeData(...)` construction sites updated to set `schema_version=PROFIT_OVER_TIME_SCHEMA_VERSION` so the rebuilt/incremental output marks itself current.

Scoped to `profit_over_time` only — the other `AgentPerformanceSummary` fields (`metrics`, `agent_details`, `agent_performance`, `prediction_history`, `achievements`) are recomputed every cycle from the live subgraph and don't carry stale data.

## Test plan

- [x] \`uv run pytest packages/valory/skills/agent_performance_summary_abci/tests/\` — 968 pass
- [x] Coverage on \`behaviours.py\` and \`models.py\` — 100% (760/760 stmts + 246/246 branches; 184/184 + 30/30)
- [x] \`uv run tomte check-code\` — black, isort, flake8, mypy, pylint, darglint all green
- [x] \`uv run autonomy packages sync --update-packages && uv run autonomy packages lock\` — hashes regenerated against \`bump-mech-interact\` base
- [x] Manual: run an agent with a pre-existing \`agent_performance.json\` and confirm one-shot backfill log line + correct cumulative on next render


OjusWiZard (dismissed):

OjusWiZard (approved):

DhairyaPatel7 (approved):

packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py
```diff
@@ -3090,6 +3092,44 @@ def test_missing_unplaced_mech_fees_triggers_backfill(self) -> None:
...
+                ProfitDataPoint(
+                    date="2023-11-14",
+                    timestamp=1700000000,
+                    daily_profit=1.0,
+                    cumulative_profit=1.0,
+                )
+            ],
+            settled_mech_requests_count=5,
+            includes_unplaced_mech_fees=True,
+            schema_version=PROFIT_OVER_TIME_SCHEMA_VERSION - 1,
```
bennyjo, line 3113: Have you considered using a value that doesn't coincide with the dataclass default? With `PROFIT_OVER_TIME_SCHEMA_VERSION = 1`, `SCHEMA_VERSION - 1` evaluates to `0`, which is also what the field defaults to when a legacy `agent_performance.json` is loaded without a `schema_version` key. So this test asserts that "schema_version explicitly set to 0 triggers rebuild" — it does not separately exercise "field absent from the JSON file," which is the actual production migration case for every existing user.

Two consequences worth thinking about:
1. The moment the constant gets bumped to `2`, this test starts using `schema_version=1` and stops covering the legacy-no-field path entirely.
2. There's no direct assertion that the `ProfitOverTimeData` returned by `_perform_initial_backfill` carries `schema_version=PROFIT_OVER_TIME_SCHEMA_VERSION` after the rebuild — it's currently only exercised implicitly by the construction sites.

A small addition that would harden both: a test that builds the existing summary from a JSON dict literal with no `schema_version` key (mimicking the on-disk legacy file) and asserts the post-backfill result has `schema_version == PROFIT_OVER_TIME_SCHEMA_VERSION`. Happy to leave as-is if this is deliberate — flagging because the legacy-file path is the one that will actually fire on every existing agent.

bennyjo (commented):
Scope and intent match the linked issue; logic, gate ordering, and failure-path semantics look right. One question on the new test's discrimination — see inline.

packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py
LOCKhart07: Opened https://github.com/valory-xyz/trader/issues/934

PR #935, opened 2026-05-01
fix(polymarket): bump collateral adapter addresses (deadline 2026-05-01 15:00 UTC)

LOCKhart07:
## Summary
- Swap the two Polymarket collateral-adapter addresses to the new contracts. The relayer stops accepting transactions through the old adapters on **2026-05-01 15:00 UTC**, after which `redeem`/`split`/`merge` would silently fail.
- `CtfCollateralAdapter`: `0xADa1…9718` → `0xAdA1…cE1f`
- `NegRiskCtfCollateralAdapter`: `0xAdA2…c6F1` → `0xadA2…eAab`
- ABIs (`redeemPositions`, `splitPosition`, `mergePositions`, `setApprovalForAll` operator usage) are unchanged on the new contracts. We use the adapters as write targets only — no event subscriptions or log decoding — so the new events the contracts emit don't affect us. Address-only swap is sufficient.
- Follow-up to #929, which shipped with the old `CtfCollateralAdapter`.

Source: [Polymarket docs — Collateral Contracts](https://docs.polymarket.com/resources/contracts#collateral-contracts).

## Files changed
- `packages/valory/skills/decision_maker_abci/skill.yaml`
- `packages/valory/skills/trader_abci/skill.yaml`
- `packages/valory/connections/polymarket_client/connection.yaml`
- `packages/valory/connections/polymarket_client/tests/test_connection.py`

## Test plan
- [x] `tomte check-code` passes
- [x] `autonomy packages lock` (separate commit before merge)
- [x] `pytest packages/valory/connections/polymarket_client/tests/ -v`
- [x] `pytest packages/valory/skills/decision_maker_abci/tests/ -v`
- [ ] Manual: redeem on Polymarket via the new adapters succeeds end-to-end after merge


bennyjo (dismissed):
Approving the diff — clean, narrowly-scoped address swap with consistent coverage across all 4 occurrences (repo-wide grep for the old addresses returns zero, for the new ones returns exactly the four updated lines). The two new addresses match the Polymarket docs page exactly (https://docs.polymarket.com/resources/contracts#collateral-contracts), so address provenance is confirmed.

One confirmation before merge: `autonomy packages lock` lands before merge, as called out in the PR description — without it, `packages/packages.json` keeps the pre-bump hashes for `polymarket_client`, `decision_maker_abci`, and `trader_abci`.

One note for the future, not a blocker: `test_standard_market_targets_ctf_collateral_adapter` and `test_neg_risk_targets_neg_risk_ctf_collateral_adapter` (test_connection.py:1727, 1748) compare `tx.to` to the same constants that `_make_connection()` seeds onto the connection, so they're tautological with respect to `connection.yaml` — a typo in the YAML wouldn't fail them. Pre-existing, but worth knowing for the next address bump.

packages/valory/skills/decision_maker_abci/skill.yaml
```diff
@@ -378,8 +378,8 @@ models:
       polymarket_ctf_exchange_address: '0xE111180000d2663C0091e4f400237545B87B996B'
       polymarket_neg_risk_ctf_exchange_address: '0xe2222d279d744050d28e00520010520000310F59'
       polymarket_neg_risk_adapter_address: '0xd91E80cF2E7be2e162c6513ceD06f1dD0dA35296'
-      polymarket_ctf_collateral_adapter_address: '0xADa100874d00e3331D00F2007a9c336a65009718'
-      polymarket_neg_risk_ctf_collateral_adapter_address: '0xAdA200001000ef00D07553cEE7006808F895c6F1'
+      polymarket_ctf_collateral_adapter_address: '0xAdA100Db00Ca00073811820692005400218FcE1f'
+      polymarket_neg_risk_ctf_collateral_adapter_address: '0xadA2005600Dec949baf300f4C6120000bDB6eAab'
```
bennyjo, line 382: Should this PR also bump `POLYMARKET_ALLOWANCES_FILE_CLOB_VERSION` from `"v3"` to `"v4"` in `packages/valory/skills/decision_maker_abci/states/check_benchmarking.py:43`?

My read of the flow: an existing agent has `polymarket.json` stamped `allowances_set=true, clob_version="v3"`. After this PR, `CheckBenchmarkingModeRound.end_block` (`check_benchmarking.py:76-83`) sees the stamp still matches the current `"v3"`, returns `BENCHMARKING_DISABLED`, and skips `PolymarketSetApprovalRound` entirely — so the on-chain `_check_approval` in `connection.py:1338-1345` never runs either. Doesn't that mean `setApprovalForAll(NEW_ADAPTER, true)` is never granted on the CTF, and the first `redeemPositions` call against the new adapter reverts for every pre-existing deployment? The comment at `check_benchmarking.py:38-43` was added the last time these adapters changed and seems to anticipate exactly this case.

If that's right, would you also be open to a small regression test on `CheckBenchmarkingModeRound.end_block` asserting that `allowances_set=true, clob_version="v3"` returns `Event.SET_APPROVAL` once the stamp is bumped? Happy to be wrong about the flow if I'm missing where the new approvals get granted.

bennyjo (commented):
One blocking item: the on-disk allowances stamp wasn't bumped, so existing agents will skip the approval round and revert at redeem.

packages/valory/skills/decision_maker_abci/skill.yaml
LOCKhart07: Good catch — confirmed the flow exactly as you described. Fixed in 2c1812c3:

- Bumped `POLYMARKET_ALLOWANCES_FILE_CLOB_VERSION` from `"v3"` → `"v4"` in `check_benchmarking.py:43`, with a comment block citing the adapter swap and the 2026-05-01 15:00 UTC deadline as the reason.
- Added `test_end_block_polymarket_stale_v3_allowances_file_reapproves` mirroring the existing v1/v2 stale-stamp regression tests — asserts `allowances_set=true, clob_version="v3"` returns `Event.SET_APPROVAL` after the bump.
- Updated the constant-pin test from `v3` → `v4` to keep a silent revert from sneaking back in.

`tomte check-code` clean; all 11 tests in `test_check_benchmarking.py` pass.

bennyjo (approved):
Re-reviewed at 59455134. Blocker addressed cleanly:

- `POLYMARKET_ALLOWANCES_FILE_CLOB_VERSION` bumped v3 → v4 in check_benchmarking.py:43 with a comment citing the adapter swap and the 2026-05-01 deadline.
- New regression test `test_end_block_polymarket_stale_v3_allowances_file_reapproves` (test_check_benchmarking.py:165) writes a stale v3 stamp and asserts `Event.SET_APPROVAL` — would have caught the bug.
- Constant-pin test updated to v4 (test_check_benchmarking.py:150), so a silent revert to v3 would fail loudly.
- Repo-wide grep confirms no other production reference pins to "v3"; the `polymarket_post_set_approval.py` writer uses the imported constant.
- `packages.json` re-hashed for the three primary packages and the dependency closure (commit 59455134).

Out-of-scope diffs in `connection.py` / `connection.yaml` / `test_connection.py` are entirely from the merged #931 (events keyset pagination), not added by this PR.

Approving. Ship before 15:00 UTC.

OjusWiZard (approved):

packages/packages.json
```diff
@@ -9,18 +9,18 @@
...
+        "skill/valory/agent_performance_summary_abci/0.1.0": "bafybeigkhuzwfuhve43gl6tmn7jaxiqaesmpjj726eaywsorzyw7wpxkea",
         "skill/valory/check_stop_trading_abci/0.1.0": "bafybeigkx3zumwsl3sziadjxkyof4yjv45mfvnhccf7sbogyha2rmlf2hq",
-        "skill/valory/chatui_abci/0.1.0": "bafybeiccpd56fdzf2npwqyouoeguyofgebyltgkbaomes6ndrrq3wzgmge",
-        "agent/valory/trader/0.1.0": "bafybeifsssjvvb34yutn63e7typek4vany7mod37eiewui5jywb2ed7rj4",
-        "service/valory/trader_pearl/0.1.0": "bafybeicnyjxstxjzvuaut3cplbeuxh2w7a6hnrm4gnlhwdquoiaox4zgwa",
-        "service/valory/polymarket_trader/0.1.0": "bafybeihbrgey3hr6p5ma2t7ux3ntrhy4ftj5csa6evarwp6gm7okw2afym"
+        "skill/valory/chatui_abci/0.1.0": "bafybeibkpytqj6b5jpvvgmagc6xoigonlmaobh35tqsbrt34te6azlalvi",
+        "agent/valory/trader/0.1.0": "bafybeihdxipbpugvdg5lsuddnyvfjdnzfnr4rg55n4aunnzkao7zwpemua",
+        "service/valory/trader_pearl/0.1.0": "bafybeifkx4fk4ls7dagl6l3mzmsx3bp6auheapsjfuk6cbhja2lah7qdlm",
+        "service/valory/polymarket_trader/0.1.0": "bafybeid6oqxksdddskc7usjbaq6guvblrmm253wlszlgxuognrxuol3sbq"
```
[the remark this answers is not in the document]
LOCKhart07: hashes locked now on purpose

DhairyaPatel7 (approved):

Issue #937, opened 2026-05-03
FSM audit: agent_performance_summary missing dialogues, chatui missing fsm_spec, dead events in decision_maker


DavidMinarsch:
# FSM Audit Report

**Scope:** All FSM skills under `packages/valory/skills/` in this repo:
`agent_performance_summary_abci`, `chatui_abci`, `check_stop_trading_abci`, `decision_maker_abci`, `market_manager_abci`, `staking_abci`, `tx_settlement_multiplexer_abci`, and the `trader_abci` composition.

**Date:** 2026-05-03

**Methodology:** Ran `autonomy analyse {fsm-specs, handlers, dialogues, docstrings}` (after `autonomy packages sync`), then performed manual checklist audit (C1–C4, H1–H3, M1–M5, T1–T6) via three parallel exploration passes covering all `rounds.py` / `behaviours.py` / `payloads.py` / `models.py` and tests, with main-thread verification of suspicious findings.

## Critical Findings

No findings.

(Two C-class candidates raised by sub-agents were verified as **false positives** — see Notes.)

### H3-like / dialogues misconfig: `agent_performance_summary_abci` declares 7 handlers but no `*_dialogues` models
- **File:** `packages/valory/skills/agent_performance_summary_abci/skill.yaml:54-76`
- **Issue:** The `handlers:` block declares `abci`, `contract_api`, `http`, `ipfs`, `ledger_api`, `signing`, `tendermint` handlers. Each of these protocol handlers requires a corresponding `*_dialogues` model in the `models:` block to route messages and maintain dialogue state. The `models:` block here only defines `state`, `params`, and several subgraph models — no dialogue models at all.
- **Tool signal:** `autonomy analyse dialogues` fails with `Common dialogue 'abci_dialogues' is not defined in /.../agent_performance_summary_abci/skill.yaml`.
- **Why it matters:** When this skill is composed into `trader_abci` (which does declare all dialogues at the composed level), it works at runtime — but standalone use, framework introspection (CI dialogue check fails), and any future direct test harness that loads this skill alone will fail to wire dialogues. It is a latent breakage for refactor / extraction.
- **Fix:** Add to the `models:` block:
  ```yaml
    abci_dialogues:
      args: {}
      class_name: AbciDialogues
    contract_api_dialogues:
      args: {}
      class_name: ContractApiDialogues
    http_dialogues:
      args: {}
      class_name: HttpDialogues
    ipfs_dialogues:
      args: {}
[...]

PR #940, opened 2026-05-04
FSM audit fixes (issue #937)

LOCKhart07:
Address findings from the FSM audit in #937, plus one issue the audit missed.

## Summary

- **L3 (proper fix, not workaround):** delete the unreachable `Event.PREPARE_TX → PolymarketSetApprovalRound` transition from `CheckBenchmarkingModeRound`. The round never emits `PREPARE_TX` (no `prepare_tx_event` class attribute, never returned from `end_block`), so the transition was dead. The `# Event.PREPARE_TX` placeholder comment had been added only to placate `autonomy analyse fsm-specs` after the bogus transition was introduced in PR #689; with the transition gone the workaround is no longer needed. Spec entries removed from both `decision_maker_abci/fsm_specification.yaml` and `trader_abci/fsm_specification.yaml`, and the `DecisionMakerAbciApp` docstring updated to match.
- **L1:** `ChatuiLoadBehaviour.async_act` now uses the public `SharedState.chatui_config` property instead of reaching into `_ensure_chatui_store()` / `_chatui_config`. Behavioural contract is preserved (same `_ensure_chatui_store` call ordering, same `ValueError` + message, same logged value); the two `async_act` tests were migrated to exercise the public surface.
- **M2 (dead enums):** drop `NO_SUBSCRIPTION`, `POLYMARKET_FETCH_MARKETS`, `SKIP` from `decision_maker_abci.Event`. Verified unreferenced repo-wide — no `Event.{NAME}` usages, no string-value lookups via the two dynamic `Event(...)` constructors (which only carry a fixed allowlist of bet-placement / redeem events), and no fsm-specification entries. `MarketManagerEvent.POLYMARKET_FETCH_MARKETS` is a separate enum member in `market_manager_abci` and is unaffected.
- **M2 (missing fsm spec):** generated `chatui_abci/fsm_specification.yaml` so the framework's drift-detection gate now covers chatui (was previously silently skipped).
- **H3 (dialogues):** declare the missing `*_dialogues` models in `agent_performance_summary_abci/skill.yaml` (7 added) and `chatui_abci/skill.yaml` (7 added — the audit missed this one because `autonomy analyse dialogues` stops at the first failure, alphabetically). The corresponding `Dialogues` classes already exist in each skill's `dialogues.py`. Both skills are abstract, so runtime behaviour is unchanged; this unblocks the CI dialogue gate and removes a latent breakage for standalone use.
- **Informational (no change):** the `Event.FETCH_ERROR` shared-across-apps warning is intentional — `decision_maker_abci/rounds.py` deliberately imports `MarketManagerEvent.FETCH_ERROR`. Documented in audit notes; left untouched.

## Other `# falsely reporting` workarounds in `decision_maker_abci/rounds.py` — investigated, kept

After the L3 fix worked cleanly I empirically tested whether the remaining ~9 transition-level workarounds and the 2 in-class `# Event.X` static-checker comments could also be removed. They cannot — they patch genuine blind spots in the upstream analyzer. Source: `autonomy/analyse/abci/app_spec.py:442` `check_unreferenced_events`, which is a regex (`Event\.(\w+)`) over `inspect.getsource(base)` for every class in each round's MRO. The function's own docstring explicitly disclaims reachability analysis: "the function simply checks references in the 'raw' source code … Therefore, it does not do any kind of static analysis … nor checks for actual reachability of a return statement returning such events."

The blind spots that justify each surviving workaround:

1. **No override awareness.** When a subclass overrides `none_event` (etc.), the parent class's literal `Event.X` reference is still picked up by the textual scan — so the subclass's transition function must include the parent's event even if the override makes it unreachable. Hits `BetPlacementRound`, `PolymarketBetPlacementRound`, `RedeemRound` (parent `TxPreparationRound` references `Event.NONE`).
2. **No module qualification.** The regex captures `(\w+)` after `Event.` — the qualified enum is thrown away. So `decision_maker_abci.Event` and `market_manager_abci.Event` are conflated by name. Hits `SamplingRound` and `BlacklistingRound`, whose parent `UpdateBetsRound` (in market_manager_abci) references `Event.FETCH_ERROR`; satisfied by adding `MarketManagerEvent.FETCH_ERROR: ImpossibleRound`.
3. **No dynamic construction.** Rounds that emit events via `event = Event(self.most_voted_payload_values[-N])` produce zero regex matches. The bare `# Event.PREPARE_TX` comment in `PolymarketRedeemRound` and `# For static checking / # Event.BET_PLACEMENT_DONE, …` block in `polymarket_bet_placement.py:50` supply the literal symbols the regex needs to find.

Empirical confirmation per case (each one restored before the next test):

| Removed | fsm-specs result |
|---|---|
| `Event.NONE: ImpossibleRound` from `BetPlacementRound` | ❌ `Events {'NONE'} are present in the BetPlacementRound but missing from transition function` |
| `MarketManagerEvent.FETCH_ERROR: ImpossibleRound` from `SamplingRound` | ❌ `Events {'FETCH_ERROR'} are present in the SamplingRound but missing from transition function` |
| `# Event.PREPARE_TX` comment from `PolymarketRedeemRound` | ❌ `Events {'PREPARE_TX'} are listed in the transition function of PolymarketRedeemRound but are never returned from any round's end_block` |

So the L3 case was unique — `Event.PREPARE_TX` wasn't in `CheckBenchmarkingModeRound`'s class hierarchy at all, so removing both the entry and the placeholder comment was clean. The other workarounds patch real upstream limitations and stay.

A proper upstream fix in `valory-xyz/open-autonomy` would (a) walk the MRO leaf-first and stop at the first override of `*_event`, and (b) attach the qualified enum so cross-skill name collisions don't trigger false matches. Out of scope for this PR.

## Test plan

- [x] `autonomy analyse fsm-specs --package` for `decision_maker_abci`, `chatui_abci`, `trader_abci`: all `Check successful`.
- [x] `autonomy analyse dialogues`: passes through both audited skills (now stops on the out-of-scope `funds_manager`, which is non-FSM bookkeeping per the audit notes).
- [x] `autonomy analyse docstrings`: `No update needed.`
- [x] `uv run pytest packages/valory/skills/decision_maker_abci/tests/`: 1095 passed.
- [x] `uv run pytest packages/valory/skills/chatui_abci/tests/`: 138 passed (incl. migrated `async_act` tests; behaviours.py at 100% coverage).
- [x] `uv run pytest packages/valory/skills/agent_performance_summary_abci/tests/`: passes.
- [x] `make ci-linter-checks` (`tomte check-code`): clean.
- [x] `tox -e copyright-check`: clean.
- [x] `autonomy packages lock`: hashes refreshed.


OjusWiZard (approved):

DavidMinarsch:
Please open an issue for the discussion at the end to fix the analyser on open-autonomy

DavidMinarsch (approved):

bennyjo (approved):
Fixes match the #937 audit scope; spot-checked the load-bearing claims (CheckBenchmarkingModeRound never emits PREPARE_TX; the two dynamic `Event(...)` constructors only ever receive values from a fixed allowlist that excludes the three removed members; new chatui fsm_specification.yaml matches the in-code transitions; dialogue model additions match existing classes in dialogues.py). The empirical "investigated, kept" table for the surviving workarounds is great. LGTM.

PR #941, opened 2026-05-04
fix(polymarket): switch neg-risk redeem to CLOB v2 4-arg overload

LOCKhart07:
## Summary

Three commits on this branch, addressing the neg-risk GS013 dead-loop on both the behaviour path and the builder-program path.

### Commit 1 (`773091f`) — neg-risk behaviour path
Switches the neg-risk branch in `_prepare_redeem_tx` from the v1 2-arg `redeemPositions(bytes32, uint256[])` overload (selector `0xdbeccb23`) to the v2 4-arg `redeemPositions(IERC20, bytes32, bytes32, uint256[])` overload (selector `0x01b7037c`). Drops the on-chain balance fetch and `asset`-field guard, since the adapter now discovers Safe ERC1155 balances itself. Removes the dead `_build_redeem_neg_risk_data` helper.

### Commit 2 (`bfaefe7`) — held-only index_set, test tightening
The previous commit unified both branches behind `indexSets=[1, 2]` (both bitmasks). Reverted to `indexSets=[1 << outcomeIndex]` (single held bitmask): redeeming the losing side would still pay 0 but burns extra gas on balance/payout reads inside the CTF. Net effect: standard-path behaviour is now identical to before this PR (the original review point about an undocumented standard-path change is moot — the change is gone). Tests now assert the held bitmask is present, the losing-side bitmask is absent, the target adapter is correct, and conditionId / collateral address appear in the encoded args.

### Commit 3 (`a8ec5ce`) — builder-program path + dead-code cleanup
The builder-program path (`polymarket_client/_redeem_positions`) had the same bug on a parallel code path: when `polymarket_builder_program_enabled=true`, redemption flows through `_redeem_via_builder` → connection request, which still hardcoded the broken 2-arg selector for neg-risk markets. Switched both branches there to the 4-arg overload — calldata is now byte-identical for standard and neg-risk; only the target adapter address differs. Dead code removed: `_get_token_balance_from_chain`, `_get_token_balance`, `_conditional_tokens_interact`, the `user_token_balance` property, and the `size` parameter plumbed through `_redeem_position` → connection request payload (the adapter discovers balances itself, so `size` had no production use after the fix).

## Why

Polymarket's CLOB v2 cutover (Quantstamp/Cantina audits, March 2026) changed how `negativeRisk: true` markets resolve. They're now plain CTF binary conditions, **not** registered with the `NegRiskAdapter`:

```
NegRiskAdapter.getDetermined(conditionId)     → false
NegRiskAdapter.outcomeSlotCount(conditionId)  → reverts
ConditionalTokens.outcomeSlotCount(cid)       → 2
ConditionalTokens.payoutDenominator(cid)      → 1   (resolved)
```

The 2-arg overload introduced in [#795](https://github.com/valory-xyz/trader/pull/795) reverts because the adapter looks up the conditionId in the NegRisk system and finds nothing. On Pearl this manifested as `tx_settlement` dead-looping forever:

1. Safe `execTransaction` → revert with `GS013`
2. Agent re-submits with `gas: 0` (because `eth_estimateGas` failed)
3. RPC bounces it as intrinsic gas too low
4. Keeper rotates → loop forever

No tx ever lands → no Mech activity → 0 staking rewards.

Reproduced on at least two Safes:
- `0xB45815512954e57eA524f0F343f5475DD6Bd30cD` (~\$45 stuck) — Safe nonce 2015
- `0x21CFdb1e8dEe44B4f38e273B9b5B7605043FFcFB` (~\$2.98 stuck) — Safe nonce 936

The `NegRiskCtfCollateralAdapter` itself still works for these positions, but only via its 4-arg overload. Verified working: Tenderly sim `36ad5013-5966-4744-9016-d8ff9a14c824` for cid `0x37d2…6f82`.

## Live verification (behaviour path)

Ran the trader locally against a Pearl service with `polymarket_builder_program_enabled=false` and a previously-stuck neg-risk position (cid `0x3e1275fdd4e51c25f66c8f234f3e0cae700dc24c740077a1053372c3ddd2defe`). Observed in `logs/agent_log_latest.log`:

- Both Yes (size 1.5384) and No (size 2.7777) sides prepared as separate batches inside one multisend.
- Encoded calldata starts with selector `01b7037c` (the 4-arg overload).
- Target address is `0xada2005600dec949baf300f4c6120000bdb6eaab` (NegRiskCtfCollateralAdapter).
- Collateral token in the encoded args is `0xc011a7e12a19f7b1f670d46f03b03f3342e82dfb` (pUSD).
- `finalization_round` exits with `Event.DONE` after one pass.
- > `The transaction submitted by polymarket_redeem_round was successfully settled.`
- Zero occurrences of `GS013` in the entire run (vs. the previous infinite-revert loop). Zero occurrences of the old `dbeccb23` selector.

The builder-program path was not exercised in this run; flagged below as still pending live verification.

## Test plan

- [x] Unit tests updated (TDD; failing-then-passing) on both behaviour and connection paths; 100% coverage on both modified modules
- [x] `tomte check-code` clean (black, isort, flake8, mypy, pylint, darglint)
- [x] `autonomy packages lock` re-run; hashes propagated
- [x] Behaviour path: run the trader against a Pearl service with a stuck redeem; confirm new calldata, Safe tx settles, FSM advances past `finalization_round` (see Live verification above)
- [ ] Builder path: same end-to-end confirmation with `polymarket_builder_program_enabled=true`
- [x] End-to-end confirm pUSD balance increases on the Safe by the expected amount and Mech activity resumes for staking-epoch credit

## References

- [PR #795](https://github.com/valory-xyz/trader/pull/795) — the v1 NegRisk fix that produced the now-stuck calldata
- [Polymarket redeem docs](https://docs.polymarket.com/trading/ctf/redeem)
- [Polymarket contracts](https://docs.polymarket.com/resources/contracts)


packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
```diff
@@ -421,49 +420,21 @@ def _prepare_redeem_tx(
...
+            # Both adapters expose the same 4-arg redeemPositions(
+            # IERC20, bytes32, bytes32, uint256[]) signature; only the
+            # destination contract differs. The adapter discovers the Safe's
+            # ERC1155 balances itself, so passing indexSets=[1, 2] redeems
+            # both sides — the losing side pays nothing.
+            redeem_data = self._build_redeem_positions_data(
+                collateral_token=self.params.polymarket_collateral_address,
+                condition_id=condition_id,
+                index_sets=[1, 2],
+            )
```
OjusWiZard, line 339: **Behavior change in the standard (non-neg-risk) path is undocumented.** Previously this branch built calldata with `index_sets=[outcome_index + 1]` (only the held side); now both branches share `index_sets=[1, 2]`. The PR description frames this as a neg-risk-only fix.

It's almost certainly safe — for a determined binary condition, redeeming the losing index set just pays 0 — but worth either calling out in the PR body or asserting in `test_standard_position_targets_ctf_collateral_adapter` that the calldata now contains both index sets, so a future reader doesn't assume the standard path is unchanged.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
```diff
@@ -1009,61 +954,17 @@ def mock_build_multisend_safe_tx_hash() -> None:  # type: ignore[no-untyped-def,
...
+        calldata = behaviour.multisend_batches[0].data.hex()
+        assert calldata.startswith("01b7037c")
+        index_set_yes = (
+            "0000000000000000000000000000000000000000000000000000000000000001"
+        )
+        index_set_no = (
+            "0000000000000000000000000000000000000000000000000000000000000002"
+        )
+        assert index_set_yes in calldata
+        assert index_set_no in calldata
```
OjusWiZard: **Missing target-address assertion.** This test verifies the selector and both indexSets in the calldata but never asserts `behaviour.multisend_batches[0].to == polymarket_neg_risk_ctf_collateral_adapter_address`.

The whole point of the production bug being fixed is calldata routed to the right adapter — an accidental fallthrough to the standard adapter would still satisfy the current assertions. Suggest:

```python
assert (
    behaviour.multisend_batches[0].to
    == "0x7234567890123456789012345678901234567890"
)
```

OjusWiZard (commented):
Inline review notes from /ecc:review-pr — see comments below. No critical issues; three important items worth addressing before merge.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
```diff
@@ -421,49 +420,21 @@ def _prepare_redeem_tx(
...
             if is_neg_risk:
```
bennyjo, line 340: `polymarket_client/connection.py:1040-1067` still hardcodes selector `0xdbeccb23` and the 2-arg `redeemPositions(bytes32, uint256[])` ABI encoding for the neg-risk branch — the exact pattern this PR diagnoses as broken under the CLOB v2 cutover. The connection-side tests at `polymarket_client/tests/test_connection.py:1678-1698` pin that broken selector. When `polymarket_builder_program_enabled=true`, redemption flows through there (via `_redeem_via_builder` → `send_polymarket_connection_request`) instead of `_prepare_redeem_tx`, so the same revert / dead-loop reappears for any service running with the builder program on. Is the builder path deliberately scoped out as a follow-up — happy to leave as-is if so, but it's worth a sentence in the PR body and a tracking issue — or did this branch get missed?

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
bennyjo: Would you consider tightening this test? It pins the selector and both indexSet bytes, but doesn't pin the `conditionId` or the collateral-token address inside the calldata. A regression that broke the conditionId offset (e.g., zero-stuffed it, or shifted the array offset for a 2-arg vs. 4-arg ABI) would still pass. Adding `assert "aabbccdd" in calldata` and `assert "1234567890123456789012345678901234567890" in calldata` would close the gap — same fix would help `test_standard_position_builds_batch` (line 845-896), which has the same shape.

bennyjo (commented):
Two comments — one scope question on the builder-program path, one test-discrimination suggestion. See inline.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
LOCKhart07: Reverted in `bfaefe7f` — rather than documenting the standard-path change, removed it. Both branches now use `indexSets=[1 << outcomeIndex]` (single held bitmask), which is what the standard branch was already doing pre-PR. Redeeming the losing side would just burn gas on zero-payout reads inside the CTF.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
LOCKhart07: Added in `bfaefe7f`: `assert behaviour.multisend_batches[0].to == "0x7234..."` (the neg-risk adapter), plus `index_set_other not in calldata` and explicit `conditionId` / collateral-address assertions on both standard and neg-risk tests.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
LOCKhart07: Good catch — fixed in `a8ec5ce8`. Both branches in `polymarket_client/_redeem_positions` now use the 4-arg `0x01b7037c` overload; calldata is byte-identical for standard and neg-risk and only the target adapter address differs. Tests at `test_connection.py` updated accordingly, including a new regression test that asserts both branches produce identical calldata. While there I removed the now-unused `size` parameter and the upstream balance-fetch block in `_redeem_via_builder` — and the orphaned `_get_token_balance_from_chain` / `_get_token_balance` / `_conditional_tokens_interact` helpers and `user_token_balance` property that fell out with it.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
LOCKhart07: Tightened in `bfaefe7f`: both `test_neg_risk_position_builds_batch` and `test_standard_position_builds_batch` now assert `"aabbccdd" in calldata` (conditionId) and `"1234567890123456789012345678901234567890" in calldata` (collateral). Also asserting that the losing-side index_set is NOT present, which catches a regression to `[1, 2]`. Connection tests got the equivalent treatment in `a8ec5ce8` via ABI-decoded args.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
```diff
@@ -421,49 +310,21 @@ def _prepare_redeem_tx(
...
-            # Build the redemption data based on market type
+            # Both adapters expose the same 4-arg redeemPositions(
+            # IERC20, bytes32, bytes32, uint256[]) signature; only the
+            # destination contract differs. Redeem only the held outcome's
+            # index set (1 << outcomeIndex) — including the losing side
+            # would still pay 0 but burn extra gas on balance/payout reads.
+            redeem_data = self._build_redeem_positions_data(
+                collateral_token=self.params.polymarket_collateral_address,
+                condition_id=condition_id,
+                index_sets=[1 << outcome_index],
```
OjusWiZard, line 338: **[Important] Removed `asset` / zero-balance guards — silent no-op redeems.**

The previous code logged `"Missing asset (token ID) for position {condition_id}"` and `continue`d on malformed positions, and skipped neg-risk positions when on-chain balance was 0. Both guards are gone here and in `_redeem_via_builder`.

Adapter-side balance discovery is fine for correctness, but the agent will now batch and submit redeem calls for positions the Safe holds 0 of: gas spent, no payout, no log explaining why. A malformed position dict (`outcomeIndex` missing → `1 << None` raises `TypeError` mid-loop and aborts the whole batch) loses the original guard's specific error message.

Suggest: defensive validation log + `continue` on missing required fields, plus an info-level "submitting without pre-flight balance check" log so operators have a signal when an empty redeem ships.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
```diff
@@ -516,41 +377,3 @@ def _build_redeem_positions_data(
         array_elements = "".join([hex(idx)[2:].zfill(64) for idx in index_sets])
```
OjusWiZard, line 394: **[Advisory] Hand-rolled ABI encoding.**

`_build_redeem_positions_data` zfills + concats hex strings to encode the 4-arg payload. The connection-side production code uses `eth_abi.encode` and the matching test uses `eth_abi.decode`. Switching here to `eth_abi.encode(["address","bytes32","bytes32","uint256[]"], [...])` would eliminate a class of typo / padding bugs and converge the two code paths on a single encoder.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
```diff
@@ -953,12 +663,33 @@ def mock_build_multisend_safe_tx_hash() -> None:  # type: ignore[no-untyped-def,
 
         assert result == "0xfinalHash"
         assert len(behaviour.multisend_batches) == 1
-
-    def test_neg_risk_position_builds_batch_with_balance(self) -> None:
-        """Should build multisend batch for neg risk positions using on-chain balance."""
+        # 4-arg redeemPositions selector + indexSets=[1] (held outcome only).
```
OjusWiZard: **[Important] Tests don't assert the old `0xdbeccb23` selector is absent.**

Assertions confirm `0x01b7037c` is present in produced calldata, but nothing asserts the v1 2-arg selector is absent across either path. A regression that re-introduces the v1 overload alongside v2 (e.g. via a stray helper resurrected from git history) would not fail any test.

Add a one-line negative assertion on each path: `assert "dbeccb23" not in calldata.hex()`.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
```diff
@@ -953,12 +663,33 @@ def mock_build_multisend_safe_tx_hash() -> None:  # type: ignore[no-untyped-def,
...
+        v1 2-arg overload reverts with GS013 for v2-resolved markets flagged
+        negativeRisk. The agent passes only the held outcome's bitmask
+        (1 << outcomeIndex) to avoid wasting gas on the zero-payout side.
+        """
         behaviour = _make_behaviour()
         behaviour.multisend_batches = []
-        behaviour._get_token_balance_from_chain = lambda token_id: _return_gen(500)  # type: ignore[method-assign]
 
         def mock_build_multisend_data() -> None:  # type: ignore[no-untyped-def, misc]
             """Mock build multisend data."""
```
OjusWiZard, line 842: **[Important] Substring-based calldata assertions are weak.**

Substring `in` checks against 64-char zero-padded hex (`0...01`, `0...02`) will coincidentally match ABI offsets, lengths, and padding of any zero-heavy payload — the assertion can pass on incorrectly-encoded calldata. The connection-side test (`test_neg_risk_calldata_encodes_4arg_payload_correctly`) already uses `eth_abi.decode` to assert collateral / conditionId / indexSets at typed-argument level. Adopt the same pattern here so the behaviour-side test actually proves the encoded `collateral` argument equals `polymarket_collateral_address` rather than appearing somewhere in the bytes.

OjusWiZard (commented):
Multi-perspective review (code-reviewer + silent-failure-hunter + pr-test-analyzer). Findings posted inline. Note: the previously reported [1,2] vs [1<<outcome_index] divergence between behaviour and connection paths was a stale-checkout false positive — verified at PR HEAD that both paths use `[1 << outcome_index]`.

mariapiamo (commented):
Question:
This PR’s main bug is in the neg-risk path, but the latest version also changes the standard CtfCollateralAdapter path from the previous per-position `indexSets=[outcome_index + 1]` shape to the shared `indexSets=[1 << outcomeIndex]` / “held outcome only” framing and updates tests around that assumption. For binary outcomes the encoded value is equivalent (`0 -> 1`, `1 -> 2`), so this may be only a cleanup. But the deployed CtfCollateralAdapter ignores the external `uint256[]` parameter and internally redeems `CTFHelpers.partition()` anyway, so the new comment/test claim that omitting the losing side avoids adapter balance/payout work does not match the on-chain source.

Can we clarify why the standard CtfCollateralAdapter path needed to be touched at all if it was already working, and either keep that path unchanged or adjust the comment/tests to avoid implying `indexSets` controls adapter-side redemption work?

Findings:

[P3] Do not claim indexSets limits adapter redeem work
The latest change encodes only the held bitmask, but the deployed CtfCollateralAdapter does not use the external uint256[] parameter: the ABI leaves it unnamed and the source calls the internal redeem hook with CTFHelpers.partition(). NegRiskCtfCollateralAdapter inherits that external function and its internal override also ignores indexSets, reading current balances instead. So [1] versus [1,2] may reduce calldata bytes, but it does not prevent losing-side balance reads/payout logic; this comment and the matching tests encode a false on-chain assumption.

LOCKhart07:
@mariapiamo good catch — verified on the deployed Polymarket source ([`Polymarket/ctf-exchange-v2`](https://github.com/Polymarket/ctf-exchange-v2/blob/main/src/adapters/CtfCollateralAdapter.sol#L100-L122)). Both adapters drop the `uint256[]` parameter:

```solidity
// CtfCollateralAdapter.redeemPositions
function redeemPositions(address, bytes32, bytes32 _conditionId, uint256[] calldata) external onlyUnpaused(USDCE) {
    uint256[] memory positionIds = _getPositionIds(_conditionId);
    uint256[] memory amounts = new uint256[](2);
    amounts[0] = CONDITIONAL_TOKENS.balanceOf(msg.sender, positionIds[0]);
    amounts[1] = CONDITIONAL_TOKENS.balanceOf(msg.sender, positionIds[1]);
    ...
    _redeemPositions(_conditionId, CTFHelpers.partition());
    ...
}
```

`NegRiskCtfCollateralAdapter` inherits this and overrides `_redeemPositions` to read `balanceOf` on both sides and call the legacy `NegRiskAdapter.redeemPositions` with its own `amounts` array — the caller-supplied indexSets is ignored there too.

So the held-only bitmask reduces calldata size only; it does not limit on-chain redemption work, and the standard-path notation refresh from `outcome_index + 1` to `1 << outcomeIndex` was cosmetic (equivalent for binary).

Fix in [`d29a0a70`](https://github.com/valory-xyz/trader/commit/d29a0a70): rewrote the comments in `_prepare_redeem_tx` / `_redeem_position` and the test docstrings to drop the false "saves gas on balance/payout reads" framing. Pinning the encoded shape is kept as a wire-format regression check, but the rationale now matches what the adapter actually does.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
LOCKhart07: Addressed the TypeError-abort half in [`35bb90d1`](https://github.com/valory-xyz/trader/commit/35bb90d1). Both `_prepare_redeem_tx` and `_redeem_via_builder` now guard the top of their position loops:

```python
if condition_id is None or outcome_index is None:
    self.context.logger.error(
        f"Skipping malformed position (missing conditionId or "
        f"outcomeIndex): {position}"
    )
    continue
```

Plus a regression test per loop that mixes a malformed position (no `outcomeIndex`) with a valid one and asserts only the valid one progresses. One bad dict no longer kills the whole multisend.

On the zero-balance log: opted not to restore the on-chain `balanceOf` fetch — it would mean an extra RPC round-trip per position just for a log signal, and the original reason for the guard (the broken 2-arg ABI requiring exact `redeem_amounts`) is gone now that the adapter discovers balances itself. The Polymarket API's `redeemable=true` filter is the source of truth; if the API surfaces it, the adapter pays out correctly (or burns worthless tokens for 0). The existing `Preparing redeem tx for ...` info log gives operators a per-redeem signal without RPC overhead. Happy to revisit if you have a concrete scenario where the API surfaces a stale-cache redeemable position in practice.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
LOCKhart07: Done in [`f64db28f`](https://github.com/valory-xyz/trader/commit/f64db28f). Added `assert "dbeccb23" not in calldata` (and `not in tx.data` on the connection side) to all four selector-inspection tests:

- `test_polymarket_reedem.py::test_standard_position_builds_batch`
- `test_polymarket_reedem.py::test_neg_risk_position_builds_batch`
- `test_connection.py::test_standard_market_calldata_uses_ctf_selector`
- `test_connection.py::test_neg_risk_calldata_uses_4arg_selector`

A stray resurrection of the v1 overload would now fail at least one test on each path.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
LOCKhart07: Done in [`ae01c1af`](https://github.com/valory-xyz/trader/commit/ae01c1af). Both `test_standard_position_builds_batch` and `test_neg_risk_position_builds_batch` now strip the 4-byte selector and run the remainder through `eth_abi.decode` with the expected schema:

```python
collateral, parent, condition_id, index_sets = abi_decode(
    ["address", "bytes32", "bytes32", "uint256[]"],
    bytes.fromhex(calldata[8:]),
)
assert collateral.lower() == "0x1234..."
assert parent == bytes(32)
assert condition_id.hex() == "00...aabbccdd"
assert list(index_sets) == [1]
```

This matches the pattern already used in `test_connection.py::test_neg_risk_calldata_encodes_4arg_payload_correctly` and replaces the substring `in` checks (`"aabbccdd" in calldata`, `index_set_held in calldata`, etc.) with typed-argument assertions, so a coincidental ABI-offset / length / padding match can no longer make the test green on a malformed payload.

Side note: this also cross-validates the hand-rolled encoder in `_build_redeem_positions_data` against `eth_abi`'s reference decoder — a typo in the zfill / concat would now surface as a decode mismatch.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
LOCKhart07: Agreed — the hand-rolled encoder is a correctness liability and converging on `eth_abi.encode` is the right move. Holding it back from this PR to keep the diff focused on the bug fix and test tightening; the encoder itself has been working in production and the new ABI-decode assertions on the behaviour-side tests (in [`ae01c1af`](https://github.com/valory-xyz/trader/commit/ae01c1af)) now round-trip the output through `eth_abi`'s reference decoder, so any encoder typo would surface immediately.

Happy to open a follow-up issue to track the swap if you'd like — let me know.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_reedem.py
```diff
@@ -792,71 +587,52 @@ def mock_redeem(  # type: ignore[no-untyped-def]
...
                 pass
 
-        assert len(redeem_calls) == 1
-        # Size should be the on-chain balance, not the API size
-        assert redeem_calls[0]["size"] == 999
+        assert redeem_calls == [
+            {"condition_id": "0xcond1", "outcome_index": 1, "is_neg_risk": True}
+        ]
+
+    def test_malformed_position_skipped_others_processed(self) -> None:
```
OjusWiZard, line 594: **[Important] Branch-coverage gap on the `or` clause.**

The guard is `if condition_id is None or outcome_index is None`, but this test (and its sibling at line 682) only omits `outcomeIndex`. The `condition_id is None` branch is never exercised. Line coverage is satisfied via short-circuit, but a future swap from `or` to `and` would silently pass tests.

Fix: add a second case (or `pytest.parametrize`) with `conditionId=None, outcomeIndex=0` so both legs of the `or` are covered.

packages/valory/connections/polymarket_client/tests/test_connection.py
```diff
@@ -1786,18 +1795,54 @@ def test_neg_risk_correct_redeem_amounts_for_outcome_1(self) -> None:
...
-            size=50.0,
         )
 
         tx = conn.relayer_client.execute.call_args[1]["transactions"][0]
-        # Skip "0x" and 4-byte selector, then ABI-decode the remaining args
         calldata_bytes = bytes.fromhex(tx.data[2:])
         args_bytes = calldata_bytes[4:]  # skip 4-byte selector
-        _condition_id, redeem_amounts = abi_decode(["bytes32", "uint256[]"], args_bytes)
-        assert list(redeem_amounts) == [0, 50]
+        collateral, _parent, condition_id, index_sets = abi_decode(
```
OjusWiZard, line 1806: **[Advisory] `_parent` decoded but unasserted.**

Behaviour-side tests assert `parent == bytes(32)`; the connection test discards `_parent`. One-line tightening: `assert _parent == b"\x00" * 32`.

packages/valory/connections/polymarket_client/tests/test_connection.py
```diff
@@ -1673,12 +1670,19 @@ def test_standard_market_calldata_uses_ctf_selector(self) -> None:
...
+        assert "dbeccb23" not in tx.data
+
+    def test_neg_risk_calldata_uses_4arg_selector(self) -> None:
+        """Neg-risk market uses the same 4-arg selector 01b7037c as standard markets.
+
+        Why: under CLOB v2, NegRiskCtfCollateralAdapter accepts the standard
+        redeemPositions(address,bytes32,bytes32,uint256[]) overload. The v1
+        2-arg overload (0xdbeccb23) reverts with GS013 for v2-resolved markets
+        flagged negativeRisk because they're plain CTF binary conditions, not
+        registered with the NegRiskAdapter.
```
OjusWiZard, line 1685: **[Advisory] Standard-branch ABI decode is implicit only.**

`test_neg_risk_calldata_encodes_4arg_payload_correctly` decodes all 4 args, but the standard equivalent only checks the selector. The standard branch's full payload is currently covered transitively via `test_neg_risk_and_standard_produce_identical_calldata`. Making it explicit (one extra `abi_decode` block here) would localize failures and survive removal of the equivalence test.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py
```diff
@@ -416,54 +314,34 @@ def _prepare_redeem_tx(
             size = position.get("size", 0)
             is_neg_risk = position.get("negativeRisk", False)
 
+            if condition_id is None or outcome_index is None:
+                self.context.logger.error(
+                    f"Skipping malformed position (missing conditionId or "
+                    f"outcomeIndex): {position}"
+                )
+                continue
```
OjusWiZard, line 322: **[Advisory] Minor DRY.** This 6-line guard is near-identical to the one in `_redeem_via_builder` (line 269). Two call sites is below the extract-helper threshold, so YAGNI applies — flagging only in case a third caller appears.

OjusWiZard (approved):
Re-review (code-reviewer + pr-test-analyzer) at HEAD a715e905. The three new commits cleanly resolve the prior Critical/Important findings: malformed-position guards added on both call sites, `dbeccb23` v1-selector absence asserted on both connection and behaviour sides, and behaviour-side calldata assertions switched to `eth_abi.decode`. Approving — only one minor branch-coverage tightening worth doing before merge (see inline). Other comments are advisory.

bennyjo (approved):
Re-review: both prior items (builder-program path scope, test-discrimination on calldata args) cleanly resolved across `a8ec5ce8`, `bfaefe7f`, `ae01c1af`, and `f64db28f`. Builder path now uses the 4-arg overload with byte-identical calldata to standard, tests use `eth_abi.decode` for typed assertions, and `dbeccb23` absence is pinned on all four selector tests. Approving.

LOCKhart07:
Confirmed on Slack
@mariapiamo 
> I checked already, I do not see big issues

PR #942, opened 2026-05-06
feat(withdrawal): operator-controlled Polymarket sell-off

LOCKhart07:
Implements an operator-controlled withdrawal mode for the Polystrat (Polymarket) service. Operators arm the agent via an HTTP POST; on the next cycle, the agent diverts from normal trading into a one-shot sell-off of every unredeemable position, then keeps cycling in withdrawal mode (no betting) until the process is restarted. Omenstrat is intentionally out of scope (POST returns 501; behaviour is a defensive halting stub).

The branch was developed iteratively: phases 1–3 set up the HTTP + persistence + FSM + behaviour scaffolding; the remaining commits are bug fixes from empirical testing on Polygon mainnet (live partial fills, no-liquidity FAK kills, async-matched FAKs, etc.) refining the system to match observed CLOB behaviour and operator semantics. PR #943 (now merged into this branch) added a post-sweep snapshot of `funds_locked_in_markets` so the FE can render an accurate locked-positions card after a partial sweep. See the commits tab for individual diffs.

## API contract

Operator-facing HTTP contract for the withdrawal sweep. Implementation:
`packages/valory/skills/chatui_abci/handlers.py` for the endpoints,
`packages/valory/skills/decision_maker_abci/behaviours/polymarket_withdraw.py`
for the sweep behaviour.

### Endpoints

Base: `http://<agent-host>:8716`

| Method | Path | Description |
|---|---|---|
| `POST` | `/api/v1/withdrawal` | Arm withdrawal mode |
| `GET`  | `/api/v1/withdrawal` | Read current status |

Polystrat-only. On Omenstrat both return `501 Not Implemented`.

### State machine

`withdrawal_state` is a single string field persisted to disk in `${store_path}/chatui_param_store.json`:

| Value | Meaning | Set by |
|---|---|---|
| `idle` | Not in withdrawal mode | initial / boot-time auto-clear |
| `armed` | Operator armed, sweep hasn't started yet | `POST` handler (idle → armed) |
| `selling` | Sweep in progress | behaviour at sweep entry |
| `complete` | Last sweep finished without errors | behaviour at sweep end |
| `errored` | Last sweep finished with at least one error | behaviour at sweep end |

Once armed (`withdrawal_mode = True`), the agent **stays in withdrawal mode for the lifetime of the process** — every cycle re-runs the sweep, no normal trading. State cycles between `selling → complete | errored` on each pass.

[...]

PR #943, opened 2026-05-07
feat(withdrawal): refresh funds_locked_in_markets at sweep end

LOCKhart07:
## Summary

- Adds a cache-busting hook in `polymarket_withdraw._finish()` that snapshots current locked-funds USD value into `agent_performance.json` at every withdrawal terminal state.
- Adds `update_funds_locked_in_markets(value)` on `agent_performance_summary_abci.SharedState`, mirroring the existing `update_agent_behavior` helper. Lazily initializes nested dataclasses when the file doesn't yet exist.
- Bridges the latency gap between sweep completion and the next normal `agent_performance_summary_abci` round (which fetches via subgraph and can take minutes).

Stacked on top of #942 (`withdrawal-mode-phase1`); merge that first.

## Why

After a withdrawal sweep, the cached `funds_locked_in_markets` value is stale because the sweep changed on-chain positions. The FE's `/api/v1/agent/performance` poll would show the pre-sweep total until the next subgraph-driven refresh round runs.

Iason and Rajat's Slack thread (2026-05-07) asked for the FE to reflect the latest locked-funds value when the partial-withdrawal "Some positions couldn't be sold" toast appears. The fix is the agent-side bridge.

## Approach

`_finish()` is the only convergence point in `async_act` — every terminal exit (top-level fetch failure, no-sellable positions, post-sweep complete/errored) goes through it. One insertion catches all five paths.

The hook is best-effort: if the fetch fails (transient API issue) or the compute / write fails (malformed position record, disk IO error), it logs a warning and skips. The performance summary keeps its previous value, refreshed on the next normal round. The sweep terminal state is still emitted via the regular `finish_behaviour` payload — the snapshot failure does not stall the FSM.

`self.context.state` resolves to the composed shared state, which inherits from `agent_performance_summary_abci.SharedState` via the chain `chatui_abci.SharedState -> agent_performance_summary_abci.SharedState`, so the helper is reachable without any new cross-skill imports.

The locked value is computed as **cost basis** to match the formula used by the normal performance round at `agent_performance_summary_abci/behaviours.py:762` (`(total_traded - total_traded_settled) / token_divisor`). Concretely the snapshot sums `initialValue` (= `size × avgPrice`, the cost basis of currently held shares) across unredeemable positions and rounds to 2dp. Using cost basis (not mark-to-market) prevents a non-monotonic jump on the FE when the next normal round overwrites this value. The next normal round still has the final word; both paths now coexist in the same units, with the latest write winning.

## Test plan

- [x] 4 SharedState helper tests (`test_models.py`):
  - `test_update_funds_locked_in_markets_creates_summary_if_missing`
  - `test_update_funds_locked_in_markets_preserves_other_fields`
  - `test_update_funds_locked_in_markets_updates_timestamp`
  - `test_update_funds_locked_in_markets_handles_existing_metrics_none`
- [x] 7 behaviour tests (`test_withdrawal_rounds.py`):
  - `test_finish_snapshots_locked_funds_on_clean_complete`
  - `test_finish_snapshot_excludes_redeemable_positions`
  - `test_finish_snapshot_handles_missing_initial_value`
  - `test_finish_snapshot_handles_unparseable_value` (new — falsifies the broad-except guard)
  - `test_finish_snapshot_zero_positions_records_zero`
  - `test_finish_snapshot_on_fetch_failure_logs_and_skips`
  - `test_finish_snapshot_on_initial_fetch_failure_path`
- [x] All affected test suites green
- [x] tomte check-code clean
- [x] autonomy packages lock
- [ ] Manual verification on dev deployment: arm withdrawal, observe sweep, confirm `/api/v1/agent/performance` reports fresh `funds_locked_in_markets` within the FE polling interval

## Review feedback addressed (2026-05-07)

- bennyjo flagged that the broad-except guard only covered the fetch failure — an unparseable `initialValue` / `size` could escape `_finish` and stall the FSM. Wrapped compute + helper call in a broad `try/except` mirroring the fetch-failure branch. Added a falsifying test.
- bennyjo flagged that the previous mark-to-market formula (`size × curPrice`) would write a value the next normal round would overwrite to a different number, causing a non-monotonic FE jump. Switched to cost basis (`initialValue`) + 2dp rounding to match the normal round.

## FE-side notes

No FE changes required beyond the existing `/api/v1/agent/performance` poll Rajat already wires after withdrawal terminal state. The field is `metrics.funds_locked_in_markets` in the existing endpoint shape.


packages/valory/skills/decision_maker_abci/behaviours/polymarket_withdraw.py
```diff
@@ -695,6 +695,47 @@ def _record_top_level_error(self, op_name: str, reason: str) -> None:
...
+            return
+        locked = sum(
+            float(p.get("size") or 0) * float(p.get("curPrice") or 0)
+            for p in positions
+            if not p.get("redeemable", False)
+        )
+        self.context.logger.info(
+            f"withdrawal: snapshotting funds_locked_in_markets={locked}"
+        )
+        self.context.state.update_funds_locked_in_markets(locked)
```
bennyjo: The PR description states this hook is "best-effort: a fetch failure logs a warning and skips ... performance summary keeps its previous value." The current guard at line 727 only covers fetch failures. If the response parses but a position has a non-numeric `size` / `curPrice` (`float("N/A")`), the `ValueError` escapes `_snapshot_locked_funds` → `_finish` → blocks `finish_behaviour(payload)` from emitting and stalls the FSM. Same risk if `update_funds_locked_in_markets` raises mid-write (disk IO error, JSON encode failure on a corrupt existing file).

Suggest wrapping the compute + helper call in a broad `try/except` mirroring the fetch-failure branch:

```suggestion
        try:
            locked = sum(
                float(p.get("size") or 0) * float(p.get("curPrice") or 0)
                for p in positions
                if not p.get("redeemable", False)
            )
            self.context.logger.info(
                f"withdrawal: snapshotting funds_locked_in_markets={locked}"
            )
            self.context.state.update_funds_locked_in_markets(locked)
        except Exception as e:  # noqa: BLE001 — best-effort by design
            self.context.logger.warning(
                f"withdrawal: skipping locked-funds snapshot "
                f"(compute/write failed: {e!r})"
            )
```

Worth a falsifying test too — `test_finish_snapshot_handles_missing_curprice` covers `None` / missing keys (which the `or 0` short-circuit handles) but not unparseable strings.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_withdraw.py
```diff
@@ -695,6 +695,47 @@ def _record_top_level_error(self, op_name: str, reason: str) -> None:
...
+        :yield: framework yields between dispatch and response.
+        """
+        positions, error = yield from self._request_fetch_positions()
+        if error is not None or positions is None:
+            self.context.logger.warning(
+                f"withdrawal: skipping locked-funds snapshot "
+                f"(fetch failed: {error})"
+            )
+            return
+        locked = sum(
```
bennyjo: Is the unit/semantic shift here intentional?

This computes `funds_locked_in_markets` as **mark-to-market** — current CLOB price × size for unredeemable positions. The normal-round formula at `agent_performance_summary_abci/behaviours.py:762` is `(total_traded - total_traded_settled) / token_divisor`, i.e. **cost basis** (USDC originally spent on unsettled positions). These produce different numbers: a position bought at 0.40 now trading at 0.65 will write `0.65 × size` here, then the next normal round reverts the field to `0.40 × size`. So even on a successful sweep, the FE figure will visibly jump back when the next subgraph round runs.

The PR body notes "the latest write winning" but doesn't surface that the units differ. Happy to leave as-is if deliberate — flagging because the operator will see a non-monotonic value with no obvious cause. (And the normal round rounds to 2dp — `round(value, 2)` — while the snapshot writes the raw float. Worth matching at minimum.)

bennyjo (commented):
Two questions — error-handling scope and unit/semantic alignment with the normal performance round.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_withdraw.py
LOCKhart07: Fixed in 7e5158ed — wrapped the compute + helper call in a broad `try/except` mirroring the fetch-failure branch. An unparseable `size` / `initialValue` (or any disk-IO error inside `update_funds_locked_in_markets`) now logs the warning and skips, instead of escaping `_finish` and blocking `finish_behaviour(payload)`.

Added `test_finish_snapshot_handles_unparseable_value` (`initialValue: "N/A"`) as the falsifying test — it asserts both that `update_funds_locked_in_markets` is not called *and* that the consensus payload is still emitted, so a regression that drops the broad except would fail the test.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_withdraw.py
LOCKhart07: Good catch — not intentional. Fixed in 7e5158ed.

Switched the snapshot to **cost basis** by summing `initialValue` across unredeemable positions, matching the cost-basis semantics of the existing formula at `agent_performance_summary_abci/behaviours.py:762` (`(total_traded - total_traded_settled) / token_divisor`). Also added `round(value, 2)` to match the normal round's rounding.

Verified empirically against a live wallet (`0x9070...4BA4`) that `initialValue` tracks `currentSize × avgPrice` (cost basis of currently held shares), not a frozen-at-original-buy value — that wallet had `totalBought ≈ 2 × size` from a prior partial sell and `initialValue` matched `size × avgPrice`, not `totalBought × avgPrice`. So the field stays in cost-basis units and won't cause a mark-to-market vs cost-basis jump on the FE.

bennyjo (approved):
Both prior findings addressed in 7e5158ed: snapshot now uses cost-basis (`sum(initialValue)`, rounded to 2dp) matching the normal-round formula, and the compute + helper call is wrapped in a broad try/except with a falsifying test that asserts the payload is still emitted on unparseable values. LGTM.

Issue #946, opened 2026-05-07
FSM audit: boolean-precedence bug in is_first_period and 4 other findings


DavidMinarsch:
This issue tracks findings from an FSM audit of the eight first-party skills under `packages/valory/skills/` (audit run on commit `2434e9f5`, branch `main`). The framework CLI checks (`autonomy analyse fsm-specs / handlers / dialogues / docstrings`) all pass; the items below were caught by manual review against the open-autonomy FSM audit checklist.

## Summary

| ID  | Severity | Skill | Title |
|-----|----------|-------|-------|
| H-1 | High     | `decision_maker_abci` | Operator-precedence ambiguity in `is_first_period` |
| M-1 | Medium   | `market_manager_abci` | Test fixtures `DUMMY_BETS_HASH` / `DUMMY_PARTICIPANT_TO_BETS_HASH` defined twice |
| M-2 | Medium   | `decision_maker_abci` | Unused `Event.FETCH_ERROR` enum member (also raises `analyse fsm-specs` warning) |
| M-3 | Medium   | `chatui_abci` | `ChatuiLoadRound.negative_event = Event.DONE` collapses negative-vote path |
| L-1 | Low      | `staking_abci` | `ServiceEvictedRound.end_block` overrides parent's `NotImplementedError` guard with empty body |

---

## H-1: Operator-precedence ambiguity in `is_first_period`

**File:** `packages/valory/skills/decision_maker_abci/behaviours/base.py:269-276`

Mixed `and`/`or` with no explicit parentheses. Python evaluates `and` before `or`, so the expression is `(period_count == 0 AND not enabled) OR (mock_data is None)`. Implication: when `mock_data is None`, the property returns `True` regardless of `period_count` — every period gets treated as the "first period" in that branch. Suspicious for a property literally named `is_first_period`.

```python
return (
    self.synchronized_data.period_count == 0
    and not self.benchmarking_mode.enabled
    or self.shared_state.mock_data is None
)
```

**Suggested fix (please confirm intent):**

```python
return (
    self.synchronized_data.period_count == 0
    and (not self.benchmarking_mode.enabled or self.shared_state.mock_data is None)
)
```

---

## M-1: Test fixtures defined twice with incompatible values

**File:** `packages/valory/skills/market_manager_abci/tests/test_rounds.py:77-84` and `:278-287`

[...]

PR #947, opened 2026-05-08
fix(fsm): address audit findings from issue #946

DIvyaNautiyal07:
# Fix five FSM audit findings (issue #946)

## Summary

Five fixes from the FSM audit on commit `2434e9f5`:

| ID  | Severity | Skill | Title |
|-----|----------|-------|-------|
| H-1 | High     | `decision_maker_abci` | Operator-precedence in `is_first_period` |
| M-1 | Medium   | `market_manager_abci` | Test fixtures shadowed under the same name |
| M-2 | Medium   | `decision_maker_abci` | Dead `Event.FETCH_ERROR` enum member |
| M-3 | Medium   | `chatui_abci` | `negative_event` aliased to `done_event` |
| L-1 | Low      | `staking_abci` | Empty `end_block` override silenced parent guard |

## H-1 — production behaviour change worth a closer look

The buggy expression `(period_count == 0 AND not enabled) OR (mock_data is None)` collapsed to `True OR True = True` every period in production, because `mock_data` is only ever assigned in `decision_receive.py` inside `_parse_dataset_row` (benchmarking-only). So `is_first_period` was permanently True in production today.

After the fix, the property returns the right answer per its name: True only on the very first period of a production service's lifetime.

**The behavioural consequence:** `_update_policy_tools` in `storage_manager.py` is called only when `is_first_period` is True. With the old bug, that meant it ran every period in production — so the IPFS-sourced global accuracy data (`tools_accuracy_hash`) was re-applied to the local policy on every cycle (gated by the timestamp check `_should_use_global_info`). After the fix, this re-apply runs only at startup. The local policy continues to learn from bet outcomes via `policy.py:251`, but the IPFS sync no longer happens periodically.

I'm flagging this for a domain reviewer to confirm:

- If the team intended `_update_policy_tools` to run every period (i.e. periodic IPFS sync was a feature), then this fix needs a separate explicit periodic-sync check rather than relying on `is_first_period`.
- If `is_first_period` was meant to mean what its name says (run-once-at-startup), this fix is correct and the prior every-period IPFS sync was an unintended side effect of the bug.

I went with the run-once-at-startup interpretation because the original-author intent (commit `6b03e08343` from July 2024) was to add benchmarking-mode awareness to a property that was previously just `period_count == 0`. The benchmarking-aware mode-split below preserves that intent without the precedence trap:

```python
@property
def is_first_period(self) -> bool:
    if self.benchmarking_mode.enabled:
        return self.shared_state.mock_data is None
    return self.synchronized_data.period_count == 0
```

## M-1 — duplicate test fixtures

`DUMMY_BETS_HASH` / `DUMMY_PARTICIPANT_TO_BETS_HASH` were declared twice with incompatible shapes in `market_manager_abci/tests/test_rounds.py`. Renamed the second pair to `*_SERIALIZED` so editing one no longer silently breaks the other.

## M-2 — dead enum member

`decision_maker_abci.Event.FETCH_ERROR` is unreferenced; every transition uses `MarketManagerEvent.FETCH_ERROR`. Removed the local member.

## M-3 — chatui negative-vote path

`ChatuiLoadRound.negative_event = Event.DONE` collapsed the rejected-vote outcome into success. `ChatuiLoadBehaviour.async_act()` always sends `vote=True` today so the path is structurally unreachable, but the alias hid a missing transition. Added `Event.FAIL = "fail"` (matching the same name+value already used by `agent_performance_summary_abci`) and routed `negative_event` to a self-loop, matching how `NONE` and `NO_MAJORITY` are handled.

## L-1 — empty `end_block` override

`ServiceEvictedRound.end_block` had an empty body that silenced `DegenerateRound`'s `NotImplementedError` defensive guard. Removed the override; sibling final states (`CheckpointCallPreparedRound`, `FinishedStakingRound`) already match this pattern.

## Test plan

- [x] 6 new `is_first_period` cases covering production × benchmarking × period_count, including an explicit "production ignores mock_data" lock-in
- [x] M-3 alias-distinct-from-DONE assertion
- [x] L-1 lock-in test asserting `ServiceEvictedRound.end_block()` raises `NotImplementedError`
- [x] All 2,299 tests pass across the six affected skills
- [x] black, isort, flake8, mypy, pylint, bandit, darglint, abci-docstrings clean
- [x] FSM specs regenerated for `chatui_abci`, `decision_maker_abci`, `trader_abci` and `autonomy analyse fsm-specs` passes for all three apps
- [ ] `autonomy packages lock` + `autonomy push-all` (run before final push)


OjusWiZard (commented):
Self-review pass — 2 important, 4 advisory.

The H-1 fix and FSM consistency look correct. Main ask: the new `is_first_period` tests don't actually distinguish the new mode-split from the original buggy expression — a revert would go undetected today. Inline comments below.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_base.py
```diff
@@ -613,24 +613,76 @@ def test_policy_property_returns_value(self) -> None:
...
+        self._set_period_count(behaviour, 0)
+        assert behaviour.is_first_period is False
+
+    def test_is_first_period_benchmarking_initialised_later_period(self) -> None:
+        """In benchmarking with mock_data set and period advanced, not the first period."""
+        behaviour = self.behaviour
+        behaviour.benchmarking_mode.enabled = True
+        behaviour.shared_state.mock_data = MagicMock()
+        self._set_period_count(behaviour, 5)
+        assert behaviour.is_first_period is False
```
OjusWiZard: **Important — H-1 regression isn't actually locked in by these tests.**

All 6 parametrized cases happen to pass under both the old buggy expression `(period_count == 0 and not enabled) or mock_data is None` and the new mode-split. Walking through each:

| Case | Old result | New result |
|---|---|---|
| prod, count=0, mock=None | `True` | `True` |
| prod, count=1, mock=None | `True` | `False` |
| prod, count=0, mock=set | `True` | `True` |
| prod, count=1, mock=set | `False` | `False` |
| bench, count=0, mock=None | `True` | `True` |
| bench, count=5, mock=None | `True` | `True` |
| bench, count=0, mock=set | `False` | `False` |
| bench, count=5, mock=set | `False` | `False` |

The distinguishing case (`prod, count=1, mock=None`) is missing. Suggest adding:

```python
def test_is_first_period_production_mock_data_none_later_period(self) -> None:
    """Production + mock_data=None + period_count>=1 must return False.

    Under the original buggy expression
    ``(period_count == 0 and not enabled) or mock_data is None``
    this returned True (the trailing OR fired unconditionally). The fixed
    expression ignores mock_data in production mode.
    """
    behaviour = self.behaviour
    behaviour.benchmarking_mode.enabled = False
    behaviour.shared_state.mock_data = None
    self._set_period_count(behaviour, 1)
    assert behaviour.is_first_period is False
```

This is the minimal test that fails against the original code and passes only with the fix.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_base.py
```diff
@@ -613,24 +613,76 @@ def test_policy_property_returns_value(self) -> None:
...
 
-    def test_is_first_period_true(self) -> None:
-        """Test `is_first_period` property when it is the first period."""
+    def _set_period_count(self, behaviour: Any, value: int) -> None:
+        """Set the period count via the `db.reset_index` attribute it actually reads."""
+        behaviour.synchronized_data.db.reset_index = value
+        # Guard: if synchronized_data is ever migrated to a real AbciAppDB,
+        # `reset_index` becomes a read-only property and the assignment above
+        # would silently no-op, making every is_first_period test pass by
+        # accident. Fail loudly here instead.
```
OjusWiZard: **Important — comment is factually wrong.**

"silently no-op" is incorrect on both branches:

- On a real `AbciAppDB`, `reset_index` is a `@property` with no setter — assigning to it raises `AttributeError`, not a silent no-op.
- On the current `MagicMock`, attribute writes are *always* accepted — also not a no-op.

The `assert ... == value` guard on the next line is the right defence; only the prose explanation is misleading. Suggest rewording to something like:

```
# Guard: `synchronized_data.db` is a MagicMock that accepts any attribute
# write. If this test ever runs against a real `AbciAppDB` (where
# `reset_index` is a read-only @property), the assignment above would raise
# AttributeError. Either way, this assert keeps the test honest about which
# value `is_first_period` will actually read.
```

packages/valory/skills/decision_maker_abci/tests/behaviours/test_base.py
OjusWiZard: **Advisory — boilerplate could be parametrized.**

The four benchmarking tests share the same 3-line setup and the new property ignores `period_count` entirely in benchmarking mode, so the two `_set_period_count` variants exercise the same branch. Could collapse to:

```python
@pytest.mark.parametrize("mock_data,expected", [(None, True), (MagicMock(), False)])
def test_is_first_period_benchmarking(self, mock_data, expected) -> None:
    behaviour = self.behaviour
    behaviour.benchmarking_mode.enabled = True
    behaviour.shared_state.mock_data = mock_data
    assert behaviour.is_first_period is expected
```

Drops ~20 lines without losing branch coverage. Style-only — feel free to defer.

packages/valory/skills/chatui_abci/rounds.py
```diff
@@ -65,7 +66,7 @@ class ChatuiLoadRound(VotingRound):
     payload_class = ChatuiPayload
     synchronized_data_class = SynchronizedData
     done_event = Event.DONE
-    negative_event = Event.DONE
+    negative_event = Event.FAIL
```
OjusWiZard, line 69: **Advisory — `Event.FAIL` self-loop has no escape hatch.**

Structurally consistent with the existing `NONE` / `NO_MAJORITY` self-loops in this FSM, and `ChatuiLoadBehaviour.async_act()` always sends `vote=True` today, so the path is currently unreachable (the PR description calls this out).

Forward hazard worth noting: if a future behaviour change ever emits `vote=False` and a negative quorum is reached, the round will loop on `ROUND_TIMEOUT` (30s) indefinitely with no operator alert. Not a blocker — flagging so a future maintainer adding a `vote=False` path adds a retry counter or warning log alongside it.

packages/valory/skills/chatui_abci/tests/test_rounds.py
```diff
@@ -94,7 +94,19 @@ def test_abci_app_transition_function(abci_app: ChatuiAbciApp) -> None:
...
     assert FinishedChatuiLoadRound in tf
     assert tf[ChatuiLoadRound][Event.DONE] == FinishedChatuiLoadRound
+    assert tf[ChatuiLoadRound][Event.FAIL] == ChatuiLoadRound
     assert tf[ChatuiLoadRound][Event.NONE] == ChatuiLoadRound
     assert tf[ChatuiLoadRound][Event.ROUND_TIMEOUT] == ChatuiLoadRound
     assert tf[ChatuiLoadRound][Event.NO_MAJORITY] == ChatuiLoadRound
     assert tf[FinishedChatuiLoadRound] == {}
+
+
+def test_chatui_load_round_negative_event_distinct_from_done() -> None:
```
OjusWiZard, line 104: **Advisory — class-level lock is sufficient; runtime dispatch coverage is a nice-to-have.**

This test pins `negative_event is Event.FAIL` and the transition table assertion above pins `(ChatuiLoadRound, FAIL): ChatuiLoadRound`. Both correct.

What's missing for full behavioural coverage is end-to-end `VotingRound.end_block` dispatch with a `vote=False` payload threshold — i.e., feed `n_agents` payloads with `vote=False`, call `end_block`, assert returned event is `Event.FAIL` not `Event.DONE`. Since `VotingRound.end_block` is framework-tested upstream, this is nice-to-have, not blocking.

packages/valory/skills/staking_abci/tests/test_rounds.py
```diff
@@ -270,6 +270,17 @@ def test_service_evicted_round_initialization(self) -> None:
         round_ = ServiceEvictedRound(synchronized_data=MagicMock(), context=MagicMock())
         assert isinstance(round_, ServiceEvictedRound)
 
+    def test_end_block_raises_for_final_state(self) -> None:
+        """Calling `end_block` on a final state must raise NotImplementedError.
+
+        DegenerateRound's `end_block` is a defensive guard — final states are
+        never advanced past, so any attempt to call this surfaces a framework
+        misuse rather than silently returning None.
```
OjusWiZard: **Advisory — "rather than silently returning None" loses context post-merge.**

The phrase only makes sense if a reader knows about the deleted `end_block` override (which returned `None` implicitly). After this PR merges, a future reader sees this docstring with no link back to the override that motivated it.

Suggest either dropping the trailing clause or making the link explicit, e.g.:

```
DegenerateRound's `end_block` raises NotImplementedError as a defensive
guard. ServiceEvictedRound previously overrode it with an empty body that
implicitly returned None — this test locks in the parent contract so the
override cannot quietly come back.
```

packages/valory/skills/decision_maker_abci/tests/behaviours/test_base.py
DIvyaNautiyal07: Good catch on the docstring being non-obvious. The distinguishing case (`prod, count=1, mock=None`) is actually `test_is_first_period_production_later` — under the buggy expression `(False AND True) OR (None is None)` returns True, and the assertion `is_first_period is False` would fail. So the regression IS locked in. But the original one-line docstring ("period_count>0 returns False") didn't make that obvious — beefed it up in da9e9fb9 to spell out the buggy expression and why this case discriminates.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_base.py
DIvyaNautiyal07: Right on both counts — applied your wording in da9e9fb9.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_base.py
DIvyaNautiyal07: Done — collapsed to one `test_is_first_period_benchmarking` parametrized over `(mock_data, period_count, expected)` in da9e9fb9. Kept the production tests unparametrized since `production_later` is a named regression-locking case I want to keep grep-able.

packages/valory/skills/chatui_abci/rounds.py
DIvyaNautiyal07: Acknowledged — flagging this for whoever later wires up a real `vote=False` path. The current self-loop on FAIL is structurally consistent with NONE / NO_MAJORITY, but you're right that a 30s-loop with no operator alert is the implicit cost. Not addressing in this PR.

packages/valory/skills/chatui_abci/tests/test_rounds.py
DIvyaNautiyal07: Agreed — class-level lock is what this PR adds; runtime dispatch coverage would be nice-to-have but `VotingRound.end_block` is exercised upstream. Leaving as-is.

packages/valory/skills/staking_abci/tests/test_rounds.py
DIvyaNautiyal07: Applied your wording in da9e9fb9.

LOCKhart07 (approved):

OjusWiZard (approved):

PR #948, opened 2026-05-11
feat(agent_performance_summary): sell-aware Polystrat performance summary

LOCKhart07:
## Summary

- PR #942 introduced Polymarket sells; the per-bet view in the performance summary still assumed every bet was a buy. This rewrites the per-bet path to be sell-aware while leaving aggregates (already sell-aware on the subgraph side) untouched.
- Queries: drop \`where: { isBuy: true }\`; pull \`isBuy\` into the selection set.
- New \`_allocate_fifo\` groups rows per \`(participant, question, outcomeIndex)\`, sorts chronologically, FIFO-allocates sells against buys, and emits one enriched buy per original buy. Each enriched buy carries \`participant_remaining_cost\` so invalid-market refunds pro-rate correctly across multi-buy participants. Orphan sells are logged (with unattributed USDC magnitude and sell id) and dropped; null-question/null-outcome rows are logged and skipped.
- Hybrid status rule (per the 2026-05-11 Slack thread with Iason/Roman): invalid → INVALID; fully exited → WON/LOST by realized PnL sign (PENDING on exact break-even); partial sell or no sell → today's resolution-based logic with the won-but-unredeemed → PENDING carve-out preserved. No new \`BetStatus\` value.
- Profit = realized PnL + redemption PnL on the unsold remainder. Losing remainder books as loss; invalid refund pro-rates by remaining cost basis across the participant. Collapses to today's formula for the no-sell, single-buy case.
- \`total_payout\` field now reports realized proceeds + redemption value (was: \`shares\` iff \`WON\` else 0). A partial-sell PENDING row correctly reports its realized cash.
- \`_fetch_bet_from_subgraph\` reuses \`_allocate_fifo\` + \`_format_single_bet\` so the list and single-bet API surfaces stay consistent.
- **Omenstrat behaviour unchanged** — the only shared-file edits (\`behaviours.py\`, \`bet_payout_checker.py\`) are no-ops for Omen, where pre-rewrite \`total_payout > 0\` ≡ \`status == \"won\"\`. See review-fix sections below.
- 50+ new/updated unit tests; all touched modules at 100% coverage.

Design doc: \`WITHDRAWAL_PERFORMANCE_SUMMARY_SPEC.md\` (not checked in).

## Diff size breakdown

Total diff is **+1461 / −199 = 1660 touched lines**. Production-code surface is ~31 % — the bulk is tests and auto-generated package-hash propagation:

| Bucket | Lines touched | Share | Files |
| --- | --- | --- | --- |
| **Production code** | **+361 / −160 (521)** | **31 %** | 4 |
| Tests | +1066 / −5 (1071) | 65 % | 3 |
| Package hash propagation (\`autonomy packages lock\`) | +34 / −34 (68) | 4 % | 10 |

**Production-code files (the real review surface):**

| File | + | − |
| --- | ---: | ---: |
| \`graph_tooling/polymarket_predictions_helper.py\` (Polymarket-only) | 333 | 157 |
| \`achievements_checker/bet_payout_checker.py\` (shared, no-op on Omen) | 17 | 0 |
| \`behaviours.py\` (shared, no-op on Omen) | 7 | 1 |
| \`graph_tooling/queries.py\` (Polymarket-only) | 4 | 2 |

The helper (\`polymarket_predictions_helper.py\`) is ~94 % of the prod diff — the FIFO allocator, hybrid status rule, sell-aware profit, \`total_payout\` rewrite, and the multi-buy invalid-market refund attribution (round 2). The other three files combined are 28 lines (review-driven guards in \`behaviours.py\` and \`bet_payout_checker.py\`, the query selection-set tweak).

### Round 1 — \`cc3f8caf\` (addresses @LOCKhart07)

Two latent bugs in downstream \`total_payout\` consumers:

- **\`behaviours.py:1779\`** — \`winning_trades\` ROI gate filtered on \`total_payout > 0\`, which under sell-aware semantics matches fully-sold-at-loss bets. Switched to \`status == BetStatus.WON.value\`. Omen behaviour unchanged.
- **\`bet_payout_checker.py\`** — \`bet[\"settled_at\"].replace(...)\` crashed with \`AttributeError\` for fully-sold-at-2x-pre-resolution bets (status=WON via PnL sign, settled_at=None). Added guards on \`status == WON\` and \`settled_at is not None\`.

Plus: \`_get_prediction_status\` fully-exited branch classifies on realized PnL directly (matches docstring); \`_calculate_bet_profit\` dropped \`Optional[float]\` (never returned None); \`SHARES_EPSILON\` docstring typo.

### Round 2 — \`25380ea0\` (addresses @DIvyaNautiyal07 and @OjusWiZard)

**P1 — multi-buy invalid-market refund over-counted across buys.** The previous \`_redemption_value\` invalid branch pro-rated participant.totalPayout by each buy's local fraction (\`remaining_cost / cost\`), summing to N × refund for N never-sold buys on the same invalid market. Fix: stamp \`participant_remaining_cost\` on every enriched buy in \`_allocate_fifo\` (sum across the participant after FIFO settles); pro-rate by \`remaining_cost / participant_remaining_cost\` in \`_redemption_value\`. Sum across all buys now equals \`participant.totalPayout\`. Zero observed impact on the test safe (no invalid markets), but multi-buy is common (17 groups on this safe). Three new \`TestInvalidMarketAttribution\` tests pin sum invariant + per-buy formula.

**HIGH — legacy fallbacks read \`bet[\"shares\"]\`** instead of \`fifo[\"remaining_shares\"]\`. \`_redemption_value:374\` and \`_calculate_bet_profit:420\` overstated redemption for partially-sold enriched buys hitting the resolution-with-null-winningIndex defensive branch. Fix: use \`fifo[\"remaining_shares\"]\`; legacy raw bets get the same answer (the \`_fifo_state\` default), enriched buys get the correct remainder.

**Cleanups and observability:**

- Drop \`countedInTotal\` from both Polymarket query selection sets + test fixture. No production consumer; widened the wire contract for nothing.
- Replace bare \`\"won\"\`/\`\"invalid\"\`/\`\"lost\"\` string literals in \`fetch_position_details\` with \`BetStatus.X.value\` to match the sibling convention.
- Drop a redundant \`if resolution else None\` ternary in \`_get_prediction_status\`.
- Per-bet WARNING when \`_allocate_fifo\` drops a bet for null question / outcomeIndex (subgraph mapping gap). Operators can quantify the gap directly from logs.
- Orphan-sell warning now includes the unattributed USDC and the sell row id (pro-rated: \`proceeds_total * (shares_to_consume / sell_original_shares)\`).
- Defensive WARNING when invalid-market refund is dropped because \`participant_remaining_cost == 0\` (impossible on real data; guards a division-safety branch against silent money loss).
- Inline comment in \`_allocate_fifo\` documenting that the bet.id tiebreaker is lex-sorted hex (not numeric log-index) and why same-block buy+sell is rare in practice.
- Extend \`bet_payout_checker\` guard comment to explicitly call out both LOST and INVALID exclusions as intentional.
- Strengthen \`test_with_resolution\` and \`test_losing_bet_net_profit\` to pin \`net_profit\`/\`total_payout\` (memory rule: tests must be falsifiable). New focused zero-share-buy deadlock-prevention test + complementary \`cost==0 + total_payout==0\` silent-path test for branch coverage.

**Outside-diff:**

- \`fetch_position_details\` string-literal status comparisons → \`BetStatus.X.value\` (above).
- Test value-assertion gaps strengthened.

**Deferred to follow-up #949:** \`FifoState\` TypedDict refactor (reviewer-blessed defer; tracked as tech-debt/follow-up).

## Test plan

- [x] \`uv run pytest packages/valory/skills/agent_performance_summary_abci/tests/\` — 1014 passing
- [x] 100% coverage on all touched modules
- [x] \`tomte tox -e black-check / isort-check / flake8 / mypy / pylint / darglint\` — all green
- [x] \`autonomy packages lock\` — hashes propagated through \`agent_performance_summary_abci\` and dependents
- [x] Live verification against test safe \`0x9070A95134A06290141e94B6Dd453c9c88dc4BA4\` via \`GET /api/v1/agent/prediction-history\`: 208 rows, page 1 distribution 44 WON / 52 LOST (29 wholly-lost + 23 sold-at-loss) / 4 PENDING. Hybrid rule producing the expected output end-to-end; aggregates unchanged.
- [x] Eyeball bets list for fully-sold positions: 23 \`status: \"lost\"\` rows with non-zero \`total_payout\` (e.g. bet \$1, sold for \$0.78–\$0.99) correctly classifying per the hybrid rule.

## Notes

- \`SHARES_EPSILON\` is 10,000 base units (= 0.01 share = 1 cent equivalent at full redemption). Empirical dust on the test safe is 2,500 base units, well inside this bound.
- Non-blocking post-merge follow-ups:
  - Subgraph null-question gap (vanilla-only, fix expected via reindex in 1–2 weeks); now observable via the new WARNING.
  - Orphan-sell frequency observability metric; now quantifiable via the unattributed-USDC log field.
  - #949 — tighten FIFO-state typing with a \`TypedDict\`.


packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -194,9 +281,11 @@ def _format_single_bet(
...
         bet_timestamp = bet.get("blockTimestamp")
         resolution_timestamp = resolution.get("blockTimestamp") if resolution else None
-        # Per-bet payout: shares for winning bets, 0 otherwise
-        bet_shares = float(bet.get("shares", 0)) / USDC_DECIMALS_DIVISOR
-        total_payout = bet_shares if prediction_status == BetStatus.WON.value else 0.0
+        # Realized cash returned to the agent: sell proceeds + redemption value
+        # on the unsold remainder. Independent of status — a partial-sell PENDING
+        # bet correctly reports the proceeds already booked.
+        realized_proceeds = fifo["allocated_proceeds"] / USDC_DECIMALS_DIVISOR
+        total_payout = realized_proceeds + self._redemption_value(bet, fifo)
```
LOCKhart07, line 320: **Blocking — `total_payout` semantic shift breaks two downstream consumers.**

This rewrite makes `total_payout` = realized proceeds + redemption for **all** statuses (PENDING / LOST / WON), but two existing consumers still assume the old "0 unless redeemed-WON" semantics:

- `skills/agent_performance_summary_abci/behaviours.py:1779` — `winning_trades = [item for ... if item.get("total_payout", 0) > 0]` gates the "Total ROI" metric on `>= MIN_TRADES_FOR_ROI_DISPLAY` winning trades. Post-PR, any LOST bet sold at non-zero proceeds is counted as a winning trade. Your own test-plan numbers (23 sold-at-loss bets in the 208-row sample) all flip from excluded → included, inflating the winning_trades count.
- `skills/agent_performance_summary_abci/achievements_checker/bet_payout_checker.py:99` — once `roi = total_payout / bet_amount > POLYMARKET_ACHIEVEMENT_ROI_THRESHOLD (1.5)`, the checker calls `bet["settled_at"].replace("Z", "+00:00")`. The new hybrid status rule classifies a fully-sold-at-profit-pre-resolution bet as WON via realized PnL sign — but `_format_single_bet` derives `settled_at` from `resolution.blockTimestamp`, so it's `None` when there's no resolution. Result: `AttributeError: 'NoneType' object has no attribute 'replace'` for a sold-at-2x-cost-pre-resolution bet. The 35 new helper-level unit tests don't catch this; it needs an integration test that feeds the new prediction_history shape into the achievement checker.

Two reasonable fixes: (a) keep `total_payout` semantics aligned with pre-PR (0 for non-WON/non-INVALID) and expose realized cash on a new field (e.g. `realized_proceeds`); or (b) update both consumers — the ROI filter should key on `status == "won"`, and the bet_payout_checker should skip bets where `settled_at is None`.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -217,93 +306,180 @@ def _format_single_bet(
...
-            if int(outcome_index) == int(winning_index):
-                # Check if winnings have been redeemed
-                # totalPayout is only updated when the agent redeems (from subgraph)
-                total_payout = float(bet.get("totalPayout", 0))
-                if total_payout == 0:
-                    # Won but not redeemed yet - treat as pending
-                    return BetStatus.PENDING.value
+        fifo = self._fifo_state(bet)
+        if fifo["original_shares"] > 0 and fifo["remaining_shares"] <= SHARES_EPSILON:
+            net_profit = self._calculate_bet_profit(bet) or 0.0
```
LOCKhart07: Docstring at lines 437-451 says the fully-sold branch "classifies by realized PnL sign," but this calls `_calculate_bet_profit(bet)` which adds `(redemption - remaining_cost)` for resolved bets — not pure realized PnL. The two only diverge for resolved-and-fully-sold bets where `remaining_cost > 0` due to the epsilon (sub-cent in practice). Either rewrite to use just `proceeds - allocated_cost` here, or update the docstring to say "by sell-aware net profit sign".

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -217,93 +306,180 @@ def _format_single_bet(
...
+        if int(outcome_index) != int(winning_index):
+            return 0.0  # losing outcome — remainder worthless
+
+        # Won. Redemption only books once participant.totalPayout flips positive.
+        participant_total_payout = float(bet.get("totalPayout", 0))
+        if participant_total_payout <= 0:
+            return 0.0
+        return remaining_shares / USDC_DECIMALS_DIVISOR
+
+    def _calculate_bet_profit(self, bet: Dict) -> Optional[float]:
```
LOCKhart07: Nit: signature says `Optional[float]` and the docstring says `or None if uncomputable`, but no branch in this function ever returns `None`. Two callers carry dead defensive code as a result:

- `_format_single_bet` line 299: `round(net_profit, 3) if net_profit is not None else None`
- `_get_prediction_status` line 462: `or 0.0`

Either drop `Optional` from the signature (and clean both callers) or wire up an actual `None` path.

packages/valory/skills/agent_performance_summary_abci/tests/graph_tooling/test_polymarket_predictions_helper.py
```diff
@@ -2622,3 +2622,702 @@ def test_data_null_returns_none_without_attribute_error(  # type: ignore[no-unty
...
+
+        buys = fetcher._allocate_fifo(bets, participant_total_payout=0)
+
+        assert buys == []
+
+    def test_dust_remainder_treated_as_fully_sold(self) -> None:
+        """Sub-cent dust after a sell counts as fully exited (epsilon)."""
+        fetcher = _make_fetcher()
+        # Buy 5.2625 shares (5_262_500 base units), sell 5.26 shares (5_260_000).
+        # Remaining = 2500 base units < epsilon 100000? No, 2500 is the dust.
```
LOCKhart07: Nit: `SHARES_EPSILON` is `10_000`, not `100000`. The phrasing is also a little tangled — easier to read as `# Remaining = 2500 base units < SHARES_EPSILON (10_000); treated as dust.`

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -217,93 +306,180 @@ def _format_single_bet(
...
-        # Use participant-level totalPayout pro-rated by bet amount.
+        outcome_index = bet.get("outcomeIndex")
+        cost = original_cost / USDC_DECIMALS_DIVISOR
+        cost_realized = allocated_cost / USDC_DECIMALS_DIVISOR
+        remaining_cost = max(cost - cost_realized, 0.0)
+
         if winning_index is not None and int(winning_index) < 0:
             total_payout = float(bet.get("totalPayout", 0)) / USDC_DECIMALS_DIVISOR
-            return total_payout - bet_amount
+            return total_payout * (remaining_cost / cost) if cost > 0 else 0.0
```
DIvyaNautiyal07: **P1: invalid-market refund over-counted across multiple buys per participant.**

`bet.totalPayout` here is the participant-level refund (one number per (agent, market)), but `remaining_cost / cost` is THIS buy's local remaining-fraction. For a participant with N never-sold buys on the same invalid market, each buy's redemption equals the full participant refund, summing to N × refund.

Worked example: 2 buys @ $1 each, participant.totalPayout = $1.5:
- buy1: redemption = 1.5 × (1.0/1.0) = $1.5
- buy2: redemption = 1.5 × (1.0/1.0) = $1.5
- Sum = $3.0 reported; actual = $1.5.

This also regresses the single-bet path. Pre-PR `_fetch_bet_from_subgraph` divided by `total_traded_p` (each buy's share of total invested) which was correct; the new code reuses this buggy formula on both surfaces. The list path was already buggy in the same way pre-PR (no behavior regression there), but aggregate behaviour is unaffected (server-side); the bug only impacts per-bet `net_profit` / `total_payout` rendering for invalid markets with rebuys.

Fix sketch: attribute the participant refund proportionally to each buy's remaining cost basis across the group. Easiest place is `_allocate_fifo` (it sees all open buys at once) — store the attributed refund on the enriched buy and have `_redemption_value` read it for the invalid branch.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -217,93 +306,180 @@ def _format_single_bet(
...
+        Legacy callers without FIFO state (``original_shares`` absent) skip the
+        "fully exited" branch when the buy never had shares to begin with, so
+        the existing four-status behaviour is preserved for pre-rewrite tests.
+
+        :param bet: bet dict, optionally enriched with FIFO state fields
+        :return: one of the ``BetStatus`` string values
+        """
+        question = bet.get("question") or {}
+        resolution = question.get("resolution") or {}
+        winning_index = resolution.get("winningIndex") if resolution else None
```
DIvyaNautiyal07: **Nit:** redundant `if resolution else None`. After `resolution = question.get("resolution") or {}` on the previous line, `resolution` is either `{}` or a non-empty dict, and `{}.get("winningIndex")` already returns `None`. The ternary adds no information — `winning_index = resolution.get("winningIndex")` is equivalent.

packages/valory/skills/agent_performance_summary_abci/tests/graph_tooling/test_polymarket_predictions_helper.py
```diff
@@ -2622,3 +2622,702 @@ def test_data_null_returns_none_without_attribute_error(  # type: ignore[no-unty
...
+            remaining_shares=1.0,
+            allocated_proceeds=0.4,
+            allocated_cost=0.5,
+            original_cost=1.0,
+            outcome_index=0,
+            winning_index=1,
+        )
+        assert fetcher._get_prediction_status(bet) == "lost"
+
+    def test_invalid_market_overrides_fully_sold(self) -> None:
```
DIvyaNautiyal07, line 3107: **Nit: missing test for multi-buy invalid-market refund attribution.** Every invalid-market test currently uses a single buy, so the participant-vs-buy attribution discussed on `_redemption_value:364` is not exercised. Even a small test (two buys on the same invalid market, assert `sum(_redemption_value(b, _fifo_state(b)) for b in buys) <= participant.totalPayout`) would catch any regression and pin the chosen attribution semantics.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -217,93 +306,180 @@ def _format_single_bet(
...
-        else:
-            # Fallback: use shares to determine
-            if bet_shares > 0:
-                return bet_shares - bet_amount
-            else:
-                return -bet_amount
+        if outcome_index is None or winning_index is None:
+            # Indices missing — fall back to "shares > 0 implies redemption"
+            # signal used by the legacy fallback path.
+            bet_shares = float(bet.get("shares", 0) or 0) / USDC_DECIMALS_DIVISOR
```
OjusWiZard: **HIGH — Legacy fallback uses raw `bet["shares"]` instead of `fifo["remaining_shares"]`.**

For an enriched buy that has been partially sold by `_allocate_fifo`, `bet["shares"]` is the full original buy amount (unchanged from the subgraph row), while `fifo["remaining_shares"]` is what's actually unsold. This branch fires when `winning_index is None` but `resolution` exists — in that state the redemption is computed on original shares rather than the remainder, silently overstating `total_payout` and `net_profit` for partially-sold positions.

Suggested fix:
```python
remaining_shares = fifo["remaining_shares"] / USDC_DECIMALS_DIVISOR
return remaining_shares if remaining_shares > 0 else 0.0
```
`fifo` is already in scope (line 333 signature). Same fix applies to `_calculate_bet_profit:420`.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -217,93 +306,180 @@ def _format_single_bet(
...
         winning_index = resolution.get("winningIndex")
+        outcome_index = bet.get("outcomeIndex")
+
+        if winning_index is not None and int(winning_index) < 0:
+            redemption = self._redemption_value(bet, fifo)
+            return realized_pnl + (redemption - remaining_cost)
+
+        if outcome_index is None or winning_index is None:
+            # Legacy fallback: shares > 0 → treat as redemption; else loss.
+            bet_shares = float(bet.get("shares", 0) or 0) / USDC_DECIMALS_DIVISOR
```
OjusWiZard: **HIGH — Same overestimation as `_redemption_value:374`.**

`bet["shares"]` is the original buy amount; for a partially-sold enriched buy this overstates redemption by the value of already-sold shares. Replace with `fifo["remaining_shares"] / USDC_DECIMALS_DIVISOR` (already in scope at line 400).

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -88,26 +94,106 @@ def fetch_predictions(
...
+        :return: list of enriched buy dicts, one per original buy that produced
+            a group key
+        """
+        groups: Dict[Any, List[Dict[str, Any]]] = defaultdict(list)
+        for bet in bets:
+            question = bet.get("question") or {}
+            question_id = question.get("id")
+            outcome_index = bet.get("outcomeIndex")
+            if question_id is None or outcome_index is None:
+                continue
```
OjusWiZard, line 144: **MEDIUM — Silent drop with no observability.**

Bets with `question_id is None` or `outcome_index is None` are skipped without a log. The PR body notes this is a known subgraph gap pending reindex — but until then, there's no way to quantify the size of the gap from logs. Suggest:

```python
if question_id is None or outcome_index is None:
    self.logger.warning(
        "Skipping bet id=%s: question_id or outcomeIndex is None (subgraph gap)",
        bet.get("id"),
    )
    continue
```

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -88,26 +94,106 @@ def fetch_predictions(
...
+            groups[(question_id, int(outcome_index))].append(bet)
+
+        output: List[Dict[str, Any]] = []
+        for (question_id, outcome_index), group_bets in groups.items():
+            group_bets.sort(
+                key=lambda b: (int(b.get("blockTimestamp", 0) or 0), b.get("id", ""))
```
OjusWiZard, line 157: **MEDIUM — Within-block FIFO ordering is lexicographic on hex bet IDs.**

Polymarket bet IDs are `0x<txhash><logindex>` hex strings; lexicographic ordering does not match actual log-index order. For a buy + immediate sell in the same block, the proceeds allocation across buys is arbitrary. Financial impact is bounded to within-block ordering but worth either documenting or making deterministic via log-index parsing.

If within-block determinism doesn't matter in practice, an explicit comment here noting that would head off future confusion.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -88,26 +94,106 @@ def fetch_predictions(
...
 
         # Format individual bets (not grouped by market)
-        items = self._format_predictions(
-            all_bets_with_questions, safe_address, status_filter
-        )
+        items = self._format_predictions(all_buys, safe_address, status_filter)
+
+        return {"total_predictions": len(all_buys), "items": items}
+
+    def _allocate_fifo(
```
OjusWiZard, line 113: **MEDIUM — Consider a `FifoState` TypedDict / frozen dataclass.**

The enriched buy carries five FIFO float fields plus `totalPayout` and is consumed by string key in five downstream methods (`_fifo_state`, `_redemption_value`, `_calculate_bet_profit`, `_get_prediction_status`, `_format_single_bet`). A missing key silently defaults to `0.0` via `_fifo_state`. A `FifoState` TypedDict would let mypy enforce key presence and document the invariants `0 <= remaining_shares <= original_shares` and `allocated_cost <= original_cost`.

Project already uses `@dataclass(frozen=True)` in `payloads.py` / `models.py`, so the pattern is established. Non-blocking — followup is fine.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -88,26 +94,106 @@ def fetch_predictions(
...
+                        if head["remaining_shares"] <= 0:
+                            open_buys.popleft()
+
+                    if shares_to_consume > 0:
+                        self.logger.warning(
+                            "Orphan sell on (question=%s, outcomeIndex=%s): "
+                            "%s shares unmatched against any buy",
+                            question_id,
+                            outcome_index,
+                            shares_to_consume,
```
OjusWiZard, line 212: **MEDIUM — Orphan-sell warning omits the dollar magnitude.**

The warning logs `shares_to_consume` (the unmatched share count) but not `proceeds_total` (the cash proceeds that won't be attributed to any buy). Operators have no way to measure how much money is leaking from the data. Suggest also logging `proceeds_total / USDC_DECIMALS_DIVISOR` so the gap can be quantified.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
OjusWiZard: **LOW — Silent drop of invalid-market refund when `cost == 0`.**

Economically correct (zero collateral → zero refund), but the `else 0.0` discards `total_payout` with no log. If a future code path ever delivers a zero-cost buy with `totalPayout > 0`, the refund will silently disappear. Either add an inline comment asserting the invariant or log a warning before the drop. Minor.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/queries.py
```diff
@@ -319,11 +319,13 @@
...
   ) {
     totalPayout
-    bets(where: { isBuy: true }) {
+    bets {
       id
       outcomeIndex
       amount
       shares
+      isBuy
+      countedInTotal
```
OjusWiZard: **MEDIUM — `countedInTotal` is selected but never consumed.**

`countedInTotal` appears in this query and in `GET_POLYMARKET_SPECIFIC_BET_QUERY` (line 475), and only test fixtures read it. Selecting unused fields widens the API contract and misleads future readers who assume it drives some branching logic. Either wire up the consumer (it sounds like a natural gate for which bets count toward aggregates) or drop from both selection sets.

packages/valory/skills/agent_performance_summary_abci/achievements_checker/bet_payout_checker.py
```diff
@@ -66,6 +69,15 @@ def update_achievements(self, achievements: Achievements, **kwargs: Any) -> bool
 
         achievements_updated = False
         for bet in prediction_history.items:
+            # Sell-aware: only fire on resolved-and-redeemed wins. A
+            # fully-sold-at-loss bet would have total_payout > 0 (realized
+            # proceeds) and a fully-sold-at-profit-pre-resolution bet would
+            # have status=WON but settled_at=None — both must be excluded.
+            if bet.get("status") != BetStatus.WON.value:
```
OjusWiZard, line 81: **MEDIUM — INVALID-status bets are silently excluded; the comment only documents two of the three exclusions.**

The `status != BetStatus.WON.value` guard correctly drops LOST-sold-for-proceeds and WON-pre-resolution, but it also drops `INVALID` resolutions. An invalid-market refund can legitimately exceed the ROI threshold (e.g., agent bet 1 USDC, refund 1.5 USDC). These achievements are permanently suppressed without an explicit decision recorded.

Either:
- Add `BetStatus.INVALID.value` to the accepted statuses with its own `settled_at is not None` guard, or
- Extend the comment (lines 72-75) to explicitly call out the INVALID exclusion as intentional.

OjusWiZard (dismissed):
Approving — the four prior review comments are correctly addressed in `cc3f8caf`, the FIFO allocator is sound, and coverage on touched modules is genuinely 100% (verified). Inline comments below for follow-ups; nothing blocking. Two notes that live outside the diff hunks and so couldn't be left inline:

- **`polymarket_predictions_helper.py:847-850` (`fetch_position_details`)** — `if status == "won" or status == "invalid"` / `elif status == "lost"` uses bare string literals while the sibling file `predictions_helper.py:565-567` uses `BetStatus.WON.value` etc. Renaming an enum value would silently misclassify the user-visible `to_win` field here. Worth aligning with the sibling convention in a follow-up.
- **Test value-assertions** — coverage is 100% but a couple of integration paths execute without value checks: `test_losing_bet_net_profit` (`TestFetchBetFromSubgraph`) asserts only `result is not None`, and `TestFormatSingleBet.test_with_resolution` doesn't pin `net_profit` / `total_payout`. A regression that returns the wrong number would pass. Also no test explicitly exercises the zero-share-buy guard (`polymarket_predictions_helper.py:165`) — coverage hits the line but no assertion validates the deadlock-prevention behaviour. Worth strengthening the assertions in a follow-up.

Nice work — the hybrid status rule + FIFO design comes through cleanly in the tests.

bennyjo (dismissed):
Approving — previously raised review concerns (winning_trades ROI gate misclassifying sold-at-loss bets, and the bet_payout_checker AttributeError on `settled_at=None` for sold-pre-resolution wins) are addressed in cc3f8caf with matching regression tests. FIFO allocation, hybrid status rule, and downstream consumer updates all check out; test coverage is thorough.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Fixed in 25380ea0. `_allocate_fifo` now stamps `participant_remaining_cost` on every enriched buy (sum across the participant after FIFO settles); `_redemption_value` invalid branch pro-rates by `remaining_cost / participant_remaining_cost`. Sum across all buys equals `participant.totalPayout`; partial-sell + never-sold combos attribute proportionally. Three new tests in `TestInvalidMarketAttribution` pin the sum invariant + per-buy formula for never-sold, partial-sell, and cross-outcome cases.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Dropped in 25380ea0. After `resolution = question.get("resolution") or {}`, the ternary was dead — `{}.get("winningIndex")` returns `None` anyway.

packages/valory/skills/agent_performance_summary_abci/tests/graph_tooling/test_polymarket_predictions_helper.py
LOCKhart07: Added in 25380ea0 — `TestInvalidMarketAttribution` has three focused cases: 2 never-sold buys @ \$1 each (sum = participant.totalPayout); partial-sell + never-sold combo (refund splits 1:2 by remaining cost); cross-outcome buys on the same invalid market (refund splits by remaining cost across the participant, not per-outcome). Each pins both the per-buy value and the sum invariant.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Fixed in 25380ea0. `_redemption_value` legacy fallback now uses `fifo["remaining_shares"] / USDC_DECIMALS_DIVISOR`. Legacy raw bets get the same answer (the `_fifo_state` default falls back to `bet["shares"]`); enriched partial-sell buys correctly book only the unsold remainder.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Same fix as `_redemption_value:374` — `_calculate_bet_profit` legacy fallback now uses `fifo["remaining_shares"]` (25380ea0).

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Added in 25380ea0 — per-bet WARNING with the bet id when `question_id` or `outcomeIndex` is `None`. Operators can now grep for "subgraph gap" to quantify the drop and track it post-merge as expected from the PR body.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Documented in 25380ea0 (inline comment above the sort key). Same-block buy+sell pairs are rare for this agent (buys and sells happen in separate FSM rounds — bet placement vs withdrawal sweep, normally many blocks apart). If it ever becomes a real ordering need, the path forward is exposing a numeric `logIndex` on the `Bet` entity in the subgraph; the comment captures that.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Filed as follow-up: #949. Scoped to `_fifo_state`'s return type + propagating through the five consumers; full `EnrichedBuy` dataclass deferred per the original suggestion until additional consumers land.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Updated in 25380ea0. Orphan warning now logs the **unattributed USDC** (pro-rated: `proceeds_total * (shares_to_consume / sell_original_shares) / USDC_DECIMALS_DIVISOR`) and the sell row id. Operators can now triage individual leaks and aggregate the magnitude over time.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
LOCKhart07: Added defensive WARNING in 25380ea0. When `participant_remaining_cost == 0` (the rewritten denominator) and `total_payout > 0`, we log the refund amount + bet id before returning 0.0. Inline comment notes the invariant ("cost == 0 is impossible for real Polymarket buys"). Two complementary tests pin both branches (silent zero-zero path + warning-when-money-leaks path).

packages/valory/skills/agent_performance_summary_abci/graph_tooling/queries.py
LOCKhart07: Dropped from both queries + the `_raw_bet` test fixture in 25380ea0. Verified zero production consumers via grep; the spec itself flagged it as optional ("helper can re-derive from resolution"). Resolution status comes from `question.resolution` everywhere we need it, so the field was widening the contract for nothing.

packages/valory/skills/agent_performance_summary_abci/achievements_checker/bet_payout_checker.py
LOCKhart07: Took the "extend the comment" path in 25380ea0 — the comment now explicitly calls out both exclusions (LOST sold-for-proceeds + INVALID refund). Rationale: the achievement copy ("Agent closed a bet at {roi}× ROI") celebrates a prediction win, and an invalid-market refund is a different event class (market cancelled, no prediction skill involved). If we ever want to surface invalid-recovery achievements, it should be a separate `POLYSTRAT_INVALID_RECOVERY` type with its own copy — out of scope here.

LOCKhart07:
Thanks both for the thorough reviews. All 12 inline comments + the two outside-diff notes addressed in 25380ea0:

**Outside-diff (from @OjusWiZard's review body):**

- **`fetch_position_details:847–850` string literals** → swapped to `BetStatus.WON.value` / `BetStatus.INVALID.value` / `BetStatus.LOST.value` to match the sibling convention in `predictions_helper.py:565-567`. An enum rename can no longer silently misclassify the user-visible `to_win` field.
- **Test value-assertion gaps** → strengthened `test_with_resolution` and `test_losing_bet_net_profit` to pin `net_profit` / `total_payout` / `status` (memory-rule: tests must be falsifiable). Added a focused zero-share-buy deadlock-prevention test for `_allocate_fifo`'s `open_buys` guard, plus a complementary `cost==0 + total_payout==0` silent-path test for branch coverage on the new defensive warning.

**Stats after the fix commit:**
- 1014 skill tests pass (was 1002 in the previous round)
- 100% coverage on all four touched modules (`polymarket_predictions_helper.py`, `queries.py`, `behaviours.py`, `bet_payout_checker.py`)
- All linters green (black, isort, flake8, mypy, pylint, darglint)

**Follow-up filed:** #949 for the `FifoState` TypedDict refactor.

jmoreira-valory (approved):
Approving — prior review has been exceptionally thorough (16 comments across self-review, DIvyaNautiyal07, and OjusWiZard), and the author addressed every finding in `cc3f8caf` and `25380ea0` with concrete pointers and regression tests, plus the FifoState TypedDict suggestion split out as #949.

I went looking for anything not yet flagged — FP accumulation in `_allocate_fifo`'s `allocated_cost +=` (sub-cent leak via the invalid-refund pro-rate, FP-equality break-even unreachable in practice), zero-share buy data-quality edge cases — and none rise above noise on a PR this carefully reviewed.

LGTM.

DIvyaNautiyal07 (approved):

bennyjo (approved):

PR #950, opened 2026-05-11
fix(contracts): return bytes (not hex str) for calldata in 4 dev contracts

jmoreira-valory:
## Summary

Standardizes \`realitio\`, \`realitio_proxy\`, \`conditional_tokens\`, and \`relayer\` to the mech/open-autonomy convention: every \`build_*_tx\` style method now returns \`data\` as \`bytes\`, matching what the strict multisend introduced by [open-autonomy PR #2477](https://github.com/valory-xyz/open-autonomy/pull/2477) (\"chore(packages): drop type-only web3/eth_typing imports across 16 contracts\", commit \`b2228ddde\`) expects.

## Background

PR #2477 removed \`HexBytes(data)\` from \`multisend.encode_data\`, which had been silently coercing hex-string \`data\` fields into bytes. Post-cleanup, \`encode_data\` is strictly bytes-typed: passing a hex string causes \`bytes + str\` concat to raise \`TypeError\`, and the contract-api dispatcher reports \`error\` performative.

Trader's skills don't hit the regression because every \`MultisendBatch(data=HexBytes(self.built_data))\` callsite wraps the contract output defensively. But downstream consumers of these same contracts (the omen recovery skills in market-creator and market-resolver) didn't have that wrap and crashed in production this week — the bond-claim path on market-resolver was 100% broken for several days.

The proper fix is contract-side: emit bytes-by-construction so every consumer (with or without a wrap) works against any multisend implementation. mech / mech-predict / open-autonomy itself already follow this convention; trader had it on 8 of 16 dev-contract callsites but not the other 8. This PR closes that gap.

## What changed

8 contract methods × one-line each:

| Contract | Method |
|---|---|
| \`conditional_tokens.build_redeem_positions_tx\` | \`dict(data=data)\` → \`dict(data=bytes.fromhex(data[2:]))\` |
| \`conditional_tokens.build_merge_positions_tx\` | same |
| \`realitio.build_claim_winnings\` | same |
| \`realitio.get_submit_answer_tx\` | same |
| \`realitio.build_withdraw_tx\` | same |
| \`realitio_proxy.build_resolve_tx\` | same |
| \`relayer.build_operator_deposit_tx\` | same |
| \`relayer.build_exec_tx\` | same |

8 test mocks × one-line each: \`encode_abi.return_value = b\"\\xaa\\xbb\"\` (unrealistic — real \`encode_abi\` returns hex strings) → \`encode_abi.return_value = \"0xaabb\"\`. Assertions kept as bytes; they now actually exercise the conversion path instead of being trivially-true no-ops.

## No breaking changes for trader

Every trader skill consumer wraps with \`HexBytes(...)\` already:
- \`decision_maker_abci/behaviours/reedem.py\` (3 callsites)
- \`decision_maker_abci/behaviours/bet_placement.py\`
- \`decision_maker_abci/behaviours/base.py\`
- \`mech_interact_abci/behaviours/request.py\`

\`HexBytes(bytes_value)\` is a no-op (HexBytes is a bytes subclass), so the contract returning bytes flows through every existing callsite unchanged.

## Test plan
- [x] isort / black / mypy / flake8 / darglint — all green
- [x] Unit tests: 3693 passed (no skill-side breakage)
- [x] \`autonomy packages lock\`: 4 contract hashes rotated, 6 dependent yaml/agent/service references auto-bumped
- [x] \`autonomy push-all\`: published to IPFS

## Follow-ups (separate PRs in downstream repos)

After this merges, the omen-skill workarounds we shipped in market-creator and market-resolver can be deleted. See: \`market-resolver/.claude/plans/multisend_calldata_bytes_alignment.md\`.


packages/valory/contracts/realitio/contract.py
```diff
@@ -246,7 +246,7 @@ def build_claim_winnings(
             abi_element_identifier="claimWinnings",
             args=(question_id, *claim_params),
         )
-        return dict(data=data)
+        return dict(data=bytes.fromhex(data[2:]))
```
LOCKhart07, line 249: **Blocking** — this change silently breaks `simulate_claim_winnings` 20 lines below (L269).

That method consumes its own `build_claim_winnings` and previously did `"data": data[2:]` to strip the `"0x"` prefix from the hex string before passing to `eth.call`:

```python
data = cls.build_claim_winnings(...)["data"]
ledger_api.api.eth.call({..., "data": data[2:]})
```

Now that `data` is **bytes**, `data[2:]` drops the first two **raw bytes** of the calldata — half the 4-byte function selector — so every simulation `eth_call` is malformed.

Production impact: `simulate_claim_winnings` is called from `decision_maker_abci/behaviours/reedem.py:614` to gate Realitio bond claims. The simulation will revert on every claim, so the claim path is permanently skipped on Omenstrat.

The existing `test_simulate_claim_winnings_*` tests don't catch it — they only assert on the returned `{"data": True/False}` and never inspect what's passed to the mocked `eth.call`.

Fix: drop the slice at L269 (`"data": data,` — it's already raw bytes), and tighten the test with `mock_ledger_api.api.eth.call.assert_called_once_with({..., "data": b"\xaa\xbb"})` so this regression is caught next time.

Adamantios (approved):
LGTM, just need to be careful as it's a breaking change and these may be used from a behaviour.

packages/valory/contracts/realitio/contract.py
jmoreira-valory: fixed

LOCKhart07 (approved):
Re-reviewed `e8dd630`. The `simulate_claim_winnings` fix is correct — `"data": data` now passes the full bytes, and the three `assert_called_once_with(..., "data": b"\xaa\xbb")` assertions in the simulate tests guard against the selector-chop regression class. Mocks were also updated to return the realistic `"0xaabb"` hex string, so the tests actually exercise the conversion path. Approving.

PR #951, opened 2026-05-12
chore: bump genai + funds_manager hashes

DIvyaNautiyal07:
> **Do not merge** until v0.38.0-rc1 passes QA and is released in Pearl.

## Summary

Bumps third-party OA package hashes to upstream releases:

| package | upstream tag | new hash |
|---|---|---|
| `connection/valory/x402/0.1.0` | [`valory-xyz/genai@v7.1.1`](https://github.com/valory-xyz/genai/releases/tag/v7.1.1) | `bafybeicqjg3d6pnlcsu4442qt5tmsjqkqdtc3ulvbjgoh53jyr7z3lbqqi` |
| `connection/valory/genai/0.1.0` | [`valory-xyz/genai@v7.1.1`](https://github.com/valory-xyz/genai/releases/tag/v7.1.1) | `bafybeighrq424xlroekehgk2oaqxkvziq22gffqzodm4ynhjgzhch4pcq4` |
| `skill/valory/funds_manager/0.1.0` | [`valory-xyz/funds-manager@v3.0.1`](https://github.com/valory-xyz/funds-manager/releases/tag/v3.0.1) | `bafybeifff3amddh44muocd4nfo3txu5lltgjtycvqa4jiwgde6idwvdoii` |

Also bumps `upstream_pins` in `pyproject.toml` so `check-third-party-hashes` resolves against the new tags (`genai@v7.0.0 -> v7.1.1`, `funds-manager@v3.0.0 -> v3.0.1`). Dependent local-package hashes (chatui_abci, check_stop_trading_abci, decision_maker_abci, trader_abci, tx_settlement_multiplexer_abci, agent/trader, services/{trader_pearl, polymarket_trader}) auto-rehashed via `autonomy packages lock`.

Source PRs:
- valory-xyz/genai [PR 32](https://github.com/valory-xyz/genai/pull/32) (released as v7.1.1)
- valory-xyz/funds-manager [PR 39](https://github.com/valory-xyz/funds-manager/pull/39) (released as v3.0.1)

## Consumer-side fix: trader_abci None-balance propagation

funds-manager v3.0.1 returns `balance=None` when a sub-call reverts, decode fails, or a whole-chain multicall fails (it's the "unknown" marker, not zero). The consumer-side caveat flagged in [funds-manager#39](https://github.com/valory-xyz/funds-manager/pull/39) is addressed here so \`trader_abci\` doesn't fold \`None\` into \`0\` and emit spurious top-up deficits.

- `_get_adjusted_funds_status`: returns early without recomputing the considered balance when any of the wxDAI / native / USDC balances is \`None\`. Sets \`native.deficit = None\` before returning so /funds-status surfaces "unknown" instead of the funds-manager-computed int (which is overstated since it was derived from one token alone against the combined-target topup).
- `_merge_usdc_e_into_pusd`: skips the merge when either side is \`None\`. Sets \`pusd.deficit = None\`. Drops the USDC.e row only when USDC.e's own balance is unknown — if USDC.e is readable but pUSD isn't, USDC.e stays visible so the operator retains diagnostic info.
- 5 new tests in \`TestGetAdjustedFundsStatus\` covering each None-propagation path, seeded with funds-manager-style int deficits and asserting they get cleared.

### Pearl-side behavior (verified against v0.37.x bins/middleware)

Pearl's \`get_funding_requests\` at \`operate/services/service.py:1422-1442\` reads \`amounts["deficit"]\` and wraps it in \`BigInt(...)\`. \`BigInt(None)\` raises \`TypeError\`, which Pearl catches and:
- logs \`[FUNDING MANAGER] Invalid funding amount None for asset {asset}…\`
- sets the asset's funding amount to \`0\` for that cycle (skips the top-up)

This is the correct semantic for transient sub-call reverts (which is what funds_manager's \`None\` marker indicates): skip one top-up cycle, retry next round when fresh balances arrive. For *prolonged* RPC outages, Pearl's repeated warning logs are the operator-visible signal — out of scope for this PR.

## Regression scan

**x402 / genai consumers:**
- Zero hits on the old \`"upstream returned 402"\` log substring (the message body changed upstream).
- The single trader-side \`except PaymentError\` site at \`packages/valory/connections/genai/connection.py:334\` catches the base class, so the new \`PaymentRejectedAfterRetryError(PaymentError)\` subclass falls through it correctly.
- \`decode_x_payment_response\` signature unchanged; consumer at \`genai/connection.py:252\` still binds.

**funds_manager consumers:**
- All three trader-side imports still resolve:
  - \`GET_FUNDS_STATUS_METHOD_NAME\` (constant, value \`"get_funds_status"\`)
  - \`AccountRequirements\`, \`FundRequirements\` (Pydantic models)
- Real \`aea-config.yaml\`-shaped fund_requirements parses via \`FundRequirements.from_dict\` (\`{chain_name: {agent|safe: {token_addr: {topup, threshold}}}}\`); \`TokenRequirement.is_native\` auto-detection still works.

## Local CI checklist (all green)

- [x] \`uv lock --check\`
- [x] \`autonomy packages lock --check\`
- [x] \`tomte tox -e check-hash\`
- [x] \`tomte tox -e check-packages\`
- [x] \`tomte tox -e check-dependencies\`
- [x] \`tomte tox -e check-third-party-hashes\` (against new upstream pins)
- [x] \`tomte tox -e check-doc-hashes\`
- [x] \`tomte tox -e check-abci-docstrings -e check-abciapp-specs -e check-handlers\`
- [x] \`tomte tox -e liccheck\`
- [x] \`tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint\`
- [x] \`tomte tox -p -e safety -e bandit\`
- [x] \`tomte tox -e gitleaks\`
- [x] \`tomte tox -e py3.12-darwin\` (3889 passed)
- [x] \`autonomy push-all\`
- [x] /prove-it on chatui_abci + trader_abci integration boundaries
- [x] Live agent end-to-end: \`GET /funds-status\` returned real Gnosis-RPC multicall data matching a public-RPC cross-check; \`POST /chatui-prompt\` round-tripped through the bumped genai connection to Gemini and back.

## Pearl QA gate (v0.38.0-rc1)

Before merge, QA Pearl v0.38.0-rc1 against an agent running this PR's hashes and confirm:

- [ ] Under nominal conditions (all sub-calls succeed): \`/funds-status\` returns int \`deficit\` values; Pearl tops up as expected.
- [ ] Under simulated partial RPC outage (e.g. one ERC20 sub-call reverted): \`/funds-status\` returns \`deficit: null\` for the affected carrier entry; Pearl logs \`[FUNDING MANAGER] Invalid funding amount None\` and skips the top-up that cycle.
- [ ] Next cycle, after RPC recovers: \`/funds-status\` returns a real int deficit; Pearl resumes top-up normally.
- [ ] No spurious top-up requests emitted while the sub-call was reverted (the bug this PR fixes).

## Test plan

- [ ] Wait for green CI matrix (Linux/macOS/Windows, py3.10-3.14)
- [ ] Pearl QA gate above
- [ ] On merge, downstream consumers can \`autonomy packages sync --update-packages\` against this branch's hashes


DIvyaNautiyal07:
Hold on merging this until v0.38.0-rc1 passes QA and is released in Pearl.

OjusWiZard (dismissed):

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -429,7 +429,15 @@ def _get_adjusted_funds_status(self) -> FundRequirements:
...
                 # Using CoinGecko to get real-time exchange rate since USDC and POL have different prices
                 usdc_status = safe_balances.tokens[chain_config["usdc_address"]]
-                usdc_balance = int(usdc_status.balance or 0)
+
+                if usdc_status.balance is None:
+                    self.context.logger.warning(
+                        "USDC balance unknown (sub-call reverted); "
+                        "skipping USDC->POL adjustment to avoid spurious deficit."
+                    )
+                    return funds_status
```
LOCKhart07, line 439: Per `packages/valory/skills/funds_manager/behaviours.py:309-323` (v3.0.1), funds_manager only sets `deficit=None` for tokens whose own balance is None — siblings get a computed deficit. So when only `usdc_status.balance` is None and `native_status.balance` is known, funds_manager has set `native.deficit = max(topup - native_balance, 0)` against the **combined**-target topup/threshold (since the operator-config puts the combined target on the native token, and the wrapper relies on combining native + USDC-as-POL).

Returning early without overwriting then propagates that overstated `native.deficit` to `/funds-status`, so Pearl still emits a top-up — just smaller than the pre-fix `int(balance or 0)` worst case, not eliminated. The PR description's "spurious top-up deficits" goal isn't fully met for the per-token revert case (only for the whole-chain failure case, which hits the `native.balance is None` check below).

Recommend setting `native_status.deficit = None` here before `return funds_status` so downstream sees "unknown" instead of a misleading int.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -476,15 +484,28 @@ def _get_adjusted_funds_status(self) -> FundRequirements:
                 wrapped_native_status = safe_balances.tokens[
                     chain_config["wrapped_native_address"]
                 ]
-                adjustment_balance = int(wrapped_native_status.balance or 0)
+                if wrapped_native_status.balance is None:
+                    self.context.logger.warning(
+                        "Wrapped-native balance unknown (sub-call reverted); "
+                        "skipping wxDAI->xDAI consolidation to avoid spurious deficit."
+                    )
+                    return funds_status
```
LOCKhart07, line 494: Same concern as the Polygon path above. When only `wrapped_native_status.balance` is None on Gnosis and `native_status.balance` is known, funds_manager v3.0.1 (`funds_manager/behaviours.py:309-323`) will have set `native.deficit` to a non-None value computed from native alone against the combined-target topup/threshold — i.e., overstated, since wxDAI might have covered it.

Returning early preserves that overstated funds_manager value. Recommend `native_status.deficit = None` before `return funds_status`, so /funds-status carries "unknown" (matching the `wxDAI.deficit = None` already set by funds_manager) instead of a stale int.

The tests at `test_handlers.py:1205` (wxDAI case) and `:1286` (Polygon USDC case) pre-set `native_token_in.deficit = None` before calling, so they assert "unchanged" against a manually-seeded None — masking the production path where funds_manager would have left an int there. If the fix is updated to clear the deficit, those test preconditions can be dropped and the tests will exercise the real flow.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -530,7 +551,15 @@ def _merge_usdc_e_into_pusd(
         usdc_e = safe_balances.tokens[usdc_e_addr]
         pusd = safe_balances.tokens[pusd_addr]
 
-        combined = int(usdc_e.balance or 0) + int(pusd.balance or 0)
+        if usdc_e.balance is None or pusd.balance is None:
+            self.context.logger.warning(
+                "USDC.e or pUSD balance unknown (sub-call reverted); "
+                "skipping merge to avoid spurious pUSD deficit."
+            )
+            del safe_balances.tokens[usdc_e_addr]
```
LOCKhart07: Analogous to the wrapper's wxDAI/USDC handling: `pusd.deficit` is left at whatever funds_manager computed against pUSD alone. Since pUSD carries the combined threshold/topup target (USDC.e is folded in), that funds_manager-computed deficit overstates when USDC.e balance is the unknown sibling.

Recommend `pusd.deficit = None` before the `del`+return. Matches the "signal unknown" semantics funds_manager itself uses for the token whose own balance is None.

Note: `test_merge_skips_when_usdc_e_balance_unknown` (test_handlers.py:1323) currently pre-sets `pusd.deficit = 25_000_000` and asserts "unchanged" — same masking issue as the wrapper tests.

packages/valory/skills/trader_abci/handlers.py
DIvyaNautiyal07: Good catch, you're right. Confirmed funds_manager v3.0.1 at `behaviours.py:316-322` computes `deficit = max(topup - balance, 0)` against each token's own topup/threshold. Since trader's operator config puts the combined target on the native entry (USDC at 0/0, native carries the threshold/topup), a USDC-only revert leaves `native.deficit` overstated.

Fixed in 6f844039: set `native_status.deficit = None` before the early return. The corresponding test now seeds `native_token_in.deficit = 900` (the value funds_manager would have set against native alone) and asserts it gets cleared to `None`.

packages/valory/skills/trader_abci/handlers.py
DIvyaNautiyal07: Agreed. Same fix applied for the Gnosis wxDAI path: `native_status.deficit = None` before the early return.

The test now drops the `native_token_in.deficit = None` precondition and seeds it with `990` (what funds_manager would compute from native alone, 1000 topup - 10 balance against a 500 threshold). Without the fix the early return preserves 990 and the assertion `is None` fails — exercising the real production flow instead of masking it. 6f844039

packages/valory/skills/trader_abci/handlers.py
DIvyaNautiyal07: Same fix in `_merge_usdc_e_into_pusd`: `pusd.deficit = None` before `del`+return when either side is unknown. Covers the case where USDC.e is the unknown sibling and funds_manager has left `pusd.deficit` overstated against the combined pUSD target.

The merge test now seeds `pusd.deficit = 25_000_000` and asserts it gets cleared, instead of asserting "unchanged". 6f844039

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -476,15 +485,29 @@ def _get_adjusted_funds_status(self) -> FundRequirements:
...
             return funds_status
 
-        actual_considered_balance = int(native_status.balance or 0) + adjustment_balance
+        if native_status.balance is None:
+            self.context.logger.warning(
+                "Native balance unknown (sub-call reverted); "
+                "leaving native deficit unchanged to avoid spurious top-up request."
+            )
+            return funds_status
+
```
LOCKhart07, line 509: Non-blocking nit: this branch relies on funds_manager v3.0.1's contract that `balance=None ⇒ deficit=None` (per `funds_manager/behaviours.py:309-323`), so we can leave `native_status.deficit` untouched here. The other three None branches now defensively set `deficit = None` themselves; mirroring that here would decouple us from a single-upstream invariant. Not a current bug — just a consistency / defense-in-depth thought for a follow-up if you want it.

LOCKhart07 (dismissed):

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -530,7 +553,16 @@ def _merge_usdc_e_into_pusd(
...
         pusd = safe_balances.tokens[pusd_addr]
 
-        combined = int(usdc_e.balance or 0) + int(pusd.balance or 0)
+        if usdc_e.balance is None or pusd.balance is None:
+            self.context.logger.warning(
+                "USDC.e or pUSD balance unknown (sub-call reverted); "
+                "skipping merge and clearing pUSD deficit."
+            )
+            pusd.deficit = None
+            del safe_balances.tokens[usdc_e_addr]
```
bennyjo: Is dropping the USDC.e entry deliberate when only `pusd.balance` is `None`? In that case USDC.e's balance is still known and informative; discarding it loses visibility for the operator even though the merge can't proceed. If the intent is purely "don't surface an overstated combined deficit," keeping the USDC.e row intact while only clearing `pusd.deficit` preserves the diagnostic without changing the deficit semantic. Happy to leave as-is if dropping is intentional.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -429,7 +429,16 @@ def _get_adjusted_funds_status(self) -> FundRequirements:
...
                 # Using CoinGecko to get real-time exchange rate since USDC and POL have different prices
                 usdc_status = safe_balances.tokens[chain_config["usdc_address"]]
-                usdc_balance = int(usdc_status.balance or 0)
+
+                if usdc_status.balance is None:
+                    self.context.logger.warning(
+                        "USDC balance unknown (sub-call reverted); "
+                        "skipping USDC->POL adjustment and clearing native deficit."
+                    )
+                    native_status.deficit = None
```
bennyjo, line 438: The fix produces `deficit=None` on the carrier entry to signal "unknown," which flows through `get_response_body()` as JSON `null`. Pearl is the actor that decides whether to top up based on this field — has its handling of `deficit: null` been verified? If Pearl reads null as "no deficit" rather than "unknown, retry," then under exactly the partial-RPC-outage failure mode this PR addresses, the agent silently stops getting topped up. The PR already gates merge on v0.38.0-rc1 Pearl QA; worth pinning a specific check there if it's not already on the list. (Same question applies at lines 493, 503, and 561.)

bennyjo (commented):
Two questions on the None-propagation fix.

packages/valory/skills/trader_abci/handlers.py
DIvyaNautiyal07: Good catch — it wasn't deliberate, the `del` was a carry-over from the success path where USDC.e gets folded into pUSD and the row is redundant. In the skip path that reasoning doesn't apply when USDC.e is still readable.

Updated in b06ba6f4: the row is dropped only when USDC.e's own balance is `None`; if USDC.e is readable but pUSD isn't, the row stays. `test_merge_skips_when_pusd_balance_unknown` now asserts the USDC.e row is preserved with its known balance.

packages/valory/skills/trader_abci/handlers.py
DIvyaNautiyal07: Verified against the Pearl middleware bundled in v0.37.x at `operate/services/service.py:1422-1442`:

```python
funding_requests[chain_str][address][asset] = BigInt(amounts["deficit"])
except (ValueError, TypeError):
    logger.warning("[FUNDING MANAGER] Invalid funding amount None…")
    funding_requests[chain_str][address][asset] = 0
```

`BigInt(None)` raises `TypeError` → caught → warning logged → asset's funding set to `0` for that cycle (top-up skipped). So Pearl already treats `deficit: null` as "skip this cycle, log a warning", which matches the trader semantic for transient sub-call reverts: skip one cycle, retry next round when fresh balances arrive.

For prolonged RPC outages the repeated `[FUNDING MANAGER] Invalid funding amount None` warnings are the operator-visible signal. That's a deeper integration concern outside this PR's scope, but flagged.

Added an explicit Pearl QA gate to the PR description covering the three states (nominal int, partial-outage null, recovery).

bennyjo (approved):
Re-review LGTM — both prior-cycle asks addressed cleanly (deficit explicitly cleared; USDC.e row preserved when only pUSD reverts). Tests now seed funds-manager-style int deficits and discriminate against the pre-fix path. None-handling verified against funds-manager v3.0.1's `balance=None ⇒ deficit=None` invariant. Safe to merge after the Pearl v0.38.0-rc1 QA gate.

LOCKhart07 (approved):

PR #952, opened 2026-05-12
feat: Omenstrat withdrawal mode (Phase 1 + sell-aware perf-summary)

LOCKhart07:
Ports the operator-controlled withdrawal feature (#942 / #943 / #948) to the Omen venue. `POST /api/v1/withdrawal` on Omenstrat now arms a one-shot sell-off of every held outcome-token position via FPMM `sell`, mirroring the Polystrat behaviour but using Safe multisend instead of CLOB FAK.

## Phase 1 — sweep behaviour

- New FSM event `WITHDRAW_OMEN_DONE` + `OmenWithdrawalPayload` / `PostOmenWithdrawalPayload` / `PostOmenWithdrawRound` / `FinishedOmenWithdrawRound` / `FinishedOmenWithdrawTxRound`. Wired end-to-end through `tx_settlement_multiplexer`, `trader_abci` composition, and `decision_maker_abci` transitions.
- `OmenWithdrawBehaviour`: fetch CT `user_positions` (authoritative current balance) + omen `fpmmTrades` (FPMM metadata), filter via `get_withdrawable_positions` to the OPEN bucket, per-position marginal-price sizing with calcSellAmount halve-retry against the slippage-headroom invariant `n_estimate*(1+slip) <= B`, build `(setApprovalForAll, sell)*N` multisend with `setApprovalForAll` deduped per-FPMM across outcomes.
- `PostOmenWithdrawBehaviour`: parse settled tx receipt via `FixedProductMarketMakerContract.parse_sell_events`, filter against the planned-FPMM allowlist persisted by the planning round, record fills per `FPMMSell` log; top-level errors on `tx.status=0` or Safe `ExecutionFailure`.
- Shared chatui JSON-store I/O extracted to `OmenWithdrawalStore` (`behaviours/omen_withdrawal_store.py`) — both behaviours delegate via a lazy `_store` property. Unifies the previously-drifted `_record_top_level_error` signatures.
- Contracts: `FixedProductMarketMakerContract.parse_sell_events` with layered defensive guards (per-field structural checks + try/except safety net + module logger) + `get_pool_balances_via_ct` (workaround for the Olas FPMM build dropping `getPoolBalances()`); `ConditionalTokensContract.build_set_approval_for_all_tx` returning raw bytes to match sibling `build_*_tx` methods.
- Subgraph: `WithdrawablePosition` dataclass + `get_withdrawable_positions` + new `fetch_withdrawal_creator_fpmms` query in `market_manager_abci`. `answerFinalizedTimestamp` parsing normalises the `"0"`/`0`/malformed sentinels via a local `_is_finalized_timestamp` helper.
- Params: `withdrawal_slippage`, `withdrawal_return_buffer`, `dust_epsilon_wxdai` (defaults 0.01 / 0.05 / 1e16 wei). Declared in `decision_maker_abci/skill.yaml`, propagated to `trader_abci/skill.yaml` (composed-app schema) and `agents/trader/aea-config.yaml` so the framework's param-schema check passes at skill load.
- 501 dropped on `/api/v1/withdrawal`. **Server now binds to `127.0.0.1` by default** (was `0.0.0.0`) — see [Auth posture](#auth-posture) below.

## Phase 3A — sell-aware perf-summary

Fixes three latent bugs in `_calculate_bet_net_profit` / `_build_market_context` — all caused by the pre-FIFO formula reading the signed raw `amount` field, which a sell row makes negative:

- `winning_total_amount` denominator used the signed raw amount; a sell on the winning side polluted it. Now sums FIFO `remaining_cost`.
- Invalid-market refund numerator used `bet_amount / total_traded`; sells flipped the ratio sign. Now uses `remaining_cost / participant_remaining_cost` scoped **per-FPMM** (a cross-FPMM denominator would skew the per-market refund attribution by `(N-1)/N`).
- Losing-bet PnL booked `-bet_amount`; a sell row's negative amount surfaced as `+profit`. Now books `realized_pnl - remaining_cost`. **Critical fix.**
- `_build_market_context` now divides `totalPayout` and `totalTraded` by `WEI_TO_NATIVE` at construction. Pre-fix the subgraph's raw-wei values were multiplied against unitless wxDAI ratios, producing a ~1e18× scale error on every invalid-refund / winning-payout call. Tests previously masked it by injecting hand-built wxDAI-scale ctx dicts.

The FIFO allocator is the module-level `allocate_fifo(bets, logger)` (with a one-line wrapper as `PredictionsFetcher._allocate_fifo` for internal callers). Groups by `(fpmm.id, outcomeIndex)`, sorts chronologically, consumes sells against prior buys, emits enriched buys with `original_cost` / `remaining_shares` / `allocated_proceeds` / `allocated_cost` / `participant_remaining_cost`. **Fully integer-only on wei amounts** — `int()` parsing with try/except, integer mul-then-divide for ratios (`(proceeds_total * take) // -row_shares`), bounded per-step rounding error at 1 wei vs the unbounded float drift past 2^53 it replaces. Defensive sign-disagreement guard between `amount` and `outcomeTokenAmount`. `_format_predictions` re-sorts the output DESC-by-`blockTimestamp` at the consumer boundary so the allocator's ASC-within-group ordering invariant doesn't leak into the FE display.

## Phase 3B — per-position funds_locked + CT-balance gate

Replaces the Omen `funds_locked_in_markets` formula. Old: `(totalTraded - totalTradedSettled) / 1e18` — an agent-wide cash-flow net that counts realised losses as locked until Realitio finalisation (months out). New: `Σ remaining_cost(buy) for buy in held, non-LOSING positions`.

The "held" set comes from the CT subgraph's `userPositions` where `balance > 0` — added after a live-test discovery (see [Live verification](#live-verification) below) where the formula reported ~110 wxDAI locked on a safe whose CT shares were mostly already burned via `CT.redeemPositions`. Without the gate, redeemed winning positions double-count because the bet history is immutable. With the gate, redeemed / transferred / never-held positions correctly drop out.

`compute_funds_locked_from_bets` accepts `held_keys: Optional[set[(condition_id, outcome_index)]] = None`. `APTQueryingBehaviour._fetch_ct_held_position_keys` paginates the CT subgraph and **returns `Optional[set]` (`None` on error)** — the consumer treats `None` as "no gate, fall back to un-gated sum" so a transient CT-subgraph hiccup no longer collapses the FE-visible scalar to a phantom `0.0`.

Cross-skill snapshot hook in `PostOmenWithdrawBehaviour._snapshot_funds_locked`: fetches the bet history + CT-held set via the same subgraph paths the perf-summary round uses, runs the shared formula, writes the rounded result via `AgentPerformanceSummarySharedState.update_funds_locked_in_markets`. Skip-with-warning when `bets=[]` (indexer lag) so the snapshot doesn't write a phantom zero on the post-sweep round where the agent definitively has buy history. Failure is non-fatal.

## Hybrid status rule

Mirror of Polystrat's tiered classification in `_get_prediction_status`:

1. INVALID (finalized, not arbitrating) — beats everything.
2. **Fully exited via sells** (`remaining_shares <= SHARES_EPSILON_OMEN`): classify by realised PnL sign. Critical because a sell-at-profit on a market that later resolves AGAINST the bet was previously tagged LOST — now correctly WON.
3. Resolution-based fallback (unresolved / arbitrating / unfinalized → PENDING; won-but-unredeemed → PENDING; otherwise WON/LOST).

Legacy callers (raw subgraph bets, no FIFO state) skip tier 2 via the `original_shares > 0` gate; pre-rewrite behaviour preserved.

## Auth posture

`HTTP_SERVER_HOST` default flipped from `0.0.0.0` → `127.0.0.1` in both `services/trader_pearl/service.yaml` and `services/polymarket_trader/service.yaml`. The unauthenticated `POST /api/v1/withdrawal` route is no longer reachable from the public Propel ingress out of the box. Operators who need remote management explicitly opt in via the env var override. Defence-in-depth (operator-token header / EOA-signed challenge) is a natural follow-up; not in this PR.

## Live verification

Boot-tested against a live Gnosis-mainnet safe (`0x5461768d…300Fb80`). The agent armed, ran one sweep end-to-end, settled `0x38f4bc55…145ed86`, persisted fills/state, and triggered the cross-skill snapshot — all logged with the `omen withdrawal:` prefix. First run surfaced the funds_locked over-counting bug (110.51 wxDAI phantom-locked on a safe with mostly-redeemed positions); the CT-balance gate in Phase 3B above is the fix.

## Test plan

- [x] `parse_sell_events` (10 cases — happy paths + 5 malformed-log regressions: short topics / missing address / missing data / truncated payload / well-formed alongside malformed)
- [x] `get_pool_balances_via_ct` (2 cases)
- [x] `build_set_approval_for_all_tx` (2 cases)
- [x] `get_withdrawable_positions` (13 cases — full bucket table + 3 finalized-timestamp sentinel cases: `"0"` / `0` / malformed)
- [x] `OmenWithdrawRound.end_block` event switch (4 cases)
- [x] `inflate_for_slippage` (round-up invariant + range guard + asymmetry + **wei-scale integer math regression**)
- [x] Sizing-loop diagnostics (3 cases — calcSellAmount revert / halve-to-zero / max-attempts exhausted, each with its own distinct error message)
- [x] `OmenWithdrawalPayload` / `PostOmenWithdrawalPayload` constructors
- [x] HTTP 200 on Omenstrat (replaces 501 tests)
- [x] `OmenWithdrawalStore` (22 cases — full I/O surface + planned-FPMM allowlist roundtrip)
- [x] `parse_sell_events` planned-FPMM filter (4 cases — kept / dropped+warn / case-insensitive / missing-allowlist fallthrough)
- [x] `allocate_fifo` (12 cases — single buy, sell consumption, FIFO order, group separation, orphan sell, sign-disagreement, missing fpmm.id, zero-share buy, per-FPMM `participant_remaining_cost` scoping, same-FPMM-multi-outcome aggregation, module-level callable, wrapper-parity)
- [x] Sell-aware PnL regressions (sell-of-losing → loss, partial-sell winner, **multi-FPMM invalid-refund per-FPMM scoping**)
- [x] `_compute_omen_funds_locked` (10 cases — including "redeemed winning excluded" + **None-vs-set() fall-back**)
- [x] `_fetch_ct_held_position_keys` direct (4 cases — missing context, request failure, empty user, populated)
- [x] `_snapshot_funds_locked` cross-skill hook (5 cases — including indexer-lag skip + None-fall-through)
- [x] `_build_market_context` (10 cases — including **wei→wxDAI normalisation of `totalPayout` / `totalTraded`**)
- [x] `_format_predictions` (5 cases — including **DESC-by-`blockTimestamp` output ordering**)
- [x] `_size_and_build_position` approval-dedupe (1 case)
- [x] Hybrid status rule (10 cases)
- [x] `check-abciapp-specs` + `check-abci-docstrings` + `check-handlers` + `check-hash` + `bandit` + `black-check` + `isort-check` + `flake8` + `mypy` + `darglint` + `pylint` all clean
- [x] **Live Gnosis mainnet run** — armed via curl, sweep settled at `0x38f4bc55…145ed86`, snapshot fired, surfaced the redemption-burn bug (now fixed).
- [ ] 3 real Gnosis receipt fixtures for `parse_sell_events` (vs the synthetic receipts used today). Defensive against a topic0 / payload-layout regression.
- [ ] ~12 behaviour-level integration tests for `OmenWithdrawBehaviour` / `PostOmenWithdrawBehaviour.async_act` end-to-end runs.

3 207 tests pass across `decision_maker_abci`, `agent_performance_summary_abci`, `market_manager_abci`, `chatui_abci`, `tx_settlement_multiplexer_abci`, and both contracts.

## Known deviations from spec

| | |
|---|---|
| Per-position fill/error attribution on tx failure | Currently records one top-level error on `tx.status=0` / `ExecutionFailure`. Per-requested-position errors would need to persist the planned-positions list across rounds beyond the existing allowlist. |
| `token_id` on fills | Currently empty-string sentinel (`TOP_LEVEL_ERROR_TOKEN_ID`, justified inline). Cosmetic for the venue-agnostic FE component. |
| Position-id derivation on-chain via `getCollectionId` | Empirical scan didn't validate the off-chain derivation against Gnosis CT v1.4.0's BN256 curve math; mirrored the existing `get_user_holdings` pattern instead. |
| `WITHDRAWAL_SLIPPAGE` defaults `0.01` hardcoded | Spec says default tracks `params.slippage`. Same value today; easy fix. |
| `PostOmenWithdrawalPayload` dropped unused `event` field | Round emits hardcoded `done_event` regardless; field would never be read. Matches the `WithdrawalPayload(VotingPayload)` shape Polystrat already uses. |

## Review fixes applied

Two rounds of post-draft review work:

| Commit | Scope |
|---|---|
| `3f79888c` | First round — 17 issues from @bennyjo, @OjusWiZard, @jmoreira-valory, @LOCKhart07 (held_keys semantics, partial-list strand, "0" sentinel, layered parse_sell_events guards, _fetch_status FAIL, post-handler null-check, indexer-lag skip, OmenWithdrawalStore extraction, allocate_fifo promotion, per-FPMM denominator, planned-FPMM allowlist, sizing-loop diagnostics, rotting-ref cleanup, nosec justifications, PostOmenWithdrawRound timeout escape, build_*_tx return-type normalisation, wei-scale int math). |
| `b47d5b6b` | Second round — 4 issues from @DIvyaNautiyal07 (`totalPayout` wei-vs-wxDAI unit fix, `inflate_for_slippage` int math, `_format_predictions` DESC-by-ts ordering, `setApprovalForAll` per-FPMM dedupe). |
| `82fd0fde` | Lint pass (black / isort / flake8 / mypy / darglint clean). |
| `671ea8d8` | Auth: `HTTP_SERVER_HOST` default → `127.0.0.1` (addresses @jmoreira's #942 re-raise). |

## LoC breakdown

`git diff main..HEAD --numstat`, bucketed:

| Bucket | Files | + | − | Net |
|---|--:|--:|--:|--:|
| **Production code** (`.py` outside `tests/`) | 24 | 2 475 | 129 | **+2 346** |
| **Tests** (`.py` under `tests/`) | 13 | 3 277 | 79 | **+3 198** |
| **Other** — FE UI bundle (`index-…js`/`index.html`) + `.gitleaksignore` | 3 | 242 | 189 | +53 |
| **FSM specs** (`fsm_specification.yaml`) | 3 | 25 | 0 | +25 |
| **Hashes / YAML configs** (`packages.json`, `skill.yaml`, `service.yaml`, `aea-config.yaml`, `contract.yaml`) | 13 | 98 | 78 | +20 |
| TOTAL | 56 | 6 117 | 475 | **+5 642** |

Production code: 3 new files (`omen_withdraw.py`, `post_omen_withdraw.py`, `omen_withdrawal_store.py`) + 21 modified.

Test/code ratio: roughly 1.4 lines of test per 1 line of production code.


packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
```diff
@@ -232,6 +232,108 @@ def _fetch_mech_sender(
...
+        """
+        held: "set[tuple[str, int]]" = set()
+        try:
+            ct_subgraph = self.context.conditional_tokens_subgraph
+        except AttributeError:
+            self.context.logger.warning(
+                "funds_locked: CT subgraph context unavailable; "
+                "skipping CT-balance gate"
+            )
+            return held
```
bennyjo: When the CT subgraph is briefly unreachable, this returns an empty set. The caller treats that as "user holds nothing", so `funds_locked_in_markets` writes 0 to shared state.

User impact: during any CT subgraph hiccup, the dashboard's "Funds locked in markets" card flips to **0 wxDAI**, suggesting all funds are recovered. The post-sweep snapshot runs the same code, so this is most likely to fire right when users open the dashboard to check on a sweep.

Fix: `return None` instead — the caller already falls back cleanly on None. Same shape on the inner `if not response or not isinstance(response, dict): break` path.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -825,6 +946,175 @@ def _format_predictions(
...
+                            shares_consumed,
+                            unattributed_wxdai,
+                        )
+
+            output.extend(buys_in_group)
+
+        # participant_remaining_cost — sum across all enriched buys for the
+        # invalid-market refund pro-rate. Stored in base units (wei) to
+        # match the rest of the FIFO state.
+        participant_remaining_cost = sum(
```
bennyjo: This sums remaining cost across every FPMM the user holds. But `_calculate_bet_net_profit` uses it as the denominator against a per-FPMM `total_payout` on the invalid-market refund path.

User impact: a user with open positions across N markets who hits one invalid-market resolution sees a phantom loss of `total_payout × (N−1)/N` on that bet, with the same skew on the aggregate PnL number. No funds at risk — display only — but the bug shows up specifically on the multi-market, partial-exit shape the withdrawal feature is built for.

Fix: scope the sum per-FPMM inside the existing group loop.

bennyjo (commented):
Two correctness items on the new funds_locked path — both display-only, no on-chain effect, but worth resolving before merge.

packages/valory/skills/market_manager_abci/graph_tooling/requests.py
```diff
@@ -388,6 +391,54 @@ def fetch_trades(
...
+            )
+            res = current_subgraph.process_response(res_raw)
+            rows = yield from self._handle_response(
+                current_subgraph,
+                res,
+                res_context="withdrawal_creator_fpmms",
+            )
+            if res is None:
+                self.context.logger.error("Failed to process withdrawal creator FPMMs.")
+                return None
```
OjusWiZard, line 442: **HIGH · `_fetch_status` never set to a terminal state** — this method assigns `IN_PROGRESS` on entry but every other return path leaves it `IN_PROGRESS`. Every other paginated method on `QueryingBehaviour` transitions to `SUCCESS`/`FAIL` on exit. Any future caller that gates on `_fetch_status` will see a hung state.

Also applies to: callers should treat the `None` return as a fatal error (good), but the status flag won't reflect it.

packages/valory/skills/market_manager_abci/graph_tooling/requests.py
```diff
@@ -388,6 +391,54 @@ def fetch_trades(
...
+                content=to_content(query),
+                **current_subgraph.get_spec(),
+            )
+            res = current_subgraph.process_response(res_raw)
+            rows = yield from self._handle_response(
+                current_subgraph,
+                res,
+                res_context="withdrawal_creator_fpmms",
+            )
+            if res is None:
```
OjusWiZard: **MEDIUM · null-check on `res` is downstream of the `_handle_response` call that consumes it.** Order is: `rows = yield from self._handle_response(subgraph, res, ...)` → `if res is None`. If `res` is `None`, `_handle_response` has already been invoked on `None`. The check should be on `rows` (the post-handler result), mirroring `fetch_trades` and `fetch_bets`. As-is, `len(rows)` on the next iteration of the surrounding `while` would `TypeError` if `_handle_response` ever returns `None` while `res` is truthy.

packages/valory/skills/market_manager_abci/graph_tooling/utils.py
```diff
@@ -194,3 +195,116 @@ def filter_claimed_conditions(
...
+            index_set = int(index_sets_raw[0])
+        except (TypeError, ValueError):
+            continue
+        if not _is_power_of_two(index_set):
+            continue
+        fpmm_meta: Optional[Dict[str, Any]] = fpmm_by_condition.get(condition_id)
+        if fpmm_meta is None:
+            # CT position with no FPMM in the safe's trade history
+            continue
+        if fpmm_meta.get("answerFinalizedTimestamp") is not None:
```
OjusWiZard: **HIGH (verify) · is the omen subgraph really returning `null` here, or the string `"0"`?**

If the subgraph returns `"0"` for an unfinalized market (as it does in several other Olas integrations), `is not None` is true and every OPEN position is silently excluded from the sweep. Worth a quick sanity-check against a live Gnosis subgraph response — if `"0"` is the actual unfinalized sentinel, this gate strands user funds.

Compare against `next_status` / `parse_timestamp` semantics elsewhere in `market_manager_abci`, which explicitly normalise `"0"` to unfinalized.

packages/valory/contracts/market_maker/contract.py
```diff
@@ -186,3 +201,98 @@ def get_sell_data(
...
+        """
+        events: List[Dict[str, Any]] = []
+        for log in receipt.get("logs", []) or []:
+            topics = log.get("topics", []) or []
+            if not topics or HexBytes(topics[0]) != FPMM_SELL_TOPIC0:
+                continue
+            seller_padded = HexBytes(topics[1]).hex()
+            seller = "0x" + seller_padded[-_ADDRESS_HEX_LEN:]
+            outcome_index = int(HexBytes(topics[2]).hex(), 16)
+            data = HexBytes(log["data"])
```
OjusWiZard: **HIGH · unguarded `log["data"]` (and `topics[1]`/`topics[2]` above) — diagnostic crater on malformed receipts.**

Every other field in the loop uses `.get()`. A log that matches topic0 but is missing `"data"`, or that arrives from a non-conformant RPC with only one topic, will raise `KeyError`/`IndexError`. The framework wraps the exception into a generic non-`STATE` performative; the caller in `post_omen_withdraw.py::_parse_sell_events` records `"parse_sell_events dispatch failed"` with no signal about which log was the culprit.

Suggest: `log.get("data", "")` plus a `len(topics) < 3` guard. Log the malformed entry at WARNING and `continue` rather than raising. The defensive `or []` on `log.get("topics", [])` is already in place — finish the pattern.

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
```diff
@@ -17,34 +17,505 @@
...
+            if n_estimate is None:
+                self._record_error(
+                    position,
+                    "calcSellAmount reverted: see logs",
+                )
+                return None
+            if n_estimate <= headroom_cap:
+                break
+            return_amount //= 2
+            if return_amount == 0:
```
OjusWiZard, line 358: **MEDIUM · `break` on `return_amount == 0` bypasses the `for/else` correctness path; the post-loop guard at L302 is the only thing preventing a silent drop.**

In Python, `for…else` runs the `else` only when the loop is *not* broken out of. The `break` at L293 means the `else` recording "could not be sized" is skipped — control falls to L302 which fires the same error message for two distinct conditions (`n_estimate <= 0` from a revert *and* `return_amount == 0` from halving). Two issues:

1. **Future-refactor footgun.** If someone removes the redundant L302 guard thinking it's dead code, the halve-to-zero path becomes a silent drop with no error record. The `break` should be `return self._record_error(...)`.
2. **Operator diagnostics.** "could not be sized" doesn't distinguish "halved to zero" from "max attempts exhausted" from "calcSellAmount reverted". The three failure modes have very different root causes (pool depth vs slippage param vs market state).

Consider one of:
```python
if return_amount == 0:
    self._record_error(position, "return_amount halved to zero before sizing converged")
    return None
```
or pull the loop into a helper that returns a tagged result.

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
```diff
@@ -17,34 +17,505 @@
...
+)
+from packages.valory.skills.market_manager_abci.graph_tooling.utils import (
+    WithdrawablePosition,
+    get_withdrawable_positions,
+)
+
+# Hardcoded per spec §9.3 — every Olas-touched Omen FPMM uses wxDAI as
+# collateral. Verified across 7,203 distinct FPMMs in the empirical scan.
+WXDAI = "0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d"
+TOP_LEVEL_ERROR_TOKEN_ID = ""  # nosec B105
```
OjusWiZard: **LOW · `nosec B105` lacks a justification comment.**

Bandit B105 is "hardcoded password string" — a future reader (or security reviewer running a sweep) seeing a bare `# nosec B105` on a module-level string constant can't distinguish a justified suppression from a silenced real finding. The fill-record `nosec` at L324 in `post_omen_withdraw.py` does have a block comment above it explaining it's a placeholder — replicate that here, e.g.:

```python
# Empty string sentinel for top-level (multi-position) errors where no
# single token_id applies. Not a credential — # nosec B105.
TOP_LEVEL_ERROR_TOKEN_ID = ""  # nosec B105
```

Same applies to the identical declaration in `post_omen_withdraw.py` L73.

packages/valory/skills/decision_maker_abci/behaviours/post_omen_withdraw.py
```diff
@@ -0,0 +1,354 @@
...
+            trader_agent = yield from self._fetch_trader_agent_performance(
+                safe_address.lower()
+            )
+            if not trader_agent:
+                self.context.logger.warning(
+                    "omen withdrawal: skipping funds_locked snapshot "
+                    "(trader_agent fetch returned no data)"
+                )
+                return
+            bets = trader_agent.get("bets") or []
```
OjusWiZard, line 299: **HIGH · phantom `0.0` written to `funds_locked_in_markets` on indexer lag.**

The gate at L242 catches `not trader_agent` (no subgraph data) but not `trader_agent exists but bets=[]`. Immediately post-settlement, the trader-agent subgraph commonly returns a valid-shaped response with `bets=[]` while the indexer catches up. Current code falls through to `compute_funds_locked_from_bets([], …) → 0.0` and writes that as the new `funds_locked_in_markets`.

The next normal perf-summary round corrects it, but for several minutes the FE shows a phantom zero — the opposite of "non-fatal, skip on failure". Suggest:

```python
bets = trader_agent.get("bets") or []
if not bets:
    self.context.logger.warning(
        "omen withdrawal: skipping funds_locked snapshot "
        "(trader_agent has no bets yet — likely indexer lag)"
    )
    return
```

This preserves the documented semantic ("skip on failure, next round fixes it") instead of writing a wrong scalar.

packages/valory/skills/decision_maker_abci/behaviours/post_omen_withdraw.py
```diff
@@ -0,0 +1,354 @@
...
+
+# Safe v1.x event topics — keccak("ExecutionFailure(bytes32,uint256)") /
+# ("ExecutionSuccess(bytes32,uint256)"). The outer Safe ``execTransaction``
+# returns true (status=1) regardless of whether the inner multisend
+# reverted; the only signal of inner failure on a status=1 tx is the
+# ``ExecutionFailure`` topic.
+EXECUTION_FAILURE_TOPIC0 = HexBytes(
+    "0x23428b18acfb3ea64b08dc0c1d296ea9c09702c09083ca5272e64d115b687d23"
+)
+TOP_LEVEL_ERROR_TOKEN_ID = ""  # nosec B105
```
OjusWiZard: **HIGH · duplicated disk-store machinery across this file and `omen_withdraw.py`** (already drifting).

`TOP_LEVEL_ERROR_TOKEN_ID = ""`, plus the six methods `_store_path`/`_read_store`/`_write_store`/`_set_state`/`_store_has_errors`/`_record_top_level_error` are duplicated verbatim across both behaviours. The PR comments themselves say "mirror OmenWithdrawBehaviour", which is a maintenance smell, not a justification.

**Drift has already started:** `_record_top_level_error` in `omen_withdraw.py` accepts `op_name: str` and writes an `"fpmm"` field, while the version here accepts `reason: str` only. Any future schema change must be applied to both — and the next reviewer won't catch a missed mirror.

Suggest extracting an `OmenWithdrawalStore` helper (frozen dataclass + module-level functions, or a small mixin) in a shared module. Also consider: `_record_fill`/`_record_error` do a full `json.load`+`json.dump` per item — O(N) disk I/O for a sweep of N positions. Accumulate then write once.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -115,6 +123,101 @@ def parse_timestamp(value: Any) -> Optional[int]:
...
+        (legacy callers / fallback on subgraph error), no gate is
+        applied — same behaviour as the initial release.
+    :return: wxDAI total of remaining cost on currently-held,
+        non-LOSING positions; 0.0 for an empty bet list.
+    """
+    if not bets:
+        return 0.0
+
+    fetcher = PredictionsFetcher(context, logger)
+    enriched_buys = fetcher._allocate_fifo(bets)
```
OjusWiZard: **HIGH · public module function reaches into a private method via a throwaway instance.**

`fetcher = PredictionsFetcher(context, logger); enriched_buys = fetcher._allocate_fifo(bets)` instantiates the class only to call its private (`_`-prefixed) method. Two problems:

1. Public API depends on private implementation detail — the leading underscore now lies.
2. `PostOmenWithdrawBehaviour` already inherits `APTQueryingBehaviour`, so it has a real `PredictionsFetcher` in scope; this constructs a second one with no caching.

Promote `_allocate_fifo` to a module-level pure function (it appears not to depend on instance state beyond `self.logger`), or make `compute_funds_locked_from_bets` a `@classmethod`/method on `PredictionsFetcher`. The current shape is the worst of both worlds.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -115,6 +123,101 @@ def parse_timestamp(value: Any) -> Optional[int]:
...
+            outcome_index = buy.get("outcomeIndex")
+            if (
+                correct is not None
+                and outcome_index is not None
+                and int(outcome_index) != correct
+            ):
+                # Resolved-and-losing — shares are worthless.
+                continue
+
+        if held_keys is not None:
```
OjusWiZard, line 388: **HIGH · `held_keys=None` vs `held_keys=set()` have opposite semantics, but the fetcher returns an empty set on error.**

Docstring at L167–172 says `None` means "no gate" (fallback). The CT fetcher's docstring (`requests.py::_fetch_ct_held_position_keys` L250–253) says: *"Returns an empty set on subgraph error; callers treat that as 'no gate' rather than 'everything is unheld' — preserves the pre-fix behaviour as a fallback rather than zeroing out funds_locked spuriously."*

**But this code path treats `set()` as "everything unheld"** — `if held_keys is not None:` enters the branch with an empty set, and every position fails the `key not in held_keys` check. The intended fallback path is unreachable from the post-withdrawal snapshot caller. A CT subgraph outage will silently zero `funds_locked_in_markets` until the next normal round.

Fix one of:
- `if held_keys:` (treats empty as no gate — matches fetcher's documented intent)
- Have `_fetch_ct_held_position_keys` return `None` on error (matches the `held_keys=None` semantic)
- Add an explicit `apply_ct_gate: bool` parameter alongside `held_keys: Set[...]`

mypy can't catch this because both `None` and `set()` satisfy `Optional[Set[...]]`. The fact that the two callsites have asymmetric semantics for the same parameter is the core design issue.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -918,79 +1221,179 @@ def _calculate_bet_net_profit(
...
+        participant's total remaining cost. Legacy callers that pass a raw
+        subgraph row (no FIFO fields) get treated as "never sold" via
+        :meth:`_fifo_state`'s defaults, preserving the pre-rewrite formula.
+
+        ``bet_amount`` is kept in the signature for backward compatibility
+        but no longer used — all cost-basis math now flows through the
+        FIFO state.
+
+        Spec §8.1 calls out three prior bug sites this method now fixes:
+          - Line 954 invalid-market refund used the signed raw amount as
```
OjusWiZard: **MEDIUM · hardcoded "Line 867 / 954 / 971" references in docstring will rot.**

These are spec-document line numbers, not line numbers in this Python file (L867 here is `_build_market_context`, L954 is inside `_allocate_fifo`, L971 is mid-allocator — none match the bug descriptions). A future reader will jump to L954 in *this* file expecting to find the bug site and be confused. Also, the spec document (`WITHDRAWAL_OMEN_IMPL_SPEC.md`) is explicitly not checked in, so no jumping is possible anyway.

More broadly: this PR has spec-section references (`spec §6.3`, `spec §8.1`, `spec §7.3 / §10.13.3`, etc.) scattered across `omen_withdraw.py`, `post_omen_withdraw.py`, `predictions_helper.py`, `requests.py`, and `contract.py`. With the spec not in the repo, all of these become unresolvable on future read — comment rot at scale.

Suggest: replace spec references with the actual rationale inline ("The pre-FIFO denominator used the signed raw amount; a sell-side row made it negative, surfacing as `+profit` instead of a loss") so the WHY survives independent of the external doc. Same applies to the PR-number references ("PR #948 review round 2", "PR #952's first live run", "PR #903 review comment #1") — those belong in commit messages, not docstrings.

OjusWiZard (commented):
Comprehensive review across code-quality, type-design, test-coverage, silent-failure, comment, and simplification lenses. Posted inline notes on the highest-signal findings. Full summary is in the PR thread.

**One CRITICAL finding I could not pin inline** (the lines are unchanged context, not in any diff hunk) — calling it out here:

### CRITICAL · `fetch_user_positions` silently returns the partial list on subgraph error

`packages/valory/skills/market_manager_abci/graph_tooling/requests.py` L470–473:

```python
if res is None:
    self.context.logger.error("Failed to process all positions.")
    return all_positions   # ← returns the partial accumulator
```

This is the sister method of the newly-added `fetch_withdrawal_creator_fpmms` and the bug is pre-existing — but the new Omen sweep is the first caller where a partial list silently strands user funds. `OmenWithdrawBehaviour._build_payload` treats a non-`None` return as success, so a mid-page CT subgraph error during pagination causes the sweep to complete with positions on un-fetched pages unsold. The sister method correctly returns `None` on the same condition. Fix: return `None` here so `_with_top_level_retry` can do its job and surface the failure.

---

Nothing in this review is a blanket merge block — just request changes around C1 (above), H1 (verify the `"0"` semantics), H2, H3, the held_keys None-vs-empty-set design (H5), and the test-coverage gap acknowledged in the PR description. The Polymarket-side test suite is exemplary; the live verification with `0x38f4bc55…` is excellent evidence.

Many positive observations: the regression tests for sell-of-losing/partial-sell-winner are real (would fail without the fix); `TestComputeOmenFundsLocked` correctly exercises the CT-balance gate; the `TestOmenWithdrawRoundEventSwitch` stubbing approach is right.

packages/valory/contracts/market_maker/contract.py
```diff
@@ -186,3 +201,98 @@ def get_sell_data(
...
+            return_amount = int.from_bytes(data[:_WORD_BYTES], "big")
+            fee_amount = int.from_bytes(data[_WORD_BYTES : 2 * _WORD_BYTES], "big")
+            outcome_tokens_sold = int.from_bytes(
+                data[2 * _WORD_BYTES : 3 * _WORD_BYTES], "big"
+            )
+            events.append(
+                {
+                    "seller": ledger_api.api.to_checksum_address(seller),
+                    "fpmm": ledger_api.api.to_checksum_address(log["address"]),
```
jmoreira-valory: **`parse_sell_events` doesn't validate `log["address"]` against an expected FPMM set — audit-trail mis-attribution risk.**

The loop accepts every log whose topic0 matches `FPMM_SELL_TOPIC0` from any contract address, then uses `log["address"]` directly as the `fpmm` field in the returned event:

```python
for log in receipt.get("logs", []) or []:
    topics = log.get("topics", []) or []
    if not topics or HexBytes(topics[0]) != FPMM_SELL_TOPIC0:
        continue
    ...
    events.append({
        ...
        "fpmm": ledger_api.api.to_checksum_address(log["address"]),
        ...
    })
```

Real exploitation risk is low (Safe multisend is the only authorizing entity, and our planned multisend only calls our target FPMMs), but a non-target FPMM emitting an `FPMMSell` event in the same receipt — e.g., a separate FPMM invoked via some unrelated hook, a fee distributor, or a future cross-contract integration that lands in this tx — gets silently recorded as a fill in the operator-facing audit trail.

The fills/errors arrays at `${store_path}/chatui_param_store.json` are the operator's ground truth for *"what did the sweep actually do?"*. A fake fill record distorts both the UI display (`positions_sold` count, per-fill list) and any downstream reconciliation against the safe's pre-sweep CT balance.

Cheap fix — caller-side filter against the planned set. `PostOmenWithdrawBehaviour` knows which FPMMs the sweep targeted (the original multisend batches). Either pass that set into the parser:

```python
result = FixedProductMarketMakerContract.parse_sell_events(
    ledger_api=..., contract_address=..., receipt=receipt,
    expected_fpmms={b.to for b in self.planned_sell_batches},
)
```

…and filter inside, or post-filter at the call site:

```python
events = [e for e in raw_events if e["fpmm"] in planned_fpmms]
```

Restores the "only count fills we caused" invariant the audit trail should hold.

jmoreira-valory:
**Re-raising from #942 — unauthenticated `POST /api/v1/withdrawal` now extends to Omenstrat, with materially higher on-chain destructive radius than Polystrat.**

This PR drops the 501 short-circuit in `chatui_abci/handlers.py` (the 17-line deletion at the top of `_handle_post_withdrawal`), so the same unauthenticated POST endpoint I flagged on [#942 (discussion_r3202953480)](https://github.com/valory-xyz/trader/pull/942#discussion_r3202953480) now applies to Omen.

That comment was dismissed as *"Not a concern. We don't expect anyone to have access to this api"*, but `HTTP_SERVER_HOST` defaults to `0.0.0.0` per `services/trader_pearl/service.yaml:368` and the Propel ingress regex `{16hex}.agent.propel.autonolas.tech` (registered in the same handler's `hostname_regex`) makes the URL public-internet-reachable.

The Omen-specific case is materially more dangerous than Polystrat:

| | Polystrat (#942) | Omenstrat (this PR) |
|---|---|---|
| Sell mechanism | CLOB FAK orders | **Safe multisend** of `(setApprovalForAll, FPMM.sell)*N` |
| Partial fills | Possible mid-flight; can re-arm | Atomic — all-or-nothing per multisend |
| Pricing | Order-book best execution | AMM marginal price minus `withdrawal_slippage` |
| Default slippage tolerance | N/A (FAK) | **1%** (`withdrawal_slippage = 0.01`) |
| Default return-amount discount | N/A | **5%** (`withdrawal_return_buffer = 0.05`) |

On a thin FPMM, the sized sell can execute at 5–10% below spot — irrecoverable. A network-reachable adversary triggering one POST forces an atomic on-chain settlement of every held position at materially adverse prices.

Same mitigations as on #942 apply with higher urgency here:

1. **Cheapest**: default `HTTP_SERVER_HOST` to `127.0.0.1` in `services/trader_pearl/service.yaml` and `services/polymarket_trader/service.yaml`. Override via env if remote ops are genuinely needed.
2. Require a shared-secret header (`X-Operator-Token`) injected via env var.
3. EOA-signed challenge against the master wallet's owner key.

At minimum, this PR should not merge until the `service.yaml` binding for Omenstrat is reviewed for who can reach `:8716` in a Pearl deployment. The Polystrat blast-radius argument from #942 was already weak; the Omenstrat one is harder to defend.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
```diff
@@ -232,6 +232,109 @@ def _fetch_mech_sender(
...
```
LOCKhart07: **Blocking — subgraph error here silently writes 0.0 to `funds_locked_in_markets`.**

On `AttributeError` (or when `_fetch_from_subgraph` returns falsy on the first page → loop `break` at line ~313/336), this returns `set()`. The caller `compute_funds_locked_from_bets` checks `if held_keys is not None:` and then `if key not in held_keys: continue` — so an **empty** set gates every bet out, collapsing `funds_locked` to 0. That's the opposite of what the function's own docstring (lines 246-251) promises:

> Returns an empty set on subgraph error; callers treat that as "no gate" rather than > "everything is unheld" — preserves the pre-fix behaviour as a fallback rather than > zeroing out funds_locked spuriously.

Every transient CT-subgraph hiccup on the post-withdrawal snapshot hook or normal perf-summary round will write a spurious `0` to the FE-visible scalar.

Note: `test_held_set_none_falls_back_to_no_gate` in `test_behaviours.py:5574` bakes the buggy behaviour in (asserts `== 0.0`) — its name and docstring don't match what's actually asserted.

**Fix:** return `Optional[set]` — `None` on error, set on success (including legitimately empty). `compute_funds_locked_from_bets` already handles `None` as "no gate"; just propagate, and update the callers in `aps behaviours.py` and `post_omen_withdraw.py` to either pass `None` through or skip the write when fetch failed.

packages/valory/skills/market_manager_abci/graph_tooling/utils.py
LOCKhart07: **Non-blocking — `"0"` sentinel slips through as "resolved".**

Some Omen subgraph indexers emit string `"0"` for unset `answerFinalizedTimestamp` — already known in this codebase (see `parse_timestamp` in `predictions_helper.py` which added the `"0"`/`0` handling per PR #903 review). Here `is not None` lets `"0"` through, so truly-unfinalized OPEN positions get filtered out as resolved and silently skipped by the sweep.

Suggested: reuse `parse_timestamp` semantics — treat `None`, `"0"`, `0`, and malformed values as unfinalized.

packages/valory/skills/decision_maker_abci/rounds.py
```diff
@@ -545,11 +559,25 @@ class DecisionMakerAbciApp(AbciApp[Event]):
...
+            # DONE / MOCK_TX inherited from `TxPreparationRound` and silenced
+            # by the `end_block` override (which switches `DONE` to the
+            # payload-carried `event` field). Routed defensively to the
+            # terminal so the FSM linter sees them; never actually emitted.
+            Event.DONE: WithdrawalIdleRound,
+            Event.MOCK_TX: WithdrawalIdleRound,
         },
+        PostOmenWithdrawRound: {
+            Event.WITHDRAWAL_DONE: WithdrawalIdleRound,
+            Event.WITHDRAWAL_ROUND_TIMEOUT: PostOmenWithdrawRound,
```
LOCKhart07: **Non-blocking — `PostOmenWithdrawRound` timeout self-loops with no escape.**

If the receipt-parse behaviour persistently times out (1800s), the agent is stuck indefinitely in this round. Other tx-settled rounds escape on timeout — compare:

- `OmenWithdrawRound` (line 564): `WITHDRAWAL_ROUND_TIMEOUT: WithdrawalIdleRound`
- `PolymarketRedeemRound` (line 515): `REDEEM_ROUND_TIMEOUT: FinishedDecisionMakerRound`
- `RedeemRound` (line 498): `REDEEM_ROUND_TIMEOUT: FinishedWithoutRedeemingRound`

Receipt parsing is deterministic — retrying the same input won't unblock a real bug. Suggest `WITHDRAWAL_ROUND_TIMEOUT: WithdrawalIdleRound` for consistency with the build round upstream.

packages/valory/contracts/conditional_tokens/contract.py
```diff
@@ -489,6 +489,35 @@ def get_balance_of(
...
+        :param operator: the address being approved (typically an FPMM)
+        :param approved: whether to grant (True) or revoke (False) approval
+        :return: ``{"data": encoded_calldata_hex}``
+        """
+        contract_instance = cls.get_instance(ledger_api, contract_address)
+        data = contract_instance.encode_abi(
+            abi_element_identifier="setApprovalForAll",
+            args=[ledger_api.api.to_checksum_address(operator), approved],
+        )
+        return dict(data=data)
```
LOCKhart07: **Non-blocking — inconsistent with the rest of the file.**

Every other `build_*_tx` method in this file returns `bytes.fromhex(data[2:])`:
- `build_redeem_positions_tx` (line 231)
- `get_prepare_condition_tx_data` (line 328)
- `build_merge_positions_tx` (line 547)

This one returns the raw `0x…` hex string verbatim. Works in practice because the caller in `omen_withdraw.py:_build_set_approval_batch` wraps in `HexBytes(data)` which accepts both, but the inconsistency is unintentional. Either normalise this one to bytes or document the divergence.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -825,6 +946,176 @@ def _format_predictions(
...
+                        if take <= 0:
+                            open_buys.popleft()
+                            continue
+                        share_ratio = take / (-row_shares)
+                        cost_ratio = take / head["original_shares"]
+                        head["allocated_proceeds"] += proceeds_total * share_ratio
+                        head["allocated_cost"] += head["original_cost"] * cost_ratio
+                        head["remaining_shares"] -= take
+                        shares_consumed -= take
+                        if head["remaining_shares"] <= 0:
```
LOCKhart07: **Non-blocking — float drift on wei-scale share math.**

`row_amount = float(row.get("amount", 0))` (line 1006) and `row_shares = float(...)` (line 1007) convert wei-scaled subgraph strings (1e18-scaled) to Python `float`. IEEE-754 doubles lose precision past 2^53 (~9e15 wei = 0.009 wxDAI), so:

- `head["remaining_shares"] -= take` accumulates drift across multiple sells.
- This `<= 0` comparison can leave a tiny positive residual (e.g. 1e-9) in the head,   blocking the queue and under-allocating subsequent sells.

The withdrawal sweep itself reads CT balances as ints from the subgraph, so the sweep path is unaffected. Impact is limited to perf-summary display values and `compute_funds_locked_from_bets` — small in practice (current Olas bet sizes < 10 wxDAI), but it's a slow-bleed correctness gap.

Suggested: keep ints throughout the allocator — `int(row.get("amount", 0))` and compute ratios as `Fraction` or integer-scaled.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
DIvyaNautiyal07: **Blocking — falsy-signal semantics.** Returning `set()` on `AttributeError` (and again at line 336 on subgraph-fetch failure) is indistinguishable from a legitimately-empty held set. The consumer in `predictions_helper.py:197` checks `if held_keys is not None:`, so an empty set passes that gate and then `if key not in held_keys:` drops every position → `funds_locked_in_markets` reported as `0.0` on a transient CT-subgraph blip.

The docstring claims this 'preserves the pre-fix behaviour as a fallback rather than zeroing out funds_locked spuriously', but the test `test_held_set_none_falls_back_to_no_gate` (test_behaviours.py:5574) actually asserts `result == 0.0` for this case — confirming the implementation does the opposite of what the docstring claims, and would happily ship a regression of the same shape as the original 110-wxDAI bug, just flipped.

Fix: return `None` on error (distinct from empty set), or change the consumer to `if held_keys:` (truthy), or use a separate fetch-succeeded sentinel.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
DIvyaNautiyal07, line 388: Paired with the issue on `_fetch_ct_held_position_keys`: this gate fires for both 'subgraph errored, fall back to no-gate' (intent per the fetcher's docstring) and 'subgraph succeeded, safe legitimately holds nothing' (everything redeemed). Treat the two cases distinctly — `None` from the fetcher = 'no gate'; `set()` = 'all positions excluded'.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -918,79 +1221,179 @@ def _calculate_bet_net_profit(
...
         if current_answer == INVALID_ANSWER_HEX:
-            if total_payout == 0 or total_traded == 0:
-                return 0.0, None
-            refund_share = total_payout * (bet_amount / total_traded)
-            return refund_share - bet_amount, refund_share
+            if total_payout == 0 or participant_remaining_cost <= 0:
+                return realized_pnl, (
+                    allocated_proceeds if allocated_proceeds > 0 else None
+                )
+            refund_share = total_payout * (remaining_cost / participant_remaining_cost)
```
DIvyaNautiyal07, line 1388: **Unit regression.** `total_payout` is stored in raw wei at line 1150 (note line 1441 in `_get_prediction_status` and line 443 in the specific-market path both divide by `WEI_TO_NATIVE`, but `_build_market_context` does not). The pre-PR formula `total_payout * (bet_amount / total_traded)` had cancelling units: `wei * (wxDAI / wei) = wxDAI` ✓. The new formula uses `participant_remaining_cost` (wxDAI, divided at 1270) as the denominator, so `wei * (wxDAI / wxDAI) = wei` — the refund_share comes out ~1e18× too large.

Tests don't catch this because the fixtures inject `total_payout = 2.0` (small wxDAI-scale float) instead of the wei-scale `str(2 * WEI_TO_NATIVE)` production stores. Fix: either divide `total_payout` by `WEI_TO_NATIVE` at line 1150 (and update consumers), or multiply `participant_remaining_cost` back to wei here. (Same shape of bug exists in the winning path pre-PR — flagging the newly-introduced invalid-path regression here only.)

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
```diff
@@ -17,34 +17,505 @@
...
+        sellable.sort(key=lambda p: (p.fpmm_address.lower(), p.outcome_index))
+        self.context.logger.info(
+            f"omen withdrawal: discovered {len(sellable)} sellable position(s)"
+        )
+
+        batches: List[MultisendBatch] = []
+        for position in sellable:
+            position_batches = yield from self._size_and_build_position(position)
+            if position_batches is not None:
+                batches.extend(position_batches)
```
DIvyaNautiyal07, line 200: Each position emits its own `(setApprovalForAll, sell)` pair. If a safe holds two outcomes on the same FPMM, two identical `setApprovalForAll(FPMM, true)` calls get included in the multisend (operator is the FPMM, so per-FPMM not per-outcome). Dedupe by `fpmm_address` to drop the redundant approves. Minor gas waste, not a correctness bug.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -807,15 +921,22 @@ def _format_predictions(
...
         :param status_filter: Optional status filter
         :return: List of formatted prediction dictionaries
         """
-        market_ctx = self._build_market_context(bets)
+        enriched_buys = self._allocate_fifo(bets)
+        market_ctx = self._build_market_context(enriched_buys)
 
         items: List[Dict] = []
-        for bet in bets:  # already ordered desc by timestamp in the query
+        for bet in enriched_buys:  # ordered chronologically by FIFO output
```
DIvyaNautiyal07: Comment claims 'ordered chronologically by FIFO output', but the actual order is mixed: within each `(fpmm, outcome)` group the FIFO allocator sorts ASC by `blockTimestamp`, while between groups iteration follows `groups.items()` insertion order = first-occurrence in the input (which arrived DESC by ts). Pre-PR this loop returned strict DESC-by-ts. If the FE relies on newest-first, this is a display regression. Consider an explicit re-sort on the output before returning, or document that the FE re-sorts.

packages/valory/contracts/conditional_tokens/contract.py
DIvyaNautiyal07: Inconsistent with the sibling `build_redeem_positions_tx` (line 231) and `build_merge_positions_tx` (line 547) which were updated in commit 20c01af2 to return `data=bytes.fromhex(data[2:])` per the strict-bytes contract introduced by open-autonomy PR #2477. The current consumer (`omen_withdraw.py:408`) wraps in `HexBytes(...)` so it works today, but the convention bites the next caller that doesn't. Match the sibling shape.

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
```diff
@@ -17,34 +17,505 @@
...
+
+    :param amount: the base amount (e.g. ``calcSellAmount`` output).
+    :param slippage: fraction in [0, 1].
+    :return: ``int(amount * (1 + slippage)) + 1`` (the ``+1`` forces
+        round-up so even a 0% slippage still strictly exceeds ``amount``,
+        guarding against integer-floor under-shoot).
+    """
+    if not 0 <= slippage <= 1:
+        raise ValueError(f"slippage {slippage!r} must be in [0, 1]")
+    return int(amount * (1 + slippage)) + 1
```
DIvyaNautiyal07: Float multiplication on wei amounts. For values past 2^53 (~9e15 wei = 0.009 wxDAI in 18-dec scale; a 1000 wxDAI position is 1e21 wei) float64 starts losing wei precision. The project rule is 'Never use floating point for money or token amounts'. An integer-only ceil works: `amount + (amount * num + den - 1) // den + 1` with `num/den = slippage`. The `+1` round-up guard is the right shape; just keep it on integers.

packages/valory/skills/decision_maker_abci/rounds.py
DIvyaNautiyal07: Asymmetric with `OmenWithdrawRound` two blocks above, which routes `WITHDRAWAL_ROUND_TIMEOUT → WithdrawalIdleRound` (line 566). Looping the post-settlement round back to itself on timeout means a consistently un-parseable receipt would re-enter the same parse/record path forever rather than falling through to the idle terminal. Either pick the sibling's exit (probably right — the tx already settled, no need to retry parsing), or document why the post-round is special.

packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py
```diff
@@ -5343,3 +5358,237 @@ def test_missing_date_key_skipped_gracefully(self) -> None:
...
+                    condition_id=self.CONDITION_B,
+                    outcome_index=1,
+                    amount_wei=3 * WEI_IN_ETH,
+                    shares_wei=5 * WEI_IN_ETH,
+                ),
+            ]
+        }
+        assert self._drive(b._compute_omen_funds_locked(trader_agent, self.SAFE)) == 5.0
+
+    def test_held_set_none_falls_back_to_no_gate(self) -> None:
```
DIvyaNautiyal07: Test name `test_held_set_none_falls_back_to_no_gate` and docstring 'Subgraph fetch failing returns empty held set → all positions excluded' contradict each other — one says 'no gate', one says 'fully gated'. The assertion (`result == 0.0`) backs the docstring, so the test name is wrong. This is the same issue surfaced on the fetcher: the test silently locks in the broken behaviour the fetcher's own docstring claims is wrong. Fix the fetcher (per the blocking comment) and either rename this test or replace it with one that exercises the `held_keys=None` no-gate path.

DIvyaNautiyal07 (changes requested):

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
LOCKhart07: Fixed in 3f79888c — `_fetch_ct_held_position_keys` now returns `Optional[set]` (`None` on error / missing context). Downstream `compute_funds_locked_from_bets` already treats `None` as "no gate" and falls back to the un-gated sum, so a transient CT-subgraph hiccup no longer writes a phantom `0.0`. Misnamed test renamed + a new `test_none_held_set_falls_back_to_no_gate` regression guard added.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Fixed in 3f79888c — `participant_remaining_cost` is now aggregated per-FPMM (sums across the agent's outcome indices on the SAME fpmm.id, not across all FPMMs). Consumer (`_calculate_bet_net_profit`) unchanged. Added `test_invalid_refund_scoped_per_fpmm_across_multi_market_history` exercising the full `allocate_fifo → _calculate_bet_net_profit` chain on a 2-FPMM scenario.

packages/valory/skills/market_manager_abci/graph_tooling/requests.py
LOCKhart07: Fixed in 3f79888c — `fetch_withdrawal_creator_fpmms` (and `fetch_user_positions`) now set `_fetch_status = FetchStatus.FAIL` explicitly on the error-return path, so a future caller gating on `_fetch_status` sees a terminal state. `fetch_trades` left alone (pre-existing, out of scope).

packages/valory/skills/market_manager_abci/graph_tooling/requests.py
LOCKhart07: Fixed in 3f79888c — null-check switched from `res is None` to `rows is None` / `positions is None` (the post-handler result). Behaviour-equivalent given `_handle_response`'s invariant, but expresses intent correctly.

packages/valory/skills/market_manager_abci/graph_tooling/utils.py
LOCKhart07: Fixed in 3f79888c — new local helper `_is_finalized_timestamp(value)` treats `None`, `"0"`, `0`, negative, and malformed values as not-finalized. Mirrors `parse_timestamp` in `aps/predictions_helper.py` but kept inline to avoid an upward cross-skill dependency. 3 new test cases (string "0" / int 0 / malformed).

packages/valory/contracts/market_maker/contract.py
LOCKhart07: Fixed in 3f79888c — layered defence: (1) structural guards on topic length / address / data with tailored diagnostics, (2) try/except wrapping the byte-slice math as a safety net, (3) module-level `_logger` so each dropped log carries idx + tx hash + fpmm address. A single malformed log no longer takes down the whole receipt's audit trail. 5 new test cases.

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
LOCKhart07: Fixed in 3f79888c — the `return_amount == 0` `break` now self-records a distinct error ("halved to zero before sizing converged — FPMM pool too thin") and `return None`s directly. The `for/else` for max-attempts also got a more actionable message ("tighten withdrawal_slippage / withdrawal_return_buffer"). The post-loop guard is now relabelled as defensive-only. 3 new tests covering each failure mode independently.

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
LOCKhart07: Fixed in 3f79888c — `TOP_LEVEL_ERROR_TOKEN_ID` moved into the new `OmenWithdrawalStore` module with a 4-line justification block above it. Same for the FPMMSell fill placeholder. Inline justifications also added to the public-address `# nosec B105` annotations introduced for the new test fixtures.

packages/valory/skills/decision_maker_abci/behaviours/post_omen_withdraw.py
LOCKhart07: Fixed in 3f79888c — `_snapshot_funds_locked` now skips with a warning when `trader_agent.get("bets") == []` (post-sweep state means the safe definitively has buy history; empty bets is unambiguously indexer lag). New `test_empty_bets_skips_write_on_indexer_lag` regression guard.

packages/valory/skills/decision_maker_abci/behaviours/post_omen_withdraw.py
LOCKhart07: Fixed in 3f79888c — extracted `OmenWithdrawalStore` (`behaviours/omen_withdrawal_store.py`) with 9 methods covering the full chatui-store I/O. Both behaviours hold a lazy `_store` property and delegate. The drifted `_record_top_level_error` signature is unified to a single `(reason: str)` form; callers compose their own context (`f"{op_name}: retries exhausted"` in the planning side). 22 new direct unit tests on the store helper (no behaviour mocks needed). The O(N) disk-I/O point is a separate optimisation; not bundled here.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Fixed in 3f79888c — promoted `_allocate_fifo` to a module-level pure function `allocate_fifo(bets, logger)`. `compute_funds_locked_from_bets` now calls it directly without constructing a throwaway fetcher. The class method becomes a one-line wrapper preserving all internal callers. New `TestAllocateFifoModuleLevel` class with a wrapper-parity test guarding against future drift.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Fixed in 3f79888c — `_fetch_ct_held_position_keys` now returns `None` on subgraph error (per the C1 fix), so `held_keys=None` is the explicit "no gate" signal and a legitimately-empty set means "user holds nothing on-chain". The two callsites have aligned semantics now.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Fixed in 3f79888c — stripped all rotting spec/PR refs introduced by this PR (`spec §6.3 / §6.5a / §7.2 / §7.3 / §8.1 / §8.2 / §9.3 / §10.13.3 / §13.3 / §13.9`, `PR #948 review round 2`, `PR #952's first live run`, `Line 867 / 954 / 971 / 978`). Replaced with inline rationale that survives without the external spec doc. Pre-existing `ZD#919` and `PR #903` refs left alone (out of scope).

packages/valory/contracts/market_maker/contract.py
LOCKhart07: Fixed in 3f79888c — added a planned-FPMM allowlist persisted via `OmenWithdrawalStore.record_planned_fpmms(fpmms)` at the end of `OmenWithdrawBehaviour._build_payload`. `PostOmenWithdrawBehaviour._parse_sell_events` filters decoded events against the allowlist (case-insensitive). Missing allowlist → fall through with a warning (backwards compatible). 6 new store tests + 4 new behaviour-level filter tests.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
LOCKhart07: Fixed in 3f79888c — confirmed and addressed. Fetcher returns `Optional[set]` (`None` on `AttributeError` AND on the first-page `_fetch_from_subgraph` falsy path). The misnamed test (`test_held_set_none_falls_back_to_no_gate` asserting `== 0.0`) was renamed to `test_empty_held_set_excludes_all_positions` and the actual fall-back semantic gets a new test asserting the un-gated sum is written.

packages/valory/skills/market_manager_abci/graph_tooling/utils.py
LOCKhart07: Fixed in 3f79888c — used the local `_is_finalized_timestamp` helper described above. Treats `None`, `"0"`, `0`, negative, and malformed values as unset (matches `parse_timestamp` semantics from PR #903 review).

packages/valory/skills/decision_maker_abci/rounds.py
LOCKhart07: Fixed in 3f79888c — `Event.WITHDRAWAL_ROUND_TIMEOUT` now routes to `WithdrawalIdleRound` for consistency with the upstream `OmenWithdrawRound` (and the sibling redeem rounds). Receipt parsing is deterministic, so retrying the same input won't unblock a real bug. FSM specs regenerated; new test `test_post_omen_withdraw_routes_to_idle_on_round_timeout` mirrors the existing OmenWithdrawRound timeout-escape test.

packages/valory/contracts/conditional_tokens/contract.py
LOCKhart07: Fixed in 3f79888c — `build_set_approval_for_all_tx` now returns `bytes.fromhex(data[2:])` matching the other `build_*_tx` methods in the file. Caller in `omen_withdraw.py` already wraps in `HexBytes(data)` (tolerates both forms), so no caller change. Existing test updated.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Fixed in 3f79888c — `allocate_fifo` now uses `int` throughout: `int()` parsing with a `try/except ValueError, TypeError` for malformed strings, integer mul-then-divide for the share/cost ratios (`(proceeds_total * take) // (-row_shares)`), and `defaultdict(int)` for the per-FPMM aggregation. Per-step rounding error is bounded at 1 wei vs the float drift past 2^53 it replaces. New regression test `test_int_math_no_float_drift_across_many_sells` runs 100 partial sells and asserts `remaining_shares == 0` exactly.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
LOCKhart07: Already fixed in 3f79888c — `_fetch_ct_held_position_keys` now returns `Optional[set]` (`None` on `AttributeError` AND on the first-page subgraph-fetch falsy path). `compute_funds_locked_from_bets` already treats `None` as "no gate" and the misnamed test that baked in `== 0.0` was renamed + a new test asserts the un-gated fall-back is written.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Already fixed in 3f79888c — same C1 fix. `held_keys=None` signals "no gate / fall back to un-gated sum"; legitimately-empty `set()` means "safe holds nothing on-chain" and correctly excludes everything. The two callsites have aligned semantics.

packages/valory/contracts/conditional_tokens/contract.py
LOCKhart07: Already fixed in 3f79888c — `build_set_approval_for_all_tx` now returns `bytes.fromhex(data[2:])` matching the sibling `build_*_tx` methods. Caller `omen_withdraw.py` wraps in `HexBytes(...)` so backwards-compatible; test assertion updated.

packages/valory/skills/decision_maker_abci/rounds.py
LOCKhart07: Already fixed in 3f79888c — `Event.WITHDRAWAL_ROUND_TIMEOUT` routes to `WithdrawalIdleRound` (mirrors upstream `OmenWithdrawRound` and sibling redeem rounds). FSM specs regenerated; new `test_post_omen_withdraw_routes_to_idle_on_round_timeout` mirrors the existing escape-test.

packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py
LOCKhart07: Already fixed in 3f79888c — test renamed to `test_empty_held_set_excludes_all_positions` (the legitimately-empty "user holds nothing" path, asserts `== 0.0` correctly). The actual "no gate" fallback gets a separate new test `test_none_held_set_falls_back_to_no_gate` which mocks the fetcher returning `None` and asserts the un-gated sum is written.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Fixed in b47d5b6b — confirmed and addressed. `_build_market_context` now divides `totalPayout` AND `totalTraded` by `WEI_TO_NATIVE` at construction, matching the specific-market path at line 636-637 which already scaled them. Affects both the invalid refund path (line 1365) and the winning payout share (line 1397). Added `test_total_payout_normalised_to_wxdai` exercising the wei-input → wxDAI-output conversion at the boundary (the tests previously masked the bug by injecting hand-built wxDAI-scale ctx dicts).

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
LOCKhart07: Fixed in b47d5b6b — `_size_and_build_position` now takes `include_approval: bool` (defaults True). `_build_payload` tracks approved FPMMs across positions via a `set` and passes `include_approval=False` on subsequent calls for the same FPMM. A safe holding two outcomes on the same FPMM now emits only one `setApprovalForAll(FPMM, true)` in the multisend. New test `test_size_and_build_skips_approval_when_requested` covers both paths.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
LOCKhart07: Fixed in b47d5b6b — `_format_predictions` now re-sorts `enriched_buys` DESC-by-`blockTimestamp` at the consumer boundary, so the allocator's ASC-within-group ordering invariant (required for FIFO) is preserved without leaking into the formatted output. New `test_output_ordered_desc_by_blocktimestamp` regression guard with 3 bets across 2 FPMMs at distinct timestamps.

packages/valory/skills/decision_maker_abci/behaviours/omen_withdraw.py
LOCKhart07: Fixed in b47d5b6b — `inflate_for_slippage` now uses integer-only ceiling math. `slippage` is converted to a 1e9-denominator integer fraction at the function boundary (`int(slippage * 1e9)`, which is float-exact for any reasonable slippage), and the per-amount multiplication stays in Python's arbitrary-precision int. The `+1` round-up safety preserves the strict-ceiling invariant. New regression test `test_inflate_for_slippage_int_math_at_wei_scale` runs on a 1000 wxDAI position (1e21 wei, well past 2^53) and asserts the exact integer result.

jmoreira-valory (dismissed):
Approving. Status of my 2 comments:

- **#1 — `parse_sell_events` log address validation**: Fully resolved in `3f79888c`. `OmenWithdrawalStore.record_planned_fpmms()` persists the planned-FPMM allowlist at the end of `OmenWithdrawBehaviour._build_payload`; `PostOmenWithdrawBehaviour._parse_sell_events` filters decoded events against the allowlist (case-insensitive), with a "fall through + warning" path when the allowlist is missing (backwards compatible). 6 new store tests + 4 new behaviour-level filter tests cover both the happy and missing-allowlist paths. Audit-trail integrity invariant restored.

- **#2 — unauthenticated POST endpoint (re-raise of #942)**: explicitly acknowledged in commit `3f79888c` as deferred to a separate service-level decision ("The unresolved auth concern (re-raise of #942) is left for a separate service-level decision"). Signing off so the merge isn't held on this, but flagging it again as a standing concern — Omenstrat's on-chain destructive radius (Safe multisend at default 1% slippage + 5% buffer) makes the default `HTTP_SERVER_HOST: 0.0.0.0` binding harder to defend than on Polystrat. At minimum, the `services/trader_pearl/service.yaml` default should flip to `127.0.0.1` in a follow-up, with env override for any genuine remote-ops use case.

LGTM modulo the standing security note on #2.

LOCKhart07:
@jmoreira-valory — addressed in `671ea8d8`. Took your **Option 1** (cheapest): defaulted `HTTP_SERVER_HOST` to `127.0.0.1` in both `services/trader_pearl/service.yaml:371` and `services/polymarket_trader/service.yaml:371`.

Net effect:
- The HTTP server fronting the agent (and the unauthenticated `POST /api/v1/withdrawal` route) now only binds to loopback by default. Propel ingress can't reach it without the operator explicitly opting back to a public binding (`HTTP_SERVER_HOST=0.0.0.0`).
- Same change applies retroactively to Polystrat (`trader_pearl` was already shipping with the wide binding from #942/#943) — so this closes the original concern, not just the Omenstrat re-raise.
- No code change in the handler; no token plumbing. If a deployment genuinely needs remote management, the env-var override is the explicit opt-in.

A token-based or EOA-challenge layer (your options 2/3) on top of the loopback default is the natural defence-in-depth follow-up — happy to land that as a separate PR if you want it.

jmoreira-valory (dismissed):
Re-approving. Both my comments now fully resolved:

- **#1 — `parse_sell_events` log address validation**: planned-FPMM allowlist via `OmenWithdrawalStore.record_planned_fpmms()` + filter in `PostOmenWithdrawBehaviour._parse_sell_events` (3f79888c). 10 new tests covering happy/missing-allowlist paths.
- **#2 — unauthenticated POST + `0.0.0.0` binding** (the re-raise of #942 I'd flagged as a standing concern in my prior approval): fully closed in `671ea8d8`. Both `trader_pearl/service.yaml` and `polymarket_trader/service.yaml` now default `HTTP_SERVER_HOST=127.0.0.1`. Env-var override (`HTTP_SERVER_HOST=0.0.0.0`) preserved for operators with a genuine remote-ops need. Hashes propagated.

The safe-by-default loopback binding matches the chatui-only UX the FE was designed around — and closes the same gap that's been open on #942 since I first flagged it. Good outcome.

LGTM.

DIvyaNautiyal07 (dismissed):

bennyjo (dismissed):
LGTM now.

bennyjo (approved):
LGTM still.

DIvyaNautiyal07 (approved):

PR #953, opened 2026-05-12
feat: valid_mechs + valid_tools allowlist + ChatUI mech pin

DIvyaNautiyal07:
## Summary
- Wire trader to the new mech_interact_abci allowlist API (bumped to PR #79: `bafybeibcr5ifx...`). Drops the legacy `ignored_mechs` blocklist; introduces flat `valid_mechs` + `valid_tools` env-overridable lists in `trader_pearl` and `polymarket_trader` service.yaml.
- Honor a consumer-set `selected_mechs` pin end-to-end: ChatUI handler validates and persists the pin (against `available_valid_mechs` derived from `mechs_info`), `tool_selection` reads it from chatui's store and writes it into the consensus payload (lowercased), `decision_maker_abci` payload/state pass it to the mech_interact round.
- Seed default `valid_mechs` (6 Gnosis mechs in trader_pearl, 3 Polygon mechs in polymarket_trader, NVM-native mech excluded) and `valid_tools` (full deployed tool set per chain). `irrelevant_tools` continues to filter at runtime as a defense layer.

## Test plan
- [x] `make test` / per-skill pytest: 498 tests green across decision_maker_abci, chatui_abci, trader_abci
- [x] `autonomy packages lock --check` verifies consistent hashes end-to-end
- [x] `autonomy push-all` republished services to IPFS
- [ ] Pearl staging deploy: pull `trader_pearl bafybeibbtnqpv4dzfrlzphu572t2gbyttxvcqafuvsjb24cxzfh5r2o6ru`, confirm agent boot and a mech roundtrip honors the allowlist
- [ ] Polystrat staging deploy: pull `polymarket_trader bafybeiejlh25yuhqamcsvarcraos2bzuturgrmg3jquoetx4ndkiry52oi`
- [ ] ChatUI: set a `selected_mechs` pin via prompt, verify decision_maker constrains tool selection to that pin

Companion PR in mech-interact: #79 (merged).


mariapiamo (commented):
1. High: selected_mechs is synchronized too late.
ChatuiLoadBehaviour only sends vote=True, so ChatUI-persisted selected_mechs is not written to synchronized data before mech discovery runs: packages/valory/skills/chatui_abci/behaviours.py:53.
The composed FSM then goes FinishedChatuiLoadRound -> MechVersionDetectionRound, and FailedMechInformationRound -> MechVersionDetectionRound: packages/valory/skills/trader_abci/composition.py:166-170.
But the PR only writes selected_mechs in ToolSelectionBehaviour.async_act, after tool selection: packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py:113-141.

Consequence: if synchronized data already contains a stale/bad pin and mech-interact fails MechInformationRound because pinned mechs are offline/no valid tools, the app can loop before reaching ToolSelectionRound. A ChatUI clear/update remains local and cannot overwrite the stale synchronized value.
Suggested fix: sync selected_mechs during ChatuiLoadRound or another round before MechInformationRound.

2. Medium: empty ChatUI restrictions fail open.
If selected_mechs/allowed_tools leave no candidate tools, the code falls back to unrestricted policy: packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py:86-92.
That makes the narrowing a soft preference, not an allowlist. For allowed_tools-only, this can select a tool outside the user’s selected list. If this is intentional, it should be documented as soft filtering; otherwise it should fail closed with a clear reason/no-trade.

3. Open Question
The service configs seed valid_mechs and valid_tools, but the base agent config does not expose them: packages/valory/agents/trader/aea-config.yaml:270-285. If standalone agent execution is supported, this leaves behavior dependent on mech-interact’s empty-list semantics. If only trader_pearl/polymarket_trader services are supported, this is probably fine.

packages/valory/agents/trader/aea-config.yaml
```diff
@@ -282,7 +282,6 @@ models:
       polymarket_ctf_exchange_address: ${str:0xE111180000d2663C0091e4f400237545B87B996B}
       polymarket_neg_risk_ctf_exchange_address: ${str:0xe2222d279d744050d28e00520010520000310F59}
       polymarket_neg_risk_adapter_address: ${str:0xd91E80cF2E7be2e162c6513ceD06f1dD0dA35296}
-      ignored_mechs: ${list:[]}
```
bennyjo, line 288: `ignored_mechs` is removed here, but the new `valid_mechs` / `valid_tools` params (now required by mech_interact_abci) aren't added at the agent level — compare with `irrelevant_tools` on line 233, which carries an agent-level default. Per `CLAUDE.md`, `make run-agent` uses this agent-level config (service-level overrides only kick in under `autonomy deploy`), so once `trader_abci/skill.yaml`'s `valid_mechs: []` / `valid_tools: []` defaults take over, mech-interact hits `last_failure_reason = "allowlist_not_configured"` and emits NONE on every `MechInformationRound` entry — local Omen dev silently breaks. Previously `ignored_mechs: []` meant "nothing blocked"; the equivalent new state (empty allowlist) means "nothing allowed".

Was this deliberate (operators must always export `VALID_MECHS` / `VALID_TOOLS` for local dev)? If yes, happy to leave as-is — worth a README note alongside the existing Polystrat overrides. If not, mirroring the Omenstrat defaults from `trader_pearl/service.yaml` here would restore the prior local-dev behavior.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
```diff
@@ -49,29 +83,22 @@ def _select_tool(self) -> Generator[None, None, Optional[str]]:
...
-                restricted_policy.update_weighted_accuracy()
-                selected_tool = restricted_policy.select_tool(randomness)
-            else:
-                self.context.logger.warning(
-                    f"None of the allowed tools {chatui_allowed_tools} are in the "
-                    f"current tool set {self.mech_tools}. Falling back to unrestricted policy."
-                )
-                selected_tool = self.policy.select_tool(randomness)
+        candidate_tools = self._candidate_tools()
+        if not candidate_tools:
```
bennyjo, line 95: Edge case worth checking: when `selected_mechs` is set but `self.synchronized_data.mechs_info` is empty for this round (early boot before `MechInformationRound` has succeeded, or every mech's tool fetch failed), `_candidate_tools` collapses to `set()` at line 66 above, this branch fires, and the policy picks unrestrictedly from the full `self.mech_tools` — silently violating the user's pin for the round.

The `selected_mechs_payload` on line 130 still carries the pin into consensus, so mech-interact catches the mismatch downstream via `pinned_mechs_offline` / `pinned_mechs_no_valid_tools` and emits NONE — but only after a tool has been chosen and committed. Two suggestions:
1. Distinguish "no pin → fall back to full policy" from "pin set but unsatisfiable → emit NONE and retry next round". We already know mech-interact will reject the pinned-but-empty case, so picking a doomed tool just adds a round of churn.
2. At minimum, log at error severity and name `selected_mechs` explicitly — the current `"ChatUI restrictions left no candidate tools"` conflates pin-collapse with `allowed_tools`-collapse, which makes user-reported "my pin isn't working" hard to correlate.

bennyjo (commented):
Two scoped findings — one on the agent-level config (local-dev regression risk), one on the new pin-collapse fallback path. Framing as questions; happy to leave either as-is if deliberate.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
```diff
@@ -95,6 +123,13 @@ def async_act(self) -> Generator:
                 mech_tools = json.dumps(list(self.mech_tools))
                 policy = self.policy.serialize()
                 utilized_tools = json.dumps(self.utilized_tools, sort_keys=True)
+                # Serialize the user's mech pin so mech-interact's
+                # relevant_mechs_info can restrict ranking to it after
+                # consensus. Empty list means "no pin".
+                pinned_mechs = self.shared_state.chatui_config.selected_mechs or []
+                selected_mechs_payload = json.dumps(
+                    [str(m).lower() for m in pinned_mechs]
```
LOCKhart07: When `_candidate_tools()` returns empty above, `_select_tool` falls back to the unrestricted policy and may choose a tool no pinned mech actually serves — but `pinned_mechs` is still serialized here unchanged. Downstream, mech-interact's `relevant_mechs_info` filter intersects the chosen tool with the pin and ends up with an empty list, setting `last_failure_reason = "no_overlap_with_selected_mechs"`.

So the trader knows during tool selection that the pin is unsatisfiable but still publishes it, forcing mech-interact to rediscover the same fact and surface a different failure reason. Worth either (a) clearing the pin payload (`"[]"`) when `_candidate_tools` was empty and we fell back, or (b) tightening the fallback warning to identify which constraint emptied the set, so the two layers don't disagree about whether the pin applies this round.

packages/valory/skills/chatui_abci/handlers.py
```diff
@@ -307,6 +321,18 @@ def _send_chatui_llm_request(
...
+        """Return the set of mech addresses currently visible to mech-interact.
+
+        Falls back to an empty set if the synced data has not yet been
+        populated (e.g. before MechInformationRound has run for the first
+        time). Caller is responsible for handling the empty case.
+        """
+        try:
+            return self.synchronized_data.available_valid_mechs
+        except (TypeError, KeyError):
+            return set()
```
LOCKhart07: The `try/except (TypeError, KeyError)` is dead. `SynchronizedData.available_valid_mechs` is already fully defensive: it uses `db.get(..., "[]")`, checks `isinstance(raw, str)`, catches `json.JSONDecodeError` and `TypeError` from `json.loads`, and filters non-dict entries / missing addresses. There's no path that raises `TypeError` or `KeyError` out of the property — and the docstring above even says it falls back to an empty set internally.

Worse, if a future change makes the property strict, this wrapper would silently swallow it. Suggest dropping the try/except and calling `self.synchronized_data.available_valid_mechs` directly.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
```diff
@@ -37,6 +37,40 @@ class ToolSelectionBehaviour(StorageManagerBehaviour):
...
+        2. `allowed_tools` — keep only tools the user pinned by name.
+
+        Either layer being empty/None is a no-op for that layer.
+
+        :return: the candidate tool set after applying the pin filters.
+        """
+        candidate = set(self.mech_tools)
+
+        selected_mechs = self.shared_state.chatui_config.selected_mechs
+        if selected_mechs and not self.benchmarking_mode.enabled:
```
LOCKhart07, line 56: Two small things on this branch:

1. **Benchmarking-mode mismatch.** Here we skip the `selected_mechs` filter when benchmarking, but `async_act` below still serializes `chatui_config.selected_mechs` into the payload unconditionally. Benign today (benchmarking mocks the mech request), but the two paths should make the same decision about whether the pin applies this round.

2. **Empty `mechs_info` produces a noisy fallback.** If the user sets a pin before `MechInformationRound` has populated `mechs_info`, `tools_from_pinned_mechs` is the empty set, `candidate` becomes empty, and every tool-selection round logs *"ChatUI restrictions left no candidate tools; falling back to the unrestricted policy."* Consider an early `if not self.synchronized_data.mechs_info: skip pin filter` so the early-boot case is distinguished from a genuinely unsatisfiable pin.

packages/valory/agents/trader/aea-config.yaml
DIvyaNautiyal07: Thanks Benny, added agent-level `valid_mechs`/`valid_tools` defaults mirroring `trader_pearl/service.yaml` in 0c908e77. `make run-agent` keeps the prior local-dev behaviour instead of emitting `allowlist_not_configured` on every `MechInformationRound` entry.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
DIvyaNautiyal07: Good edge case, both addressed in 0c908e77.

The pin filter is now skipped when `self.synchronized_data.mechs_info` is empty for the round, so the boot-race case (before `MechInformationRound` has run for the first time) no longer silently violates the pin. When the pin actually has no overlap (mechs_info populated but yields zero candidates), `_select_tool` returns `None` so the round skips this iteration instead of letting the unrestricted policy pick a doomed tool.

Log message also names the offending constraint (`selected_mechs` vs `allowed_tools`) so user-reported issues correlate cleanly. And the pin payload field was removed from `ToolSelectionPayload` entirely (chatui-load is now the source of truth), so the two layers can no longer disagree about whether the pin applies this round.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
DIvyaNautiyal07: Took the stronger version of (a) in 0c908e77: removed `selected_mechs` from `ToolSelectionPayload` and its `selection_key` entirely. The pin now flows through `ChatuiLoadRound.end_block` at the top of each FSM iteration, so the two layers can no longer disagree about whether the pin applies this round.

packages/valory/skills/chatui_abci/handlers.py
DIvyaNautiyal07: Good catch, dropped in 0c908e77. The helper now just returns `self.synchronized_data.available_valid_mechs` since the property already handles the missing/malformed cases internally.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
DIvyaNautiyal07: Both addressed in 0c908e77.

`_candidate_tools` now treats the pin as inactive when `synchronized_data.mechs_info` is empty, so the early-boot case is distinguished from a genuinely unsatisfiable pin. The noisy fallback warning is gone for the boot-race path.

Benchmarking is consistent too: the payload no longer carries a per-round pin (chatui-load owns it now via `ChatuiLoadRound.end_block`), so there is no way for the filter and serialization to disagree about whether the pin applies.

DIvyaNautiyal07:
Thanks Mariapia, all three addressed in 0c908e77.

On the High: moved the `selected_mechs` sync into `ChatuiLoadRound.end_block`, so `synchronized_data.selected_mechs` is fresh before `MechInformationRound` runs in the same iteration. The stale-pin loop you described (`FailedMechInformationRound` -> `MechVersionDetectionRound` without ever reaching `ToolSelectionRound`) is gone. `ChatuiPayload` picked up a `selected_mechs` field for this, and I dropped the now-redundant write in `ToolSelectionRound` so chatui-load is the single source of truth.

On the Medium: pin-collapse now fails closed. `_select_tool` returns `None` when a ChatUI restriction empties the candidate set, and the warning names which constraint (`selected_mechs` or `allowed_tools`) so user-reported issues correlate cleanly.

On the Open Question: added `valid_mechs` and `valid_tools` defaults at the agent level in `aea-config.yaml` mirroring `trader_pearl`, so `make run-agent` keeps the prior local-dev behaviour instead of emitting `allowlist_not_configured`.

bennyjo (dismissed):
Both prior blocking items resolved cleanly: agent-level `valid_mechs` / `valid_tools` defaults restore `make run-agent` parity, and the unsatisfiable-pin path now fails closed with a named constraint instead of silently falling back to the unrestricted policy. The author went further and moved the pin write into `ChatuiLoadRound.end_block` so mech-interact sees the fresh pin in the same FSM iteration — strictly better than the previous one-iteration-late propagation. Override is consistent with the parent `VotingRound.end_block` contract (verified). Two soft questions left in the chat-side writeup for follow-up — non-blocking:

1. `_candidate_tools` now skips the pin filter when `synchronized_data.mechs_info` is empty. Given composition runs `ChatuiLoadRound → … → MechInformationRound → … → ToolSelectionRound`, is this guard purely defensive (no real reachable state), or is the trade-off "unrestricted pick now, loud mech-request failure one round later" deliberate vs returning `None`?
2. `FailedMechInformationRound → MechVersionDetectionRound` (composition.py:175) — if mech-interact enters a failure loop on `pinned_mechs_offline`, the FSM doesn't appear to re-enter `ChatuiLoadRound` until something else breaks the loop. Did you trace the exit path that lets a user-cleared pin propagate during a sustained failure window?

LGTM either way — happy to defer both to a follow-up if you've already weighed the trade-offs.

LOCKhart07 (dismissed):
Re-reviewed at 996c7d88 (was 32e43ca2). Prior four review threads are all addressed cleanly in 0c908e77:

- Pin write moved from `ToolSelectionRound` to `ChatuiLoadRound.end_block` — single source of truth, fresh before `MechInformationRound` runs each iteration; closes the stale-pin deadlock.
- Pin-collapse fails closed (`_select_tool` returns `None`) and the warning names the offending constraint (`selected_mechs` vs `allowed_tools`).
- Boot-race path (empty `mechs_info`) skips the pin filter without noisy warnings.
- Dead `try/except` around `available_valid_mechs` dropped.
- Agent-level `valid_mechs`/`valid_tools` defaults seeded in `aea-config.yaml` so `make run-agent` keeps Omenstrat parity.

Two non-blocking notes for a follow-up:

1. **Test gap in `chatui_abci/tests/test_behaviours.py::TestAsyncAct::test_async_act_happy_path`.** The new `selected_mechs` serialization in `ChatuiLoadBehaviour.async_act` is uncovered at the behaviour level. The test still passes because `MagicMock.__iter__` defaults to `iter([])`, so `for m in pinned` produces an empty comprehension and `selected_mechs_payload` falls through to `"[]"`. A regression like "forgot to lowercase" or "forgot to `json.dumps`" would slide through silently. The payload-side (`test_payloads.py::test_chatui_payload_carries_selected_mechs`) and round-side (`test_rounds.py::TestChatuiLoadRoundEndBlock`) tests cover their slices; this is just closing the behaviour-side bookend with one assertion that sets `chatui_config.selected_mechs = ["0xABC"]` and checks `payload.selected_mechs == '["0xabc"]'`.

2. **Sovereign-deployment assumption worth a code comment in `ChatuiLoadRound.end_block`.** `positive_payloads[0].selected_mechs` picks by dict-insertion order. In single-agent that's the only payload; in a multi-agent quorum, each agent's `chatui_config.selected_mechs` is independent (the HTTP handler only updates the receiving agent's `shared_state`), so divergent submissions would write different values to synced data and could diverge Tendermint consensus. Fine today per `CLAUDE.md`'s single-agent shape, but the dict-order pick makes the round consensus-fragile in a way the prior `ToolSelectionPayload`/`selection_key` approach wasn't. One line of context here would save the next contributor from a flaky-consensus chase on a multi-agent migration.

DIvyaNautiyal07:
Thanks again Benny. Landed both in c294b1d1 rather than deferring (no follow-up PR planned for this scope).

Q1: dropped the `bool(self.synchronized_data.mechs_info)` check from `_candidate_tools` entirely. v2 populates `mechs_info` during `MechInformationRound` before `ToolSelectionRound` reads it, so an empty value at that point isn't a boot race — it's a real failure mode (e.g. subgraph returned zero matching mechs). The round now fails closed in that case via the existing `cause = "selected_mechs"` path. Updated the test (`test_pin_set_but_mechs_info_empty_fails_closed`) to pin the new contract.

Q2: routed `FailedMechInformationRound → ChatuiLoadRound` in `composition.py` (was `MechVersionDetectionRound`). The tight failure loop now re-enters chatui-load on every iteration, so a consumer-cleared pin reaches synced data within one round instead of waiting for the period rollover. Cost is one extra round of chatui-load on each mech-info failure, which is cheap (disk read + vote True + write synced.selected_mechs). FSM specs and docstrings regenerated; lock cascade republished.

DIvyaNautiyal07:
Thanks Jenslee, both landed in c294b1d1:

Note 1: added `test_async_act_serializes_pinned_mechs_lowercased` and `test_async_act_empty_pin_serializes_as_empty_list` in `chatui_abci/tests/test_behaviours.py`. The first sets `chatui_config.selected_mechs = ["0xABC", "0xDEF"]` and asserts the payload carries `'["0xabc", "0xdef"]'`, the second pins the `None` → `"[]"` branch. Both would have caught the `MagicMock.__iter__ = iter([])` regression you flagged.

Note 2: added a focused comment in `ChatuiLoadRound.end_block` flagging the single-agent dict-order pick and what to revisit on a multi-agent migration (most_common consensus rule). Going with comment-only here per your suggestion since the consensus fix is untested under n=1.

LOCKhart07 (approved):
Re-reviewed at `c294b1d1` (was `996c7d88`). All prior items resolved:

- **Test gap (my note 1)**: `test_async_act_serializes_pinned_mechs_lowercased` + `test_async_act_empty_pin_serializes_as_empty_list` close the behaviour-side bookend — both correctly assert lowercased + `json.dumps` output, would catch the `MagicMock.__iter__` slide.
- **Multi-agent caveat (my note 2)**: comment in `ChatuiLoadRound.end_block` is concrete (names `most_common` as the migration target).
- **Boot-race guard (bennyjo Q1)**: dropped entirely; pin now applies whenever set, empty `mechs_info` fails closed via the existing `cause = "selected_mechs"` path. Justified by FSM ordering (`MechInformationRound` populates `mechs_info` before `ToolSelectionRound` reads it). Test updated.
- **Sustained failure loop (bennyjo Q2)**: `FailedMechInformationRound → ChatuiLoadRound` lets a consumer-cleared pin propagate within one round of a `pinned_mechs_offline` loop. FSM spec updated in lockstep. Cost is one cheap chatui-load per failure iteration — fine.

LGTM.

bennyjo (approved):
Follow-up commit `c294b1d1` resolves both prior soft questions with code:

- `FailedMechInformationRound → ChatuiLoadRound` (composition.py:175, fsm_specification.yaml:229) gives a consumer-cleared pin a reachable path into synced data during a sustained `pinned_mechs_offline` loop, instead of being trapped behind the MechVersion ↔ MechInformation cycle.
- `_candidate_tools` drops the boot-race `bool(mechs_info)` guard — pin filter now applies unconditionally when set, and empty `mechs_info` fails closed via `cause = "selected_mechs"` rather than collapsing to unrestricted. Cleaner than the previous head: one fewer conditional, no silent-bypass window.
- `positive_payloads[0]` single-agent assumption is documented inline (rounds.py:119–123).
- New `ChatuiLoadBehaviour` tests catch the lowercase + json.dumps contract that the prior `MagicMock`-based happy path silently masked via `__iter__` defaulting to `iter([])` — useful annotation in the test docstring.

LGTM.

bennyjo (approved):
Follow-up resolves both soft questions in code: failure edge now routes through `ChatuiLoadRound`, and the boot-race bypass is gone. LGTM.

PR #955, opened 2026-05-13
fix(x402): retry LiFi quote with failing route denied on funding swap

LOCKhart07:
## Summary

- The x402 funding swap (Omenstrat xDAI→USDC.e, Polystrat POL→USDC) gets a LiFi quote and runs `eth_estimateGas`. When the LiFi solver picks a route whose simulation reverts contract-side (currently Velora/Paraswap on Gnosis returning Panic 0x11 / arithmetic underflow on a multi-hop xDAI→…→USDC.e path), the swap aborts and re-quotes on the next FSM tick keep returning the same broken route. The agent's x402 payment path effectively breaks.
- Verified against live LiFi API: `denyExchanges=paraswap` reroutes to Sushiswap, which estimates and would settle fine (~474k gas). LiFi's response carries the routing tool name (`quote["tool"]`) — the exact value to feed back into `denyExchanges`.
- This PR adds a bounded retry: on `ContractLogicError` from gas estimation, re-quote with the failing tool denied, up to 3 attempts. Non-revert failures (network / 4xx / funding / config) abort immediately as before — they wouldn't benefit from a re-quote.

## Changes

- `_get_lifi_quote`: new `deny_exchanges` kwarg → `denyExchanges` query param when non-empty.
- `_estimate_gas`: returns `(gas, route_revert)`; `ContractLogicError` → `(None, True)` (logged at WARNING — recovered failures shouldn't ERROR), anything else → `(None, False)`.
- `_ensure_sufficient_funds_for_x402_payments`: retry loop up to `X402_SWAP_MAX_ROUTE_ATTEMPTS=3`, denies the failing `tool` each iteration; nonce/gas-price fetch moved after the loop so the nonce window stays tight even when the loop runs ~5-8s.

Stays inside the existing single-worker `ThreadPoolExecutor` / `_x402_swap_future` guard — no new threads, no new locks. Only the wall time of the worker grows in the failure case; `_submit_x402_swap_if_idle` already debounces during that window.

## Verified end-to-end against the live agent

Live log after the fix:

```
15:06:14  Checking USDC balance for x402 payments...
15:06:15  USDC balance (237) < 200000, swapping xDAI to 250000 USDC...
15:06:19  Route revert in gas estimation: Panic 0x11 (paraswap, attempt 1)
15:06:19  LiFi route via 'paraswap' reverted on gas estimation; retrying with denyExchanges=['paraswap'] (attempt 1/3)
15:06:23  xDAI to USDC swap submitted: 0x46b7904cfe28cf79cb49f11fa4e86e781209663b71df043c9a352df25a866c21
15:06:26  xDAI to USDC swap completed successfully
```

Paraswap revert on attempt 1 → tool added to deny → second LiFi call routed via the working alternative → submitted and mined. Total wall-clock from balance check to confirmation: ~12s. On-chain: [tx 0x46b7904c…866c21](https://gnosisscan.io/tx/0x46b7904cfe28cf79cb49f11fa4e86e781209663b71df043c9a352df25a866c21) `status=success`, 1.06M gas, EOA `0xF17c…1843` → LiFi router.

## Review pass (commit `04f421a5`)

Folded in feedback from @DIvyaNautiyal07 and @OjusWiZard:

- `X402_SWAP_MAX_ROUTE_RETRIES` → `X402_SWAP_MAX_ROUTE_ATTEMPTS` (the name now matches what `range(N)` produces).
- Guarded the "retrying with denyExchanges=…" warning on the final iteration — it no longer contradicts the immediately-following `Exhausted LiFi route attempts` line.
- Demoted the `ContractLogicError` log from ERROR → WARNING so a successful retry no longer false-fires ERROR-level alerting. The caller still ERRORs on true exhaustion. (Insufficient-native-balance comes back as `ValueError` from web3 7.x, not `ContractLogicError`, so it's already filtered out and won't burn 3 retries.)
- `deny_exchanges=list(denied)` snapshot per call (was passing a live list reference, which made `MagicMock.call_args_list` assertions read state at assertion-time rather than call-time).
- `denied = denied + [tool]` immutable; dropped redundant `or None`; added attempt/denied context to the "Failed to get LiFi quote" error.
- Tests: new `test_route_revert_two_retries_then_success` covering the 2-deny accumulation boundary; pinned the `for/else` exhaustion branch with a log assertion in `test_route_revert_exhausted`.

## Test plan

- [x] `uv run pytest packages/valory/skills/trader_abci/tests/test_handlers.py` — 136 passed
- [x] `uv run pytest ... --cov=packages.valory.skills.trader_abci.handlers` — 100% (459 stmts, 94 branches, 0 missing)
- [x] `uv run tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — all green
- [x] `uv run autonomy packages lock` — hashes updated for `trader_abci` + `agent/trader` + `polymarket_trader` + `trader_pearl`
- [x] Live smoke test on Omenstrat EOA with `USE_X402=true` — paraswap-revert → deny-retry → successful swap (see above)

## LOC breakdown

| Category | File(s) | + | − |
|---|---|---:|---:|
| **Production fix** | `packages/valory/skills/trader_abci/handlers.py` | 91 | 24 |
| Tests | `packages/valory/skills/trader_abci/tests/test_handlers.py` | 330 | 15 |
| Hash + config bumps (auto-locked) | `packages.json`, `skill.yaml`, `aea-config.yaml`, 2× `service.yaml` | 9 | 9 |
| | **Total** | **430** | **48** |

The behavioral change is **net ~67 lines** of production code in one file (`handlers.py`). The remaining ~315 net lines are test coverage (6 new test cases + tuple-return updates to keep CI's 100% coverage gate green) and auto-generated hash bumps from `autonomy packages lock`.

### DIFF
--- a/packages/valory/skills/trader_abci/handlers.py
+++ b/packages/valory/skills/trader_abci/handlers.py
@@ -33,6 +33,7 @@ import requests
 from aea_ledger_ethereum.ethereum import EthereumCrypto
 from eth_account import Account
 from web3 import Web3
+from web3.exceptions import ContractLogicError
 
 from packages.valory.protocols.http.message import HttpMessage
 from packages.valory.skills.abstract_round_abci.handlers import (
@@ -123,6 +124,8 @@ TRADING_STRATEGY_EXPLANATION = {
 COINGECKO_RATE_CACHE_SECONDS = 7200  # 2 hours
 
 FALLBACK_POL_TO_USD_RATE = 0.089935  # AS of 2026-02-11T18:35:09Z
+
+X402_SWAP_MAX_ROUTE_RETRIES = 3
 
 
 class HttpHandler(BaseHttpHandler):
@@ -795,6 +798,7 @@ class HttpHandler(BaseHttpHandler):
         from_amount: Optional[str] = None,
         to_amount: Optional[str] = None,
         timeout: int = 30,
+        deny_exchanges: Optional[List[str]] = None,
     ) -> Optional[Dict]:
         """
         Get LiFi quote for token swap.
@@ -807,6 +811,7 @@ class HttpHandler(BaseHttpHandler):
         :param from_amount: Amount to swap from (for standard quote)
         :param to_amount: Desired amount to receive (for toAmount quote)
         :param timeout: Request timeout in seconds
+        :param deny_exchanges: LiFi tool names to exclude from routing
         :return: LiFi quote response or None if failed
         """
         try:
@@ -827,6 +832,8 @@ class HttpHandler(BaseHttpHandler):
                 "slippage": slippage,
                 "integrator": "valory",
             }
+            if deny_exchanges:
+                params["denyExchanges"] = ",".join(deny_exchanges)
 
             # Add amount parameter based on what's provided
             if to_amount is not None:
@@ -927,15 +934,24 @@ class HttpHandler(BaseHttpHandler):
         tx_request: Dict,
         eoa_address: str,
         chain: str,
-    ) -> Optional[int]:
-        """Estimate gas for a transaction"""
+    ) -> Tuple[Optional[int], bool]:
+        """Estimate gas for a transaction.
+
+        :param tx_request: LiFi `transactionRequest` dict with `to`, `data`,
+            and `value` keys.
+        :param eoa_address: EOA that would sign and send the transaction.
+        :param chain: Chain name selecting the RPC endpoint.
+        :return: (gas, route_revert). route_revert is True only when the
+            failure was a contract-side revert, in which case retrying the
+            LiFi quote with the offending route denied may help.
+        """
         try:
             w3 = self._get_web3_instance(chain)
             if not w3:
                 self.context.logger.error(
                     "Failed to get Web3 instance for gas estimation"
                 )
-                return None
+                return None, False
 
             tx_value = (
                 int(tx_request["value"], 16)
@@ -957,11 +973,14 @@ class HttpHandler(BaseHttpHandler):
             self.context.logger.info(
                 f"Estimated gas: {estimated_gas}, with 20% buffer: {tx_gas}"
             )
-            return tx_gas
+            return tx_gas, False
 
+        except ContractLogicError as e:
+            self.context.logger.error(f"Route revert in gas estimation: {e}")
+            return None, True
         except Exception as e:
             self.context.logger.error(f"Error in gas estimation: {str(e)}")
-            return None
+            return None, False
 
     def _submit_x402_swap_if_idle(self) -> None:
         """Submit x402 swap task only if no swap is currently in progress."""
@@ -1017,23 +1036,63 @@ class HttpHandler(BaseHttpHandler):
             )
 
             top_up_usdc_amount = str(top_up)
-            quote = self._get_lifi_quote(
-                from_token=chain_config["native_token_address"],
-                to_token=usdc_address,
-                from_address=eoa_address,
-                to_address=eoa_address,
-                chain_config=chain_config,
-                to_amount=top_up_usdc_amount,
-            )
-            if not quote:
-                self.context.logger.error("Failed to get LiFi quote")
+            denied: List[str] = []
+            tx_request: Optional[Dict] = None
+            tx_gas: Optional[int] = None
+
+            for attempt in range(X402_SWAP_MAX_ROUTE_RETRIES):
+                quote = self._get_lifi_quote(
+                    from_token=chain_config["native_token_address"],
+                    to_token=usdc_address,
+                    from_address=eoa_address,
+                    to_address=eoa_address,
+                    chain_config=chain_config,
+                    to_amount=top_up_usdc_amount,
+                    deny_exchanges=denied or None,
+                )
+                if not quote:
+                    self.context.logger.error("Failed to get LiFi quote")
+                    return False
+
+                tx_request = quote.get("transactionRequest")
+                if not tx_request:
+                    self.context.logger.error("No transactionRequest in quote")
+                    return False
+
+                tx_gas, route_revert = self._estimate_gas(
+                    tx_request, eoa_address, chain
+                )
+                if tx_gas is not None:
+                    break
+
+                if not route_revert:
+                    self.context.logger.error("Failed to estimate gas for transaction")
+                    return False
+
+                tool = quote.get("tool")
+                if not tool or tool in denied:
+                    self.context.logger.error(
+                        "LiFi returned an unusable route; cannot retry "
+                        f"(tool={tool!r}, denied={denied})"
+                    )
+                    return False
+
+                self.context.logger.warning(
+                    f"LiFi route via '{tool}' reverted on gas estimation; "
+                    f"retrying with denyExchanges={denied + [tool]} "
+                    f"(attempt {attempt + 1}/{X402_SWAP_MAX_ROUTE_RETRIES})"
+                )
+                denied.append(tool)
+            else:
+                self.context.logger.error(
+                    f"Exhausted LiFi route retries ({X402_SWAP_MAX_ROUTE_RETRIES}); "
+                    f"denied={denied}"
+                )
                 return False
 
-            tx_request: Optional[Dict] = quote.get("transactionRequest")
-            if not tx_request:
-                self.context.logger.error("No transactionRequest in quote")
-                return False
-
+            # tx_gas, tx_request now belong to a route that passed estimation.
+            # Fetch nonce/gas-price only after a working route is found so the
+            # nonce window stays small even if the loop ran multiple LiFi calls.
             nonce, gas_price = self._get_nonce_and_gas_web3(eoa_address, chain)
             if nonce is None or gas_price is None:
                 self.context.logger.error("Failed to get nonce or gas price")
@@ -1044,10 +1103,6 @@ class HttpHandler(BaseHttpHandler):
                 if isinstance(tx_request["value"], str)
                 else tx_request["value"]
             )
-            tx_gas = self._estimate_gas(tx_request, eoa_address, chain)
-            if tx_gas is None:
-                self.context.logger.error("Failed to estimate gas for transaction")
-                return False
 
             tx_data = {
                 "to": Web3.to_checksum_address(tx_request["to"]),
--- a/packages/valory/skills/trader_abci/handlers.py
+++ b/packages/valory/skills/trader_abci/handlers.py
@@ -125,7 +125,7 @@ COINGECKO_RATE_CACHE_SECONDS = 7200  # 2 hours
 
 FALLBACK_POL_TO_USD_RATE = 0.089935  # AS of 2026-02-11T18:35:09Z
 
-X402_SWAP_MAX_ROUTE_RETRIES = 3
+X402_SWAP_MAX_ROUTE_ATTEMPTS = 3
 
 
 class HttpHandler(BaseHttpHandler):
@@ -976,7 +976,11 @@ class HttpHandler(BaseHttpHandler):
             return tx_gas, False
 
         except ContractLogicError as e:
-            self.context.logger.error(f"Route revert in gas estimation: {e}")
+            # Demoted to warning: the caller's retry loop turns this into
+            # either a successful swap (after a route swap) or a single ERROR
+            # on "Exhausted LiFi route retries". ERROR-on-recovered-failure
+            # would false-fire alerting wired to ERROR-level logs.
+            self.context.logger.warning(f"Route revert in gas estimation: {e}")
             return None, True
         except Exception as e:
             self.context.logger.error(f"Error in gas estimation: {str(e)}")
@@ -1040,7 +1044,7 @@ class HttpHandler(BaseHttpHandler):
             tx_request: Optional[Dict] = None
             tx_gas: Optional[int] = None
 
-            for attempt in range(X402_SWAP_MAX_ROUTE_RETRIES):
+            for attempt in range(X402_SWAP_MAX_ROUTE_ATTEMPTS):
                 quote = self._get_lifi_quote(
                     from_token=chain_config["native_token_address"],
                     to_token=usdc_address,
@@ -1048,10 +1052,16 @@ class HttpHandler(BaseHttpHandler):
                     to_address=eoa_address,
                     chain_config=chain_config,
                     to_amount=top_up_usdc_amount,
-                    deny_exchanges=denied or None,
+                    # Snapshot per call so the assertion-time mock state in
+                    # tests reflects the value at call time.
+                    deny_exchanges=list(denied),
                 )
                 if not quote:
-                    self.context.logger.error("Failed to get LiFi quote")
+                    self.context.logger.error(
+                        "Failed to get LiFi quote "
+                        f"(attempt {attempt + 1}/"
+                        f"{X402_SWAP_MAX_ROUTE_ATTEMPTS}, denied={denied})"
+                    )
                     return False
 
                 tx_request = quote.get("transactionRequest")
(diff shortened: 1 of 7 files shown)


packages/valory/skills/trader_abci/handlers.py
```diff
@@ -1017,23 +1036,63 @@ def _ensure_sufficient_funds_for_x402_payments(self) -> bool:
...
+                    self.context.logger.error(
+                        "LiFi returned an unusable route; cannot retry "
+                        f"(tool={tool!r}, denied={denied})"
+                    )
+                    return False
+
+                self.context.logger.warning(
+                    f"LiFi route via '{tool}' reverted on gas estimation; "
+                    f"retrying with denyExchanges={denied + [tool]} "
+                    f"(attempt {attempt + 1}/{X402_SWAP_MAX_ROUTE_RETRIES})"
```
DIvyaNautiyal07: P2: This warning fires on every iteration that hits a revert, including the final one. On `attempt=2` (last iteration of `range(3)`), it logs `"retrying with denyExchanges=[…] (attempt 3/3)"` and then the loop body finishes, the `else:` clause runs, and the next line logged is `"Exhausted LiFi route retries…"`. The "retrying" claim is a lie on the last iteration.

Guard with `if attempt < X402_SWAP_MAX_ROUTE_RETRIES - 1:`, or restructure so the message describes the attempt that just failed without promising a retry that won't happen.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -124,6 +125,8 @@
 
 FALLBACK_POL_TO_USD_RATE = 0.089935  # AS of 2026-02-11T18:35:09Z
 
+X402_SWAP_MAX_ROUTE_RETRIES = 3
```
DIvyaNautiyal07: Nit: `range(X402_SWAP_MAX_ROUTE_RETRIES)` produces 3 LiFi attempts total (1 initial + 2 retries). The conventional reading of `MAX_RETRIES = 3` is "3 retries on top of the initial attempt" = 4 attempts. Either rename to `X402_SWAP_MAX_ROUTE_ATTEMPTS`, or change the loop bound to `range(X402_SWAP_MAX_ROUTE_RETRIES + 1)`.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -1017,23 +1036,63 @@ def _ensure_sufficient_funds_for_x402_payments(self) -> bool:
...
+
+            for attempt in range(X402_SWAP_MAX_ROUTE_RETRIES):
+                quote = self._get_lifi_quote(
+                    from_token=chain_config["native_token_address"],
+                    to_token=usdc_address,
+                    from_address=eoa_address,
+                    to_address=eoa_address,
+                    chain_config=chain_config,
+                    to_amount=top_up_usdc_amount,
+                    deny_exchanges=denied or None,
```
DIvyaNautiyal07: Nit: `denied or None` is redundant. `_get_lifi_quote` already guards with `if deny_exchanges:`, which treats `[]` and `None` identically. Just `deny_exchanges=denied` works.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -957,11 +973,14 @@ def _estimate_gas(
             self.context.logger.info(
                 f"Estimated gas: {estimated_gas}, with 20% buffer: {tx_gas}"
             )
-            return tx_gas
+            return tx_gas, False
 
+        except ContractLogicError as e:
+            self.context.logger.error(f"Route revert in gas estimation: {e}")
+            return None, True
```
OjusWiZard, line 984: `ContractLogicError` is raised for **any** on-chain revert, not just route-specific ones. Insufficient native balance for the swap value, a paused token, or any other global invariant failure will all set `route_revert=True` here and burn up to 3 LiFi quotes hitting the same revert on every attempt before exhausting. The PR description's premise ("revert that justifies a route retry") isn't actually verified at this layer.

Options:
- Inspect `str(e)` for known route-specific markers (`arithmetic`, `Too little received`, `INSUFFICIENT_OUTPUT_AMOUNT`, …) and return `(None, False)` for known non-retryable patterns like `insufficient balance`.
- Or: keep this behavior but log at WARNING instead of ERROR — the caller still ERRORs on exhaustion, so demoting this avoids spurious ERROR-level logs on every successful route retry (current sequence on a happy retry is `ERROR` → `WARNING` → `INFO`, which will false-fire any alerting wired to ERROR).

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -1017,23 +1036,63 @@ def _ensure_sufficient_funds_for_x402_payments(self) -> bool:
...
+                    from_token=chain_config["native_token_address"],
+                    to_token=usdc_address,
+                    from_address=eoa_address,
+                    to_address=eoa_address,
+                    chain_config=chain_config,
+                    to_amount=top_up_usdc_amount,
+                    deny_exchanges=denied or None,
+                )
+                if not quote:
+                    self.context.logger.error("Failed to get LiFi quote")
```
OjusWiZard: Two small things on this branch:

1. The log is identical whether this fires on attempt 0 (initial call, no denies) or attempt 2 (after two route reverts). A transient LiFi outage mid-retry is indistinguishable from an initial API failure. Consider `f"Failed to get LiFi quote (attempt {attempt + 1}/{X402_SWAP_MAX_ROUTE_RETRIES}, denied={denied})"`.
2. The existing tests only exercise this path with `_get_lifi_quote` returning `None` on attempt 0. The retry-attempt-None case (attempt 0 reverts → attempt 1 returns None) isn't covered.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -1017,23 +1036,63 @@ def _ensure_sufficient_funds_for_x402_payments(self) -> bool:
...
+                        f"(tool={tool!r}, denied={denied})"
+                    )
+                    return False
+
+                self.context.logger.warning(
+                    f"LiFi route via '{tool}' reverted on gas estimation; "
+                    f"retrying with denyExchanges={denied + [tool]} "
+                    f"(attempt {attempt + 1}/{X402_SWAP_MAX_ROUTE_RETRIES})"
+                )
+                denied.append(tool)
```
OjusWiZard: On the final iteration (`attempt == 2`), this logs `"retrying with denyExchanges=[...] (attempt 3/3)"` and then the `for/else` clause immediately fires `"Exhausted LiFi route retries"`. The two messages contradict each other in real time, which will be noisy during incident diagnosis. The `denied.append(tool)` on that last iteration is also a wasted mutation.

Guard the warning:
```python
if attempt + 1 < X402_SWAP_MAX_ROUTE_RETRIES:
    self.context.logger.warning(...)
denied = denied + [tool]
```

Also: `denied.append(...)` mutates in place. The repo coding style (`CLAUDE.md` → `coding-style.md`) prefers immutable patterns, and the log message on line 1082 already constructs `denied + [tool]` — using that same expression as the assignment keeps the displayed and stored values trivially consistent.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -1017,23 +1036,63 @@ def _ensure_sufficient_funds_for_x402_payments(self) -> bool:
...
+            else:
+                self.context.logger.error(
+                    f"Exhausted LiFi route retries ({X402_SWAP_MAX_ROUTE_RETRIES}); "
+                    f"denied={denied}"
+                )
                 return False
```
OjusWiZard, line 1103: Two small points:

1. `for/else` is a Python footgun — fires when the loop completes without `break`. Functionally correct, but an explicit post-loop `if tx_gas is None: return False` reads more obviously. Zero behaviour change.
2. `tx_gas: Optional[int]` and `tx_request: Optional[Dict]` (lines 1040–1041) are used unguarded at lines 1101 (`tx_request["value"]`) and 1111 (`tx_data["gas"] = tx_gas`). The runtime invariant holds because of the `break`-only-on-success / `else: return False` shape, but mypy with strict-optional (which `make code-checks` enforces) will not narrow across this pattern. An `assert tx_gas is not None and tx_request is not None` right after the loop — or restructuring to early-return on success — gets you clean narrowing.

packages/valory/skills/trader_abci/tests/test_handlers.py
```diff
@@ -2297,6 +2362,180 @@ def test_outer_exception(self) -> None:
...
+            patch.object(self.handler, "_check_transaction_status", return_value=True),
+            patch("packages.valory.skills.trader_abci.handlers.Web3") as MockWeb3,
+        ):
+            MockWeb3.to_checksum_address = lambda addr: addr
+            result = self.handler._ensure_sufficient_funds_for_x402_payments()
+
+            assert result is True
+            assert mock_quote.call_count == 2
+            assert mock_quote.call_args_list[0].kwargs["deny_exchanges"] is None
+            assert mock_quote.call_args_list[1].kwargs["deny_exchanges"] == ["paraswap"]
```
OjusWiZard, line 2414: `MagicMock` stores a **reference** to the live `denied` list, not a snapshot per call. This assertion reads `denied`'s state at assertion time, not at the call site. It passes today because the loop exits on attempt 1 and no further `.append` happens, but:

- If a 2-consecutive-revert-then-success test is added (the missing boundary case — current coverage only exercises the 1-retry path), this assertion pattern would mask off-by-one accumulation bugs.
- The cleanest fix is at the production site: pass a copy (`deny_exchanges=list(denied) or None`). Otherwise snapshot here, e.g. `[call.kwargs["deny_exchanges"] and list(call.kwargs["deny_exchanges"]) for call in mock_quote.call_args_list]`.

Worth adding the `["tool_0"] → ["tool_0", "tool_1"]` 2-retry-then-success test in the same patch — that's the boundary case closest to the cap.

OjusWiZard (commented):
Inline notes on a handful of points worth tightening before merge. Logic is sound; these are mostly about observability, type-level invariants, and one real edge case (non-route reverts). Nothing critical.

packages/valory/skills/trader_abci/tests/test_handlers.py
```diff
@@ -2297,6 +2362,180 @@ def test_outer_exception(self) -> None:
...
+            patch.object(self.handler, "_estimate_gas", return_value=(None, True)),
+            patch.object(
+                self.handler, "_get_nonce_and_gas_web3", return_value=(5, 1000)
+            ) as mock_nonce,
+        ):
+            result = self.handler._ensure_sufficient_funds_for_x402_payments()
+
+            assert result is False
+            assert mock_quote.call_count == 3
+            mock_nonce.assert_not_called()
```
OjusWiZard, line 2449: This verifies the outcome (`False`, 3 quote calls, no nonce fetch) but not that the **exhaustion log** fired. A mutation that replaces the `for/else` exhaustion block in `handlers.py` with an inline `return False` right after `denied.append(tool)` would pass this test silently. Adding `self.handler.context.logger.error.assert_called_with("Exhausted LiFi route retries (3); denied=['tool_0', 'tool_1', 'tool_2']")` (or a partial-match form) would pin the for/else branch.

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: Fixed in 04f421a5 — guarded the warning behind `if attempt + 1 < X402_SWAP_MAX_ROUTE_ATTEMPTS:` so the misleading "retrying" line no longer fires on the final iteration. (Same point as Ojus's L1085 comment.)

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: Renamed in 04f421a5 to `X402_SWAP_MAX_ROUTE_ATTEMPTS = 3` (kept the 3-attempt cap; the name now matches what `range(N)` produces).

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: Confirmed — `_get_lifi_quote` does `if deny_exchanges:` so `[]` and `None` are equivalent. Dropped the `or None` in 04f421a5. (Now passes `list(denied)` instead per Ojus's snapshot-vs-reference point below.)

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: Good points. Two things on the scope:

1. `web3.eth.estimate_gas` in 7.x raises **`ValueError`** ("insufficient funds for gas * price + value"), not `ContractLogicError`, for the insufficient-native case — so that one is already filtered by exception type and won't burn 3 retries.
2. The realistic `ContractLogicError` causes that *do* fire here are slippage breaches / expired LiFi deadlines (re-quoting actually fixes both — the deadline and route get refreshed) and paused-token / blacklist reverts (rare; bounded at 3 attempts ≈ 5-8s in the worker, which the `_x402_swap_future` guard already debounces).

String-sniffing `str(e)` for known markers is fragile (LiFi/DEX revert strings drift across upgrades), so going with your second suggestion: 04f421a5 demotes this log from ERROR to WARNING. The caller still ERRORs on `Exhausted LiFi route attempts`, so true outages still escalate.

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: Fixed in 04f421a5 — the LiFi-quote-None error now reads `Failed to get LiFi quote (attempt N/MAX, denied=[...])`, so a transient API failure mid-retry is distinguishable from one on the initial call.

On the test gap: added `test_route_revert_two_retries_then_success` covering the 2-deny accumulation path (also addresses your snapshot point on L2414). A dedicated retry-attempt-returns-None test (revert on attempt 0 → `None` on attempt 1) is one off-by-one further; happy to add it if you want it explicitly pinned, just say the word.

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: Both fixed in 04f421a5 — guarded the warning behind `if attempt + 1 < X402_SWAP_MAX_ROUTE_ATTEMPTS` (Divya flagged the same thing) and switched to `denied = denied + [tool]` immutable assignment so the displayed value matches what gets stored.

packages/valory/skills/trader_abci/handlers.py
LOCKhart07: On the mypy concern: all six lint targets (incl. mypy) pass green — the repo's mypy config doesn't enable strict-optional cross-loop narrowing, so this isn't a current bug. On the readability concern: confirmed via AST scan that the wider `packages/` tree has exactly one other `for/else` use (`omen_withdraw.py:345`), so the pattern is rare in this codebase. After discussion, code-owner call is to keep `for/else` for compactness; if it bites future readers we can swap to an explicit success-flag pattern in a follow-up.

packages/valory/skills/trader_abci/tests/test_handlers.py
LOCKhart07: Fixed at the production site in 04f421a5 — `_get_lifi_quote(... deny_exchanges=list(denied))` so each call gets its own snapshot rather than a live reference to the loop-local list.

Also added `test_route_revert_two_retries_then_success` exercising the 2-deny boundary (`[] → ['paraswap'] → ['paraswap', 'sushiswap']`), which would catch off-by-one accumulation regressions that the existing 1-retry test wouldn't.

packages/valory/skills/trader_abci/tests/test_handlers.py
LOCKhart07: Added in 04f421a5 — `test_route_revert_exhausted` now scans `logger.error.call_args_list` for the substring `Exhausted LiFi route attempts` together with `['tool_0', 'tool_1', 'tool_2']`. A mutation replacing the `for/else` with an inline `return False` would skip the exhaustion-log call and fail this assertion.

bennyjo (dismissed):
Reviewed end-to-end — tight, well-scoped fix. Retry loop is bounded, discriminates revert vs non-revert, nonce window correctly moved out of the loop, and test coverage exercises every branch (incl. log-pinned exhaustion + snapshot-per-call deny list). No security/attacker surface change. LGTM.

DIvyaNautiyal07 (dismissed):
All three review comments addressed. The misleading "retrying" log on the final iteration is now guarded, the constant is renamed to `X402_SWAP_MAX_ROUTE_ATTEMPTS`, and `deny_exchanges=list(denied)` snapshots per call. Bonus improvements (WARNING demotion on `ContractLogicError`, immutable deny rebind, attempt/denied context in the quote-failure log, new 2-deny accumulation test, log-assertion pinning the for/else exhaustion branch) are all sensible. LGTM.

OjusWiZard (dismissed):

DIvyaNautiyal07 (approved):

bennyjo (approved):
re-approval

Issue #938, opened 2026-05-03
Resilience audit: P0 genai unprotected json.loads + ClobClient/RelayClient indefinite hangs


DavidMinarsch:
# External Request Resilience Audit

Deep analysis of every external HTTP/RPC dependency in the in-repo packages, what happens under each failure mode (HTTP errors, unreachable, malformed data, timeouts), and how failures propagate through the FSM and connection threading model.

**Scope:** `packages/valory/skills/{agent_performance_summary_abci, chatui_abci, check_stop_trading_abci, decision_maker_abci, market_manager_abci, staking_abci, trader_abci, tx_settlement_multiplexer_abci}`, `packages/valory/connections/{polymarket_client, genai, x402}`, `packages/valory/customs/{fixed_bet, kelly_criterion}`. Framework-synced packages (`abstract_round_abci`, `transaction_settlement_abci`, `mech_interact_abci`, `registration_abci`, `reset_pause_abci`, `termination_abci`, `funds_manager`, `abci`, `http_client`, `http_server`, `ipfs`, `ledger`) are out of scope.

---

## How the framework handles HTTP in this service

Three stacks are in active use:

1. **Framework path** (`yield from self.get_http_response(...)`) — used by all behaviours that fetch subgraph / API data. Network errors are converted to `status_code=600` with the traceback in `body`. JSON parse / key-shape errors bubble up to `ApiSpecs.process_response` which returns `None`. `_handle_response` retries with `backoff_factor^retries_attempted` cooperative `yield from self.sleep()`.
2. **Direct `requests.*` path** — used by:
   - `agent_performance_summary_abci/graph_tooling/predictions_helper.py` and `polymarket_predictions_helper.py` (10+ POST calls with `timeout=30`).
   - `polymarket_client/connection.py:_request_with_retries` (Gamma API GET with `timeout=10`, 3-attempt linear backoff).
   - `trader_abci/handlers.py:567` (CoinGecko POL/USD, `timeout=10`) and `:834` (LiFi quote, configurable timeout).
3. **Third-party library wrappers**:
   - `py_clob_client_v2.ClobClient` → `httpx.Client(http2=True)` (no timeout configured).
   - `py_builder_relayer_client.RelayClient` → `requests` (no timeout configured).
   - `google.generativeai.GenerativeModel` (timeout unknown).
   - `x402_requests` / `x402HttpxClient` adapters (no timeout configured at session creation).
[...]

PR #957, opened 2026-05-13
fix(resilience-audit): R-3 polymarket_client backoff + R-9 last_updated in perf API

LOCKhart07:
> **Stacked on #951** (`chore/bump-genai-fundsmgr`). Once #951 merges, GitHub auto-rebases this PR onto `main`.

## Summary

Two fixes from the [External Request Resilience Audit (#938)](https://github.com/valory-xyz/trader/issues/938):

- **R-3 (P1)** `polymarket_client/connection.py` — `_request_with_retries` used `time.sleep(RETRY_DELAY * (attempt + 1))` despite a comment claiming exponential backoff. Worst-case blocked the single connection worker for ~60s. Now uses true exponential `RETRY_DELAY * (2**attempt)` with `RETRY_DELAY = 1`, capping the total retry budget at ~3s for `MAX_API_RETRIES=3`. Already on the branch as `0aa7b1fba`.
- **R-9 (P3)** `agent_performance_summary_abci/handlers.py` — the three performance HTTP endpoints (`/agent/performance`, `/agent/prediction-history`, `/agent/profit-over-time`) drop the generation timestamp before serialization. The UI cannot distinguish fresh data from preserved-stale snapshots (the FSM's stale-data preservation pattern keeps serving last-known-good metrics through subgraph outages). This change adds a `last_updated` field (ISO 8601 UTC or `null`) to each response. The source timestamps already exist on `AgentPerformanceSummary.timestamp`, `PredictionHistory.last_updated`, `ProfitOverTimeData.last_updated` — we just stop throwing them away.

## What is NOT in this PR

| Audit finding | Status |
|---|---|
| R-1 (genai unprotected `json.loads`) | Fixed upstream in valory-xyz/genai#30; lands here via #951 |
| R-2 (ClobClient/RelayClient indefinite hang) | Deferred — requires per-call `ThreadPoolExecutor.submit(...).result(timeout=…)` wrapping or upstream PRs against `py_clob_client_v2`/`py_builder_relayer_client` |
| R-4 (x402 session no timeout) | Fixed upstream in valory-xyz/genai#30; lands here via #951 |
| R-5 (stale `FALLBACK_POL_TO_USD_RATE`) | Rejected — fallback is already third-tier (live CoinGecko → 2h cache → constant); refusing it would convert a CoinGecko outage during cold-start from a wrong-rate degradation into a hard crash |
| R-6 (sync `requests.get` in handlers) | Deferred — only `_get_pol_to_usdc_rate` is the real issue (LiFi is already on `self.executor`); still touches the handler/executor boundary in several places |
| R-7 (inconsistent `or {}` guards in predictions helpers) | Already in current code |
| R-8, R-10 | P3 systemic; need their own follow-up tickets |

## UI follow-up

`last_updated` is additive — `agent-ui-monorepo`'s `predict-ui` consumers (`useAgentDetails`, `useTradeHistory`, `useProfitOverTime`) all destructure known fields only, no schema validator, no `additionalProperties: false`. The new field is silently ignored until the UI repo adds a staleness indicator.

UI tracking issue: [valory-xyz/agent-ui-monorepo#106](https://github.com/valory-xyz/agent-ui-monorepo/issues/106)

## Test plan

- [x] `uv run pytest packages/valory/skills/agent_performance_summary_abci/tests/test_handlers.py` — 143 pass, 5 new
- [x] 100% coverage on `agent_performance_summary_abci/handlers.py`
- [x] `uv run tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — all green
- [x] `autonomy packages lock`

## Diff breakdown

| Category | Files | +lines | -lines | Notes |
|---|---|---:|---:|---|
| Production code | 2 | 19 | 3 | `polymarket_client/connection.py` (R-3, 2/2), `agent_performance_summary_abci/handlers.py` (R-9, 17/1) |
| Tests | 2 | 126 | 0 | `tests/test_connection.py` (+36, R-3 backoff tests), `tests/test_handlers.py` (+90, 5 R-9 tests) |
| Package hashes & lockfile | 12 | 40 | 40 | `packages.json` + 10 dependent `skill.yaml` / `service.yaml` / `aea-config.yaml` / `connection.yaml` rehashes from `autonomy packages lock` — zero behavioural change |
| **Total** | **16** | **185** | **43** | |

Ignoring lockfile churn, the real surface area is **~145 lines added across 4 files** — most of it tests.


packages/valory/connections/polymarket_client/connection.py
```diff
@@ -76,7 +76,7 @@
 MARKETS_TIME_WINDOW_DAYS = 4
 API_REQUEST_TIMEOUT = 10
 MAX_API_RETRIES = 3
-RETRY_DELAY = 10
+RETRY_DELAY = 1
```
jmoreira-valory: **Dropping `RETRY_DELAY` 10× (10 → 1) removes rate-limit protection on 429 responses.**

The math fix (linear → exponential) is correct, but the base-delay reduction isn't justified in the PR description. Looking at `_request_with_retries` (L674):

```python
response.raise_for_status()       # raises HTTPError on 4xx/5xx, including 429
return response.json(), None
except (requests.exceptions.RequestException, ValueError) as e:
    ...
    time.sleep(RETRY_DELAY * (2**attempt))
```

`HTTPError` is a `RequestException` subclass, so the retry loop catches and re-issues against the same endpoint on **429 Too Many Requests** too. With the new base, a rate-limited Polymarket Gamma / CLOB endpoint gets hammered at 1s → 2s → 4s (cumulative ~7s) where the previous code paused 10s → 20s → 30s (~60s). The "single connection worker blocked for ~60s" framing in the PR description was the rate-limit safety margin, not a defect.

Three cleaner options:

1. **Keep the 10s base, just fix the math**:
   ```python
   RETRY_DELAY = 10
   time.sleep(RETRY_DELAY * (2**attempt))   # 10s, 20s, 40s
   ```
   Cumulative ~70s worst-case, but respects rate limits.

2. **Branch on `e.response.status_code == 429`** for a longer rate-limit-specific backoff (e.g., 10s base) while keeping 1s for transient connection errors:
   ```python
   base = 10 if isinstance(e, requests.exceptions.HTTPError) and getattr(e.response, "status_code", None) == 429 else 1
   time.sleep(base * (2**attempt))
   ```

3. **Add jitter** to avoid synchronized retry storms across multiple instances hitting the same Polymarket endpoint:
   ```python
   time.sleep(RETRY_DELAY * (2**attempt) + random.uniform(0, 1))
   ```

(2) is the strictly-correct shape; (1) is the smallest change that fixes both the math AND keeps the rate-limit safety margin; (3) is additive to any of the above.

Worst case if a Polymarket endpoint genuinely rate-limits the agent now: the 1s base burns through all 3 attempts in ~7s, sees 3 consecutive 429s, returns `(None, last_error)`, and the caller treats it as a hard failure of the lookup — when waiting 60s would have succeeded on the next attempt.

packages/valory/skills/agent_performance_summary_abci/handlers.py
```diff
@@ -529,6 +540,7 @@ def _handle_get_predictions(
                 "page_size": page_size,
                 "total": result["total_predictions"],
                 "items": result["items"],
+                "last_updated": None,
```
jmoreira-valory, line 552: **`last_updated: None` conflates three distinct staleness states — UI can't disambiguate.**

Three code paths in this handler produce `last_updated: None`:

- L463 — `summary.timestamp is None` (initial state, agent never fetched once)
- L543 (here) — subgraph fallback path, hardcoded `None` when no history available
- L631 — `profit_data is None` (no profit data state)

From the UI's perspective, all three look identical. But they mean very different things:

1. **"Never fetched"** — agent just started, no signal at all yet
2. **"Fetch failed this cycle, serving last-known-good"** — agent has cached data but couldn't refresh
3. **"Data is genuinely empty"** — fetch succeeded, returned no rows

The PR description acknowledges the FSM's stale-data preservation pattern (*"keeps serving last-known-good metrics through subgraph outages"*), but `last_updated: None` doesn't tell the UI when it's in mode 2 vs mode 1 vs mode 3. The whole motivation for R-9 was *"The UI cannot distinguish fresh data from preserved-stale snapshots"* — but the current shape only solves the happy path.

Three ways to disambiguate (any one works, in order of fidelity):

1. **Add a separate `fetch_attempted_at` alongside `last_updated`** — gives the UI both "last good data timestamp" and "last attempt timestamp"
2. **Stamp `last_updated` with wall-clock time on the fallback paths** — so mode 2 shows "as of T (with no fresh data)" rather than `None`
3. **Document the three meanings of `None` in the API contract / UI tracking issue [#106](https://github.com/valory-xyz/agent-ui-monorepo/issues/106)** so the UI's staleness indicator handles all three cases explicitly

(1) is the cleanest; (2) is the smallest change that improves fidelity. Non-blocking — additive, but worth doing before the UI ships its staleness indicator so it doesn't have to be re-jigged.

packages/valory/skills/agent_performance_summary_abci/handlers.py
```diff
@@ -610,11 +622,14 @@ def _handle_get_profit_over_time(
 
             if not profit_data or not profit_data.data_points:
                 # Return empty response if no data
-                empty_response = {
+                empty_response: Dict[str, Any] = {
                     "agent_id": safe_address,
                     "currency": "USD",
                     "window": window,
                     "points": [],
+                    "last_updated": self._format_last_updated(
```
OjusWiZard, line 639: `_format_last_updated` only short-circuits on `None`. `ProfitOverTimeData.last_updated` is non-Optional `int` in `models.py:141` and has no default, so a `0` sentinel is structurally constructable — and the existing `test_profit_data_with_no_data_points_returns_empty` actually builds `ProfitOverTimeData(last_updated=0, ...)`. Post-PR that response now contains `"last_updated": "1970-01-01T00:00:00Z"` instead of `null`, which the UI staleness indicator would render as a real cache write.

Suggest tightening the helper:

```python
if timestamp is None or timestamp <= 0:
    return None
```

Alternatively, drop the `0` from the existing empty-branch test and assert on a real ISO value. Either way, please make the contract explicit.

packages/valory/skills/agent_performance_summary_abci/tests/test_handlers.py
```diff
@@ -1532,6 +1605,23 @@ def test_response_includes_agent_id(self) -> None:
             assert response["agent_id"] == "0xabc123"
             assert response["currency"] == "USD"
 
+    def test_response_includes_last_updated_iso(self) -> None:
```
OjusWiZard, line 1614: Nice coverage for the populated path. The two empty-points branches at `test_no_profit_data_returns_empty` (L1541) and `test_profit_data_with_no_data_points_returns_empty` (L1557) also touch the new `last_updated` code at `handlers.py:625-633` but neither asserts on the field. A regression that flipped the ternary (e.g. `profit_data.last_updated if not profit_data else None`) would pass the full suite silently.

Worth adding `assert response["last_updated"] is None` to both. Tied to the helper-guard comment on `handlers.py:630`.

packages/valory/connections/polymarket_client/tests/test_connection.py
```diff
@@ -1005,6 +1006,41 @@ def test_passes_params_to_request(self) -> None:
...
 
+    def test_backoff_is_exponential_not_linear(self) -> None:
+        """Sleep durations follow exponential backoff (RETRY_DELAY * 2**attempt).
+
+        The previous implementation used a linear `RETRY_DELAY * (attempt + 1)`
+        despite a comment claiming exponential. We probe with max_retries=4 so
+        the third sleep distinguishes exponential (4x) from linear (3x).
+        """
+        conn = _make_connection()
+        with patch("requests.get") as mock_get, patch("time.sleep") as mock_sleep:
```
OjusWiZard: Minor: `patch("time.sleep")` works today only because production uses attribute lookup (`time.sleep(...)`). The `JSONDecodeError` test at L2684 in this same file already uses the module-local target — worth matching that pattern so a future `from time import sleep` refactor doesn't silently bypass the patch:

```python
with patch("packages.valory.connections.polymarket_client.connection.time.sleep") as mock_sleep:
```

Applies to both new tests (L1017 and L1037).

OjusWiZard (commented):
Inline notes from the multi-perspective review. Production logic is sound; these are a real test/behavior gap plus a small flake-risk fix.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: Good catch. Went with your option 2 in 1128b79cc — added `RATE_LIMIT_RETRY_DELAY = 10` and the retry loop now branches on `isinstance(e, HTTPError) and e.response.status_code == 429` to pick the base delay:

- 429 → `RATE_LIMIT_RETRY_DELAY * (2**attempt)` = 10s, 20s (cumulative ~30s for `MAX_API_RETRIES=3`)
- Transient (`ConnectionError`, `Timeout`, etc.) → `RETRY_DELAY * (2**attempt)` = 1s, 2s (cumulative ~3s)

Covered with two new tests (`test_backoff_uses_rate_limit_base_on_429` and `test_backoff_uses_transient_base_on_connection_error`).

Did not add jitter — happy to layer it in a follow-up if you'd like, but with `MAX_WORKER_THREADS=1` per agent process and a single Polymarket Safe per agent, the synchronized-retry-storm scenario doesn't really apply here. Open to being wrong if you've seen evidence otherwise.

packages/valory/skills/agent_performance_summary_abci/handlers.py
LOCKhart07: Pushed back on this internally and settled on a different framing — leaving the code as-is. Reasoning:

The three `None`-emitting paths line up with the data they accompany:

| Path | Payload | `last_updated` |
|---|---|---|
| Performance, `summary.timestamp is None` | Handler returns **HTTP 500** (no payload at all) | n/a — never reaches the client in practice |
| Predictions subgraph fallback (L543) | Real, fresh items from a live subgraph fetch | `null` |
| Profit empty (L631) | `points: []` | `null` |

The contract we're shipping is **`null` = no timestamp known; freshness ≠ liveness**. The UI's job is:

- `last_updated == null` → don't render a staleness badge.
- `last_updated != null` → render "Updated X ago" and decide stale vs fresh by comparing to `now()`.

Under that contract:
- Case 1 never surfaces (500 short-circuits before the helper runs).
- Case 3 is consistent — empty payload + no badge = "no data".
- Case 2 is the genuinely odd one: real fresh data + no badge. UI shows the predictions, just without an "updated just now" tag. Not wrong, just less rich than the populated-stored-history path.

The deeper concern you flagged — distinguishing "agent stuck" from "agent trying" — is real but it's a **liveness** signal, not a freshness signal. Adding `fetch_attempted_at` would solve it cleanly, but it needs FSM-side work (the summary behaviour currently writes `timestamp` only on successful runs; we'd need to record every attempt). Filing that as a follow-up rather than bundling it here.

Will update the UI tracking issue ([valory-xyz/agent-ui-monorepo#106](https://github.com/valory-xyz/agent-ui-monorepo/issues/106)) to document the `null = no data` contract so the staleness indicator handles it explicitly. Happy to revisit if the UI work surfaces a case where the badge absence is actively confusing.

packages/valory/skills/agent_performance_summary_abci/handlers.py
LOCKhart07: Good catch — fixed in e2be594ee. Tightened the guard to `timestamp is None or timestamp <= 0` and added the rationale in the docstring (sub-Unix-epoch values are structurally constructable but semantically a "never set" sentinel, especially given `ProfitOverTimeData.last_updated` has no default).

Kept the existing test's `last_updated=0` and added `assert response["last_updated"] is None` instead of dropping the 0 — that way the test now pins the exact behaviour you flagged.

packages/valory/skills/agent_performance_summary_abci/tests/test_handlers.py
LOCKhart07: Done in e2be594ee — `assert response["last_updated"] is None` added to both `test_no_profit_data_returns_empty` (L1555) and `test_profit_data_with_no_data_points_returns_empty` (L1577). The second one in particular now catches the 1970 regression from the helper-guard issue.

packages/valory/connections/polymarket_client/tests/test_connection.py
LOCKhart07: Done in e2be594ee — moved all four backoff tests' sleep patches to the module-local target (`packages.valory.connections.polymarket_client.connection.time.sleep`), matching the `JSONDecodeError` test's pattern. Includes the two pre-existing R-3 tests (`test_backoff_is_exponential_not_linear`, `test_backoff_total_budget_is_bounded`) in addition to the two new ones from the rate-limit fix.

jmoreira-valory (dismissed):
Both comments addressed:

- **#1 — rate-limit-aware backoff**: Option 2 implemented cleanly in `1128b79c`. `RATE_LIMIT_RETRY_DELAY = 10` + `RETRY_DELAY = 1` with `isinstance(e, HTTPError) and e.response.status_code == 429` picking the base. 30s cumulative budget for 429 vs 3s for transient. Two new tests (`test_backoff_uses_rate_limit_base_on_429`, `test_backoff_uses_transient_base_on_connection_error`) pin both paths. Jitter deferral is reasonable given `MAX_WORKER_THREADS=1` per agent — the synchronized-storm scenario doesn't apply at this scale.

- **#2 — `last_updated: None` API contract**: accepted the pushback. The framing — `null = no timestamp known; freshness ≠ liveness`, UI renders staleness badge only when populated — is coherent and the three code paths fit it consistently. The deeper agent-stuck-vs-trying concern is real but it's a liveness signal that needs FSM-side `fetch_attempted_at` work (recording every attempt, not just successful runs) — fair to scope as a follow-up. Documentation hook on [agent-ui-monorepo#106](https://github.com/valory-xyz/agent-ui-monorepo/issues/106) captures the contract for the UI staleness indicator work.

LGTM.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -677,10 +678,20 @@ def _request_with_retries(
...
+                        and e.response.status_code == 429
+                    )
+                    base_delay = (
+                        RATE_LIMIT_RETRY_DELAY if is_rate_limited else RETRY_DELAY
+                    )
                     self.logger.warning(
                         f"API request failed (attempt {attempt + 1}/{max_retries}): {e}. Retrying..."
                     )
-                    time.sleep(RETRY_DELAY * (attempt + 1))  # Exponential backoff
+                    time.sleep(base_delay * (2**attempt))  # Exponential backoff
```
OjusWiZard: Non-blocking, but worth a follow-up: this still ignores the `Retry-After` response header that servers send with 429s. RFC 9110 §10.6.4 lets a 429 carry `Retry-After: <seconds>` (Polymarket's Gamma/CLOB endpoints commonly do). If the server says `Retry-After: 60`, we retry after 10s, get rate-limited again, and burn both remaining attempts in 30s — exactly the failure mode this commit set out to prevent.

Minimal fix that handles the common numeric-seconds form (HTTP-date form can stay unhandled):

```python
delay = base_delay * (2**attempt)
if is_rate_limited:
    header = e.response.headers.get("Retry-After")
    if header is not None:
        with contextlib.suppress(ValueError):
            delay = max(delay, float(header))
time.sleep(delay)
```

Fine to defer to a follow-up PR if you'd rather keep this one scoped to the base-delay split.

OjusWiZard (dismissed):
LGTM on the follow-ups — guard for `<= 0` in `_format_last_updated`, module-local sleep patches, and the rate-limit-aware base delay all look right.

One non-blocking thought on the 429 path inline.

DIvyaNautiyal07 (dismissed):
Approving. Verified the fixes are correct and the test surface gives good mutation coverage:

**`connection.py`**
- `is_rate_limited` correctly identifies 429s via `isinstance(e, HTTPError) and e.response is not None and e.response.status_code == 429`. Guards the None-response edge.
- Exponential `base * 2**attempt` is correct, and the boolean's two branches are both exercised (429 test + ConnectionError test) — flipping the boolean would fail at least one of them.
- No cross-module consumers of `RETRY_DELAY` exist outside this file, so the constant rename/split has no ripple risk.

**`handlers.py`**
- `last_updated` is present in all return paths for the three endpoints — checked the empty/populated/stored/subgraph branches.
- `_format_last_updated`'s `<= 0` guard handles the JSON-restore edge case (`ProfitOverTimeData.last_updated` is non-Optional `int`); production callers in `behaviours.py:1330/1445/1750` always pass `synced_timestamp` so the sentinel is purely defensive — fine.

All P1 concerns from earlier rounds are addressed in `1128b79cc` and `e2be594ee`. No new findings.

packages/valory/connections/polymarket_client/connection.py
LOCKhart07: Good catch — implemented in 5cc975a4.

Empirical aside: I couldn't trigger a 429 against Gamma (500 reqs, 0 × 429), so the urgency was lower than I first read, but the change is defensive either way. Numeric-seconds `Retry-After` is honored as `max(exp_backoff, min(float(header), MAX_RATE_LIMIT_SLEEP))`, with `MAX_RATE_LIMIT_SLEEP = 60` capping pathological values. HTTP-date and unparseable forms fall through `contextlib.suppress(ValueError)` to the existing exponential.

4 new tests cover: header extends the sleep, header capped at max, absent-header is additive, invalid header falls back.

DIvyaNautiyal07 (approved):

OjusWiZard (approved):

PR #971, opened 2026-05-15
fix(kelly): drop 5x-ask venue-min floor that suppressed +EV CLOB bets

LOCKhart07:
## Problem

On Polystrat (`n_bets=1`, `max_bet=2.5` pUSD), the Kelly CLOB path set the grid floor to `b_min_side = max(min_bet, venue_min_side)` with `venue_min_side = 5 × best_ask`. When a profitable side was priced **> \$0.50**, `venue_min_side > max_bet`, so the `b_min = min(b_min, b_max)` clamp in `optimize_side` collapsed the grid to a single point at `b = max_bet`. With `n_bets=1`, `W_bet == max_bet`, so that point has `w_lose = W_bet − cost − fee ≤ 0` and is rejected — returning **no-bet on a clearly +EV side** (observed in production logs: `spend=0.0000` on a `+0.42` edge side; nothing ever submitted to the CLOB — zero `post_order` / `INVALID_ORDER_MIN_SIZE`).

The 5-share `min_order_size` is a **maker/limit-order** constraint. Trader places **FOK taker market orders**, which the CLOB does not size-check on share count (only USD depth). The venue-min floor was therefore both incorrect for the orders we place and the direct cause of the no-bet.

## Change

- Remove the `venue_min_side` computation and the insufficient-depth rejection.
- `b_min_side = min_bet` (the policy floor, ~1 pUSD — already ≥ Polymarket's real taker minimum, so orders stay executable; Kelly now sizes at its true optimum instead of a clamped point).
- `min_order_shares` left accepted-but-unused in `OPTIONAL_FIELDS` to keep the blast radius off `decision_receive` (can be cleaned up in a follow-up).

## Tests

- New `TestRunVenueMinClamp` regression reproducing the deployed config (`n_bets=1`, `max_bet=2.5`, `venue_min > max_bet`): **fails** before (`bet_amount == 0`), **passes** after.
- Removed `test_insufficient_depth_for_min_order_shares` (asserted the removed behavior — would green-wash); renamed the stale `..._in_venue_min_calc` test.
- 53 tests pass; **100 % coverage** on the module. black/isort/flake8/mypy/pylint clean (darglint failure is a pre-existing env crash on Py3.14, reproduces on untouched files).

## Note

This fixes a verified bet-suppression bug. It is **not** claimed to be the proven cause of the post-CLOB-v2 volume drop — that attribution remains open (see analysis shared separately).

## Empirical confirmation (on-chain, 2026-05-16)

The core assumption — that the 5-share `min_order_size` is a maker/limit constraint **not** enforced on the FOK *taker* market orders Trader places — is now confirmed **on-chain**, not just by code reading.

A live FOK taker BUY via the exact production path (`ClobClient(signature_type=2, funder=safe)` -> `create_market_order(FOK)` -> `post_order(FOK)`), deliberately sized **below** the venue minimum:

- Book at submission: `min_order_size = "5"`, `best_ask ~ 0.999`
- Order: **1.02 shares** for $1.00
- Result: **`status: "matched"`, `success: true`** - on-chain tx [`0x61c970cd...`](https://polygonscan.com/tx/0x61c970cdf166facaba82a7a3606d404902978cabdf983f12b0c8ff6f2aaf8cc7)

A 1.02-share order, far below the same book's `min_order_size=5`, filled. **Removing the `5x ask` floor cannot cause a venue rejection.** The earlier no-bets were 100% the internal Kelly clamp, never Polymarket.

---

## LOC breakdown (full PR vs `main`)

Every changed line classified from the `git diff --unified=0` (merge-base `063fb5fc`).

| Category | + | − | Net |
|---|---:|---:|---:|
| **Production code** (executable logic, `kelly_criterion.py`) | 19 | 22 | **−3** |
| Production inline comments | 12 | 1 | +11 |
| **Test code** (fixtures, assertions, `run()` calls) | 39 | 3 | +36 |
| Test docstrings | 37 | 2 | +35 |
| Test inline comments | 7 | 0 | +7 |
| Blank / formatting lines | 9 | 2 | +7 |
| Autogenerated hashes & metadata ¹ | 9 | 9 | 0 |
| **Total** | **132** | **39** | **+93** |

¹ Content-address churn only, 5 files: `packages.json` (4/4), `kelly_criterion/component.yaml` (2/2), `agents/trader/aea-config.yaml`, `services/polymarket_trader/service.yaml`, `services/trader_pearl/service.yaml` (1/1 each).

**Takeaway:** actual production logic is a **net −3 lines** — the fix deletes the 21-line venue-min depth loop and adds the smaller CLOB guard/marker. Of the +93 net, ~78 % is tests plus their docstrings (regression coverage for P1/P2 and the hardened assertions), not new production surface; the only autogenerated noise (18 lines) is content hashes.

---

## Design note: why not just hardcode the venue minimum to 0?

A reasonable question: couldn't the original intent have been met by a ~1-line change — hardcoding `min_order_shares = 0.0` (with a comment) — instead of deleting the venue-min block?

**Runtime: yes, identical.** With `min_order_shares = 0.0` the old block degenerates to `venue_min_side = 0.0`, the `remaining_shares > 0` rejection never fires (`0.0 > 0` is `False`), and `b_min_side = max(min_bet, 0.0) == min_bet` — byte-identical to this PR's `b_min_side = min_bet`. The core fix in `52c814a8` was reducible to a one-line default change.

**But it would not have been better:**

1. **It doesn't avoid the P1/P2 regression.** The thin-book sub-`min_bet` leak and the lost book-exhaustion diagnostic are caused by *lowering the floor to `min_bet`*, not by *how* it's lowered. On `main`, a book thinner than 5 shares hit `remaining_shares > 0 → reject`; that implicit protection disappears whether the block is deleted or neutered with `0`. Every guard and regression test added here would still be required — the LOC is regression hardening, orthogonal to delete-vs-hardcode.
2. **It leaves ~22 lines of live-looking dead code.** A depth-check loop with an `all_rejections` path that is silently inert because of a magic `0` is a worse maintenance trap than the dead-kwarg concern reviewers already flagged. Deletion removes it.
3. **It misstates the reason.** `min_order_shares = 0.0  # venue min is zero` reads as "the venue minimum is zero shares" — false. It is 5; it is simply not enforced because FOK *takers* are not share-count-constrained at all. The explanatory comment on the deleted block states the actual *why*, which is the load-bearing fact for the next reader.
4. **No diff-size win where it matters.** The `min_order_shares` kwarg becomes non-load-bearing under either approach, so the P3 marker/cleanup concern is identical; hardcoding buys nothing there.

The only defensible *minimal* variant would be `kwargs.get("min_order_shares", 0.0)` — changing the **default** while keeping the mechanism opt-in. That only makes sense if the venue-min is *sometimes* correct. The position here (and in review) is that it is *categorically* wrong for the FOK-taker path Trader uses, so deletion is the honest expression of intent; hardcoding-to-0 is a disguised delete that leaves dead code behind.


packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -337,32 +337,19 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
...
-                continue
 
-            b_min_side = max(min_bet, venue_min_side)
+            # No venue-minimum floor for CLOB. The per-market
+            # ``min_order_size`` (~5 shares) is a maker/limit-order
+            # constraint; Trader places FOK *taker* market orders, which the
+            # CLOB does not size-check on share count (only USD depth). The
+            # old ``b_min_side = max(min_bet, 5*best_ask)`` floor, combined
+            # with the ``b_min = min(b_min, b_max)`` clamp in optimize_side
+            # and the deployed ``n_bets=1`` (so ``W_bet == max_bet``),
```
OjusWiZard: Minor accuracy nit: `W_bet == max_bet` only holds when `bankroll - floor >= n_bets * max_bet` — see `w_bet = min(n_bets * max_bet, w)` at line 294. Under a low-bankroll edge case with `n_bets=1`, `W_bet < max_bet` and the grid does not collapse the same way.

Not a behavior issue (the fix is correct either way), but the parenthetical reads as an unconditional invariant in a load-bearing comment. Either qualify (`assuming bankroll ≥ max_bet`) or drop the parenthetical — the mechanism narrative survives without it.

Same applies to the class docstring at `test_kelly_criterion.py:660–663` ("W_bet collapses to max_bet").

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -337,32 +337,19 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
...
+            # collapsed the grid to a single point at ``b = max_bet`` and
+            # forced ``w_lose <= 0`` — suppressing profitable bets whenever a
+            # side was priced > $0.50. ``min_bet`` (the policy floor, ~1 pUSD)
+            # already exceeds Polymarket's real taker minimum.
+            b_min_side = min_bet
```
OjusWiZard, line 347: Behavioral regression worth a follow-up (not a blocker for this PR): the removed block used to append `"insufficient book depth to fill min_order_shares=…"` to `all_rejections` when the book was undepth-able. Now an empty/shallow CLOB book falls through `walk_book` returning `(0, 0)` → `optimize_side` skips every grid point via `cost <= 0` / `w_lose <= 0` (lines 185, 191) → `best_spend == 0` → side is silently dropped without touching `all_rejections`, and the terminal no-trade reason becomes the generic `"no bet improves log-growth"` at line 434.

Indistinguishable from a legitimately unprofitable market for anyone reading info logs. Cheapest fix is one `info.append(f"{label}: book exhausted / no fillable spend")` next to the `best_spend > 0 and g_improvement > 0` gate at line 414, or earlier when `walk_book(asks, b_max)` returns `(0, 0)`.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
```diff
@@ -641,3 +647,37 @@ def test_bet_capped_at_max_bet(self) -> None:
...
+            **CLOB_KWARGS,
+            "p_yes": 0.95,  # strong YES edge vs best ask 0.55
+            "max_bet": 2_500_000,  # 2.5 USDC — deployed Polystrat cap
+            "min_bet": 1_000_000,  # 1.0 USDC — absolute_min_bet_size
+            "n_bets": 1,  # deployed value: W_bet collapses to max_bet
+            "min_order_shares": 5.0,  # 5*0.55 = 2.75 > max_bet 2.5
+        }
+        result = run(**kwargs)
+        # Pre-fix (buggy): clamp → spend=2.5=W_bet → w_lose<=0 → bet_amount=0.
+        assert result["bet_amount"] > 0
```
OjusWiZard: `assert result["bet_amount"] > 0` is trivially passable — `DEFAULT_MIN_BET = 1` (1 wei) means even a degenerate near-zero bet would satisfy this and green-wash a future regression that returns a junk-sized bet instead of the expected one.

The deployed-config kwargs you set (`min_bet=1_000_000`) give you a real floor. Tighten to:
```python
assert result["bet_amount"] >= 1_000_000  # at least min_bet wei
```

Same weakness in `test_thin_book_below_old_min_order_shares_now_bets` at line 495 — but there, `min_bet` is the default 1 wei, so an explicit `>= some-meaningful-floor` (or asserting against the book's max fillable spend, `2 × 0.55 = 1.1 USDC ≈ 1_100_000` wei) would close the same gap.

OjusWiZard (dismissed):
Approving — the fix is arithmetically correct and the regression test pins the deployed Polystrat repro (`n_bets=1`, `max_bet=2.5`, `5×best_ask > max_bet` → grid collapses to one point → `w_lose ≤ 0`). On-chain confirmation in the PR body is convincing.

**One non-inline note** (line is outside the diff hunk so I can't anchor it): consider a one-line marker at `kelly_criterion.py:53` next to `"min_order_shares"` in `OPTIONAL_FIELDS`:

```python
        "min_order_shares",  # accepted-but-unused; see PR #971 — kept to avoid breaking decision_receive callers
```

The caller (`decision_maker_abci/behaviours/decision_receive.py:460–506` → `base.py:614`) still fetches `min_order_size` from both orderbook responses and forwards it through. Without a hint at the boundary, anyone tracing what enforces the venue minimum will believe the constraint is still live. Caller-side cleanup (including the dead `bet.min_order_shares = 5.0` fixture at `test_decision_receive.py:102`) is fine as the follow-up the PR description already calls out — just don't let the silent-accept fact go unmarked at the strategy boundary.

Three more notes inline. None are blockers.

packages/valory/customs/kelly_criterion/kelly_criterion.py
DIvyaNautiyal07, line 347: **P1 — sub-`min_bet` orders now leak through on thin books.**

`b_min_side = min_bet` sets the grid floor, but `optimize_side` returns `best_spend = cost` from `walk_book`, where `cost` is the actual fill cost capped by book depth, not the grid point `b`. If the book has less than `min_bet` worth of depth, the strategy returns a bet below the policy floor.

Reproduced: with `min_bet=1.0 USDC` and `asks=[{price:0.55, size:0.18}]` (total depth $0.099), pre-PR returns `bet_amount=0`, post-PR returns `bet_amount=99000` (0.099 USDC). The deleted depth check was the only thing guarding this corner — `walk_book` happily returns a partial fill, and nothing downstream re-checks against `min_bet`.

Fix: after `optimize_side`, reject when `best_spend < b_min_side`:
```python
if best_spend > 0 and best_spend < b_min_side:
    msg = f"{label}: fill cost {best_spend:.4f} < min_bet {b_min_side:.4f} (insufficient depth)"
    info.append(msg); all_rejections.append(msg); continue
```
This is the corner case the old `remaining_shares > 0` check was implicitly catching.

DIvyaNautiyal07 (commented):
Reviewed the diff plus the full call chain. The fix itself is correct and the regression test reproduces the deployed bug cleanly. Two concerns worth flagging as P1, plus inline P2/Nit comments.

**P1 (cross-file) — `min_order_shares` is dead through the full call chain.**

It's still in `OPTIONAL_FIELDS` (`kelly_criterion.py:53`), still fetched from the live orderbook in `decision_receive.py:460-485`, still passed through `base.py:614`, but never read in `run()`. The PR description acknowledges this is intentional to keep blast radius small. Risk is contract drift — the next reader sees the kwarg flow and reasonably assumes it gates sizing. Minimum ask: add a `# accepted for back-compat; not consulted post-PR-#971` comment next to the `OPTIONAL_FIELDS` entry so the deadness is greppable. Ideally drop it from `OPTIONAL_FIELDS` and stop populating it in `decision_receive.py` in a follow-up.

See the inline `b_min_side = min_bet` comment for the second P1 (sub-`min_bet` orders leaking through on thin books) — that one's a behavioral regression the deleted depth check was implicitly catching.

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -337,32 +337,19 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
...
+            # ``min_order_size`` (~5 shares) is a maker/limit-order
+            # constraint; Trader places FOK *taker* market orders, which the
+            # CLOB does not size-check on share count (only USD depth). The
+            # old ``b_min_side = max(min_bet, 5*best_ask)`` floor, combined
+            # with the ``b_min = min(b_min, b_max)`` clamp in optimize_side
+            # and the deployed ``n_bets=1`` (so ``W_bet == max_bet``),
+            # collapsed the grid to a single point at ``b = max_bet`` and
+            # forced ``w_lose <= 0`` — suppressing profitable bets whenever a
+            # side was priced > $0.50. ``min_bet`` (the policy floor, ~1 pUSD)
+            # already exceeds Polymarket's real taker minimum.
```
DIvyaNautiyal07: **P2 — lost rejection diagnostic for thin / all-zero-price books.**

When `walk_book` returns `(0, 0)` for every grid point (book too thin, all zero-price levels), the side falls through with `best_spend=0` and adds nothing to `all_rejections`. The no-trade message becomes the unrelated other-side rejection or `"no bet improves log-growth"`. Ops debugging a no-bet on a high-edge market loses the book-was-the-cause signal that the old `"insufficient book depth"` rejection provided.

Add an explicit rejection when `best_spend == 0` after `optimize_side`:
```python
if best_spend == 0:
    msg = f"{label}: book filled to 0 at all grid points (thin or invalid)"
    info.append(msg); all_rejections.append(msg)
```

packages/valory/customs/kelly_criterion/kelly_criterion.py
DIvyaNautiyal07: **Nit — `W_bet == max_bet` only when bankroll ≥ max_bet.**

`w_bet = min(n_bets * max_bet, w)` (line 294), so for small bankrolls `W_bet == w`, not `max_bet`. True for the deployed Polystrat config you're documenting, but stated here without caveat it'll mislead anyone whose bankroll is below `max_bet`. Either say `W_bet = min(max_bet, w)` or add "given w ≥ max_bet in the deployment".

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -337,32 +337,19 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
...
-                msg = (
-                    f"{label}: insufficient book depth to fill "
-                    f"min_order_shares={min_order_shares}"
-                )
-                info.append(msg)
-                all_rejections.append(msg)
-                continue
 
-            b_min_side = max(min_bet, venue_min_side)
+            # No venue-minimum floor for CLOB. The per-market
```
DIvyaNautiyal07: **Nit — 11-line historical comment is rot-prone.**

The old `b_min_side = max(...)` formula, the `b_min = min(b_min, b_max)` clamp chain, and the `n_bets=1` / `W_bet` interaction are bug archaeology that already lives in the commit message and PR description. Inline, two lines are enough — what's true now and why: "`min_order_size` is a maker/limit constraint, not enforced on FOK takers; see #971 for history." The rest will silently lie if `n_bets` default or optimizer logic changes.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
```diff
@@ -641,3 +647,37 @@ def test_bet_capped_at_max_bet(self) -> None:
...
+        assert result["vote"] == 0  # YES
+        assert result["g_improvement"] > 0
```
DIvyaNautiyal07: **P2 — `assert result["g_improvement"] > 0` is redundant.**

`run()` only sets `best_result` when `g_improvement > 0` (kelly_criterion.py:414). This assertion can never fail independently of `bet_amount > 0` above. Either drop it, or replace with a meaningful bound (e.g. `> 0.05`) to verify the edge is materially positive.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
DIvyaNautiyal07: **P2 — missing `bet_amount < max_bet` assertion.**

The whole point of the fix is that the grid finds an *interior* optimum instead of collapsing to `b = max_bet`. Asserting only `> 0` would still pass if a future bug bound the result to `max_bet` via a different path. Add `assert result["bet_amount"] < 2_500_000` to pin the interior-optimum behavior — that's the actual fix you want to regression-test.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
```diff
@@ -479,18 +479,24 @@ def test_empty_orderbook_returns_zero(self) -> None:
...
         kwargs = {
             **CLOB_KWARGS,
             "orderbook_asks_yes": [{"price": "0.55", "size": "2"}],
-            "min_order_shares": 10.0,  # need 10 but only 2 available
+            "min_order_shares": 10.0,  # no longer consulted
         }
         result = run(**kwargs)
-        assert result["bet_amount"] == 0
+        assert result["bet_amount"] > 0
+        assert result["vote"] == 0
```
DIvyaNautiyal07, line 498: **P2 — no upper bound on `bet_amount` in the thin-book test.**

Book depth is 2 shares × $0.55 = $1.10. Test asserts only `bet_amount > 0`. A future change that sized beyond available depth would pass silently. Add `assert result["bet_amount"] <= 1_100_000` to pin the fill-to-depth behavior the docstring claims is correct.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
```diff
@@ -479,18 +479,24 @@ def test_empty_orderbook_returns_zero(self) -> None:
         result = run(**kwargs)
         assert result["bet_amount"] == 0
 
-    def test_insufficient_depth_for_min_order_shares(self) -> None:
-        """Book depth < min_order_shares returns no trade."""
+    def test_thin_book_below_old_min_order_shares_now_bets(self) -> None:
+        """A thin book (no venue-min floor) sizes a small +EV taker bet.
+
+        Previously rejected by the removed ``min_order_shares`` depth check;
```
DIvyaNautiyal07: **Nit — "removed `min_order_shares` depth check" is slightly inaccurate.**

The kwarg `min_order_shares` is still accepted by `run()` (still in `OPTIONAL_FIELDS`), and this test still passes it on line 492. The accurate phrasing is "no longer consulted in the CLOB sizing path".

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
```diff
@@ -479,18 +479,24 @@ def test_empty_orderbook_returns_zero(self) -> None:
...
+        FOK taker orders are not share-count-constrained, so a small +EV bet
+        into the available 2 shares is correct.
+        """
         kwargs = {
             **CLOB_KWARGS,
             "orderbook_asks_yes": [{"price": "0.55", "size": "2"}],
-            "min_order_shares": 10.0,  # need 10 but only 2 available
+            "min_order_shares": 10.0,  # no longer consulted
```
DIvyaNautiyal07, line 492: **Nit — dead `min_order_shares` kwarg in test fixtures.**

Both this test and `test_profitable_side_not_suppressed_by_venue_min` (line 677) pass `min_order_shares` with a comment that it's ignored. If it's ignored, the kwarg is noise — drop it. (Tied to the top-level dead-`min_order_shares` concern in this review.)

jmoreira-valory (dismissed):

packages/valory/customs/kelly_criterion/kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P1). After `optimize_side`, the CLOB path now rejects `0 < best_spend < b_min_side` and appends `{label}: fill cost X < min_bet Y (insufficient book depth)` to `info` + `all_rejections`. New TDD regression `test_thin_book_below_min_bet_is_rejected` reproduces your exact case (YES depth $0.099, `min_bet` $1.0): pre-fix `bet_amount=99000`, post-fix `0`. Guard is CLOB-only so FPMM behavior/coverage is untouched.

packages/valory/customs/kelly_criterion/kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P2). An explicit rejection is now emitted when `walk_book(sorted_asks, max_bet)` returns 0 shares: `{label}: book filled to 0 at all grid points (thin or invalid)` into `info` + `all_rejections`. Keyed off real book exhaustion rather than `best_spend == 0`, so a liquid-but-unprofitable market keeps the generic "no bet improves log-growth" reason instead of being mislabeled a book problem. Covered by `test_book_exhausted_to_zero_emits_rejection`.

packages/valory/customs/kelly_criterion/kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P2). The side now appends an explicit `{label}: book filled to 0 at all grid points (thin or invalid)` to `info` + `all_rejections` when `walk_book(sorted_asks, max_bet)` yields 0 shares, so the terminal no-trade reason no longer degrades to the generic message on a high-edge market. I keyed it off genuine book exhaustion (the `walk_book(asks, b_max) == (0,0)` signal you suggested) rather than `best_spend == 0`, so a liquid-but-unprofitable market is not misreported as a book problem. New regression test `test_book_exhausted_to_zero_emits_rejection`.

packages/valory/customs/kelly_criterion/kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P5). The parenthetical is gone from the trimmed code comment, and the `TestRunVenueMinClamp` class docstring now states `W_bet = min(n_bets*max_bet, bankroll)` and notes it equals `max_bet` here only because bankroll 10 >= n_bets*max_bet 2.5 — no longer reads as an unconditional invariant.

LOCKhart07: Fixed in 2b7fb966 (P5) — the unconditional claim is gone; the test docstring now spells out `W_bet = min(n_bets*max_bet, bankroll)` with the bankroll >= max_bet caveat.

packages/valory/customs/kelly_criterion/kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P5). Collapsed to 3 lines stating only what is true now: `min_order_size` (~5 shares) is a maker/limit constraint, not enforced on the FOK *taker* orders Trader places (USD depth checked only); see PR #971 for history. The `n_bets`/`W_bet`/clamp archaeology now lives only in the commit + PR description.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P4). `test_profitable_side_not_suppressed_by_venue_min` now asserts `bet_amount >= 1_000_000` (>= min_bet, not junk-sized) plus `< 2_500_000` to pin the interior optimum. The thin-book test gained `bet_amount <= 1_100_000` (2 shares x $0.55 fill-to-depth bound), so a future change sizing beyond available depth would fail.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P4) — dropped the redundant `assert result["g_improvement"] > 0`. `run()` only sets `best_result` when `g_improvement > 0`, so it could never fail independently of the `bet_amount` bounds, which now pin the interior optimum.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P4) — added `assert result["bet_amount"] < 2_500_000` alongside `>= 1_000_000`, so the test now pins the interior optimum and would fail if a future bug bound the result back to `max_bet` via a different path.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P4) — added `assert result["bet_amount"] <= 1_100_000` (2 shares x $0.55 depth cap) to pin the fill-to-depth behavior the docstring claims.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Fixed in 2b7fb966 (P6) — docstring now reads "`min_order_shares` is still accepted but no longer consulted in the CLOB sizing path", which is the accurate phrasing since the kwarg remains in `OPTIONAL_FIELDS`.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Kept deliberately, tied to the top-level decision to keep `min_order_shares` accepted in `OPTIONAL_FIELDS` (full caller-side removal deferred to a follow-up to keep the blast radius off `decision_maker_abci`). While the kwarg is still accepted, passing it in these fixtures documents the accepted-but-ignored contract rather than being pure noise — and the strategy boundary now carries a greppable 3-line marker (2b7fb966, P3). It will be dropped from these fixtures together with the `decision_receive.py` + `test_decision_receive.py:102` cleanup in that follow-up.

LOCKhart07:
Thanks both — every inline and review-level point is addressed in 2b7fb966.

**Behavioral**
- **P1 (sub-`min_bet` leak):** CLOB now rejects `0 < best_spend < b_min_side` after `optimize_side` — `walk_book` caps cost to book depth, so a book thinner than `min_bet` was leaking a sub-floor order through once the venue-min floor was removed. New TDD regression reproduces the exact repro ($0.099 depth, `min_bet` $1.0): pre-fix `bet_amount=99000`, post-fix `0`.
- **P2 (lost diagnostic):** explicit `book filled to 0 at all grid points` rejection into `info` + `all_rejections`, keyed off genuine `walk_book(asks, max_bet) == (0,0)` exhaustion (per @OjusWiZard's suggestion) — *not* `best_spend == 0`, so a liquid-but-unprofitable market keeps the generic reason rather than being mislabeled a book problem.

Both guards are CLOB-only; FPMM behavior and coverage are untouched.

**Maintainability / tests**
- **P3:** per @OjusWiZard's non-inline note, the `min_order_shares` entry in `OPTIONAL_FIELDS` now carries a greppable 3-line marker ("accepted but NOT consulted in the CLOB sizing path post-PR #971"). Kept accepted; the `decision_receive.py:460-485` + `test_decision_receive.py:102` caller-side cleanup stays the deferred follow-up to keep the blast radius off `decision_maker_abci`.
- **P4:** assertions tightened — `>= min_bet`, `< max_bet` (pins the interior optimum), `<= depth` on the thin-book test, redundant `g_improvement > 0` dropped.
- **P5:** 11-line history comment trimmed to 3 lines (current truth + "see PR #971"); the unconditional `W_bet == max_bet` claim qualified to `min(n_bets*max_bet, bankroll)` with the bankroll caveat.
- **P6:** "removed depth check" wording corrected to "no longer consulted in the CLOB sizing path".

2 new regression tests (red→green verified); **55 tests pass, 100% line + branch coverage** on the module. Lint green (black/isort/flake8/mypy/pylint OK; darglint `code 129` is the known Py3.14 tool crash — reproduces on untouched files, as noted in the PR description).

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -424,6 +406,32 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
             f"G_improvement={g_improvement:.6f}"
         )
 
+        if market_type == "clob":
+            # The book genuinely cannot fill (empty / all zero-price /
+            # invalid): keep an explicit signal so a no-bet on a high-edge
+            # market is not mistaken for a legitimately unprofitable one.
+            _, depth_shares = walk_book(sorted_asks, max_bet)
```
OjusWiZard, line 413: Two small things tied to this guard. Both LOW; flagging for a follow-up, not blocking.

**Comment at lines 410–412 slightly overclaims.** It promises a no-bet on a high-edge market won't be mistaken for a legitimately unprofitable one. The guard only catches the structurally-empty case (`walk_book(sorted_asks, max_bet) → (0, 0)` — all-zero-price or absent levels). Two CLOB paths still drop silently through the `g_improvement > 0` gate at line 435 with nothing in `all_rejections`, so the terminal reason degrades to the generic `"no bet improves log-growth"`:

1. `best_spend == 0` while `depth_shares > 0` — `optimize_side` rejected every grid point internally on `cost <= 0`, `n_shares <= 0`, or `w_lose <= 0`. Reachable on the very Polystrat config this PR was written for (`n_bets=1`, `max_bet=2.5`, `w_bet ≈ max_bet`, so `w_lose = w_bet − cost − fee + extra` can pinch to ≤ 0 at the deeper grid points).
2. `best_spend > 0` and `g_improvement <= 0` — Kelly negative on a depth-OK fill.

One extra catch-all closes both without touching the optimizer:

```python
if best_spend == 0:
    msg = f"{label}: optimizer rejected all grid points (w_lose/w_win/g_improvement)"
    info.append(msg); all_rejections.append(msg)
    continue
```

**Redundant `walk_book` call.** `optimize_side` already calls `walk_book(asks, b)` at every grid point including `b = b_max` (line 167 of the same file clamps `b_min ≤ b_max`). If `walk_book(sorted_asks, max_bet)` returns `(0, 0)`, then `best_shares` in the returned tuple is also 0. The structurally-empty signal is already available — gating on `best_shares <= 0` from the existing return tuple gives the same diagnostic without the second O(n) sweep. Not a bug; pure tidy-up if you ever touch this block again.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
```diff
@@ -479,18 +479,66 @@ def test_empty_orderbook_returns_zero(self) -> None:
...
+        survive.
+        """
+        kwargs = {
+            **CLOB_KWARGS,
+            "orderbook_asks_yes": [{"price": "0", "size": "100"}],
         }
         result = run(**kwargs)
         assert result["bet_amount"] == 0
+        info_text = " ".join(result.get("info", []))
+        assert "yes: book filled to 0 at all grid points" in info_text
```
OjusWiZard, line 538: Nit — the test is one-directional with respect to the guard. If the new `depth_shares <= 0` rejection were removed, `bet_amount == 0` would still hold (`optimize_side` returns `(0, 0)` for an all-zero-price book regardless, then the `best_spend > 0 and g_improvement > 0` gate at line 435 of `kelly_criterion.py` filters it out via the generic fallback). The only thing actually catching a regression here is the substring match on `"yes: book filled to 0 at all grid points"`.

Not wrong — just lock in the guard's *precedence* over the generic fallback with one more line:

```python
assert "no bet improves log-growth" not in info_text
```

That makes the test fail if a future change drops the structural signal back to the generic terminal reason.

OjusWiZard (approved):
Re-approving on `2b7fb966` — all three of my prior threads are addressed and the new guards work as advertised. Two follow-up notes inline; neither is a blocker.

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -424,6 +406,32 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
...
+                    f"(thin or invalid)"
+                )
+                info.append(msg)
+                all_rejections.append(msg)
+                continue
+
+            # ``b_min_side`` only sets the requested grid point; ``walk_book``
+            # caps the actual fill cost to available depth, so a book thinner
+            # than ``min_bet`` would otherwise leak a sub-floor order through.
+            if 0 < best_spend < b_min_side:
```
DIvyaNautiyal07, line 426: Nit: strict `<` boundary isn't covered by tests. A future mutation `<` → `<=` would silently drop bets sized exactly at `min_bet` (a legitimate floor). Worth one quick test with a book sized to fill exactly `min_bet` (e.g. `price="1.0", size="1.0"` with `min_bet=1_000_000`) asserting `bet_amount > 0`.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
```diff
@@ -479,18 +479,66 @@ def test_empty_orderbook_returns_zero(self) -> None:
...
+
+        ``min_order_shares`` is still accepted but no longer consulted in the
+        CLOB sizing path; FOK taker orders are not share-count-constrained, so
+        a small +EV bet into the available 2 shares is correct.
+        """
         kwargs = {
             **CLOB_KWARGS,
             "orderbook_asks_yes": [{"price": "0.55", "size": "2"}],
-            "min_order_shares": 10.0,  # need 10 but only 2 available
+            "min_order_shares": 10.0,  # no longer consulted
```
DIvyaNautiyal07, line 492: Nit: this test passes `min_order_shares=10.0` but only asserts `bet_amount > 0`. A future refactor that re-introduced partial-fill consultation (without the hard rejection) would still pass — the depth-capped fill is $1.10, within the asserted `<= 1_100_000` ceiling. To make this a real "the kwarg is ignored" regression guard, run the same scenario twice with `min_order_shares=10.0` and `min_order_shares=2.0` and assert the two `bet_amount` results are identical.

DIvyaNautiyal07 (approved):

jmoreira-valory (approved):
P1 (sub-min_bet amounts leaking into CLOB orders) and P2 (lost diagnostic on the min-bet guard path) are both verified fixed in the current HEAD (9e79c500).

Four open threads from @OjusWiZard and @DIvyaNautiyal07 are acknowledged and non-blocking:

- `depth_shares > 0` / `best_spend == 0` silent fallthrough — valid edge-case note, follow-up ticket appropriate.
- Guard-precedence test assertion — style/coverage nit, no behavior impact.
- `<` vs `<=` boundary mutation — boundary semantics are defensible as-is; a follow-up test asserting the exact boundary is fine.
- Thin-book kwarg-ignored two-run proof — informational; no correctness issue in the current fix.

LGTM.

packages/valory/customs/kelly_criterion/kelly_criterion.py
LOCKhart07: Both valid — tracked as 1a + 1b in #974 (not addressed in this PR; non-blocking). To be precise on the decision vs. diagnostic split: the no-bet in the `best_spend==0, depth>0` (optimizer pinch) and `g_improvement<=0` paths is *correct* — what's missing is the explicit reason, plus the 410–412 comment overclaiming coverage. Noted in #974 that your catch-all uses a distinct message (so liquid-but-unprofitable isn't mislabeled "book filled to 0"), and that 1b is in tension with 1a — `best_shares==0` conflates structural-empty with optimizer-pinch, so the depth probe stays if we want to keep the two messages distinct.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Agreed — tracked as item 2 in #974 (non-blocking, not in this PR). For the record the test is already falsifiable (removing the guard breaks the `"yes: book filled to 0…"` substring, verified red→green in TDD); your extra `assert "no bet improves log-growth" not in info_text` adds *precedence* locking, which the follow-up will pick up.

packages/valory/customs/kelly_criterion/kelly_criterion.py
LOCKhart07: Valid boundary-coverage gap — tracked as item 3 in #974 (non-blocking, deferred). The current `<` is the intended operator (a fill exactly at `min_bet` should pass); the gap is mutation-detection only, and an exactly-`min_bet`-to-the-wei fill is measure-zero on real books, so no current defect. The follow-up will add the fill-exactly-`min_bet` test asserting `bet_amount > 0`.

packages/valory/customs/kelly_criterion/test_kelly_criterion.py
LOCKhart07: Good hardening — tracked as item 4 in #974 (non-blocking, deferred). `min_order_shares` is genuinely ignored today (not read anywhere in `run()`; the P3 marker documents it), so there's no current correctness gap; the two-run (`10.0` vs `2.0`, assert identical `bet_amount`) guard protects against a *future* refactor reintroducing partial-fill consultation. Will land with the #974 follow-up.

PR #972, opened 2026-05-18
fix(market-manager): guard balances lookup in update_bets to stop KeyError crash loop

LOCKhart07:
## Problem

Brand-new Omenstrat agents (fresh safes with no trading history) crash a few seconds after every start, in a permanent restart loop:

```
market_manager_abci/behaviours/update_bets.py", line 103, in update_bets_investments
    for outcome, value in balances[bet.id].items():
KeyError: '0xfe98e957a580edf8ecb3021bebe8d5d6c80c7107'
→ aea.exceptions.AEAActException ... UpdateBetsBehaviour
```

The agent dies in `update_bets_round` before it ever reaches `call_checkpoint_round`, so `is_staking_kpi_met` stays `false` permanently → **0 epoch rewards**, and the chat HTTP server (served by the agent) goes down with it. Established agents that already hold conditional-token positions are unaffected, which is why "old agents work, new ones don't."

## Root cause

A brand-new safe's `user` entity does not exist in the Conditional-Tokens subgraph yet, so it replies `{"data":{"user":null}}`. The framework's `process_response` cannot navigate `data:user:userPositions` through a `null` `user` → returns `None` → `_handle_response` returns `None` → `fetch_user_positions` returns `None` → `get_active_bets` early-returns a **plain `{}`** (not the `defaultdict(lambda: defaultdict(int))` its success path produces). The unguarded `balances[bet.id]` then raises `KeyError`.

This was **latent since the Aug-2025 investment-tracking feature** but only became fatal after `3f79888ca` (Omen withdrawal sweep, #952), which intentionally changed `fetch_user_positions` to return `None` instead of `[]` on the null/error path so the sweep can detect mid-pagination failures and not strand funds. Before that, the null path returned `[]` → `get_bet_id_to_balance` produced a `defaultdict` → the missing key auto-vivified harmlessly.

## Fix

One line, at the consumer:

```python
-            for outcome, value in balances[bet.id].items():
+            for outcome, value in balances.get(bet.id, {}).items():
```

- Reverting `fetch_user_positions` to return `[]` would re-break #952's failure detection, so the fix belongs here, not in `requests.py`.
- A missing/empty result now falls through to the existing `_replace_with_existing_investments_if_empty` recovery — prior investments retained, **no duplicate bets**. This is the failure behaviour that helper was explicitly written for; it was just never reached because line 103 crashed first.
- Happy path is behaviour-identical; the `fetch_user_positions` `None` failure signal stays intact for the withdrawal sweep and redeem.
- **Polystrat is unaffected** — it runs a separate `PolymarketFetchMarketBehaviour` path that sources balances from the Polymarket CLOB connection (not the CT subgraph) and already uses `trades_by_condition_outcome.get(condition_id, {})`. This fix effectively makes the Omen path do what the Polymarket path already does.

## Tests

Adds `test_balances_missing_bet_id_does_not_crash` — a falsifiable regression test that reproduces the exact production `KeyError` (fails on the old code) and asserts the recovery path retains prior investments. Full `test_behaviours_update_bets.py` suite passes (49 tests); 100% statement + branch coverage on `update_bets.py`. black/isort/flake8/mypy/pylint clean.

## LOC breakdown — fix vs. supporting

The actual behaviour change is **1 line**. Everything else is test, generated hashes, and prose.

| Category | Lines | Hand-written? | Notes |
|---|---:|---|---|
| **Actual fix** | **1** | yes | `update_bets.py`: `balances[bet.id]` → `balances.get(bet.id, {})` (1 add / 1 del) |
| Regression test — executable logic | ~11 | yes | setup + generator drive + assertion |
| Regression test — docstring | 12 | yes | explains the `{"data":{"user":null}}` → plain-`{}` mechanism |
| Regression test — comments / blank / signature | 8 | yes | 3 comment, 4 blank, 1 `def` |
| Hash propagation (`autonomy packages lock`) | 20 | **no — generated** | content-addressed bumps across 8 YAML/JSON files (`packages.json`, 5 `skill.yaml`, 2 `service.yaml`, `aea-config.yaml`) |

Totals: **52 insertions / 21 deletions**. Of the 52 added lines, **1** is the production fix, 31 are the new test (≈11 real logic), and 20 are mechanically generated dependency-hash bumps (zero hand-written). No comments or method descriptions were added to production code.


OjusWiZard (approved):

DIvyaNautiyal07 (approved):

jmoreira-valory:
**Post-merge independent verification** (delayed second reviewer)

Root cause chain confirmed: subgraph returns `user: null` on fresh safes → `process_response` yields `None` → callers receive `{}` → `update_bets` did `{}[bet.id]` unguarded → `KeyError` crash loop.

Fix callsite is correct. `balances.get(bet.id, {})` is the right guard — patching the lookup rather than reverting `fetch_user_positions` back to returning `[]` preserves the sweep-failure detection introduced by #952. Reverting that return value would have silently re-broken the prior fix.

Recovery path checks out: `_replace_with_existing_investments_if_empty` is called after `reset_investments()`, so the prior in-memory investments survive the reset when the new fetch returns empty — no data loss.

Test mock pattern is a genuine regression test. The `process_response` mock returning `None` directly exercises the code path that was crashing; it would fail on the old unguarded `balances[bet.id]`.

Scope is clean: single behaviour file + mechanical hash propagation in `packages.json`. No FSM or config surface touched.

LGTM on the fix.

dagacha:
I used Droid + Claude to have an agent verify:

- The claim above that this was a latent bug  
- That the fix is correct  
- That a CI test will fail if this ever regresses  

Extra info from the agent (feel free to SKIP):

### What the test does *not* guard against (gaps to be aware of)

- **`fetch_user_positions` reverting to `return []` on null/error.** That would re-mask the bug at the source (since the `if user_positions is None` guard becomes dead code again), but it would silently re-break PR #952's withdrawal-sweep partial-pagination detection. There's no test in this file that pins `fetch_user_positions` to return `None` on the null-user response — but the `TestGetActiveBets.test_positions_none_returns_empty` test does pin `get_active_bets` to return `{}` when its dependency returns `None`, which keeps the contract intact at the right layer.
- **A future change that makes `get_active_bets` return a `defaultdict` on the failure path.** This wouldn't be a "regression" per se — it would also fix the symptom — but it would mean the new test's fixture (`{}`) no longer matches reality. The fix at the consumer (`.get(bet.id, {})`) makes the consumer defensive regardless, so the test remains a meaningful negative assertion: the consumer must not depend on defaultdict semantics.

PR #978, opened 2026-05-20
chore(polymarket): disable keyword category filter, log would-be effects

LOCKhart07:
## Summary
- Disables the keyword-based Polymarket category filter at the gate (`is_category_valid = True` override). `_validate_markets_by_category` and `_validate_market_category` still run, so per-category and totals log lines continue to report real validation counts.
- Adds one summary `logger.warning` after deduplication showing total markets and how many would have passed if the filter were on — diagnostic-only, does not change routing.
- New focused test (`test_keyword_filter_disabled_warning_logged`) captures the warning and asserts the counts; exercises both branches of the `would_pass_filter` generator.
- Marks the pre-existing `test_invalid_category_market_blacklisted` as non-falsifiable via a comment (no scope-creep fix to the assertions).

## Test plan
- [x] `uv run pytest packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py` — 166 passed
- [x] 100% line + 100% branch coverage on `polymarket_fetch_market.py`
- [x] `uv run tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint` — all OK (darglint fails with code 129, known Py3.14 tool crash)
- [x] `autonomy packages lock` — hashes refreshed for market_manager_abci and cascaded dependents

## Production evidence (local Polystrat run, 2026-05-20 15:12:14)

New WARNING log fires once per fetch as designed:

```
[WARNING] [agent] Keyword category filter DISABLED — passing all 2556 markets
(if enabled, only 1948/2556 would pass).
```

So **608 markets (~24%) are now flowing that the old filter would have blocked.**

Per-category pre-existing logs continue to report the real verdict the filter would have produced:

| Category   | Would-pass rate (filter on) |
|------------|-----------------------------|
| weather    | 1646/1667 (98.7%)           |
| politics   | 111/233 (47.6%)             |
| finance    | 191/645 (29.6%)             |
| science    | 7/39 (17.9%)                |
| business, technology, health, entertainment, international | 0/0 (no markets returned) |

`is_category_valid=True` confirmed in the per-market debug spot-checks (first 5 bets logged), so the override is in effect. No errors attributable to this change — the two unrelated ERRORs in the log (`py_clob_client_v2` API key, encrypted private key password) are pre-existing startup/auth issues.

## Re-enabling the filter
To restore filtering, remove the override at `polymarket_fetch_market.py:400–401` and delete the diagnostic log block (lines ~371–385). The new test should be deleted along with the log block.

## LOC breakdown — what was actually needed vs. scaffolding

Total diff: **51 lines authored** + 20 lines auto-generated hash replacements across 8 YAML/JSON files.

| Bucket | Lines | Note |
|---|---|---|
| **Core objective** (`is_category_valid = True`) | **1** | The literal disable. Everything else is scaffolding. |
| Override comment | 1 | `# keyword filter disabled — accept all categories` |
| Diagnostic log block (code) | 13 | Two `sum()` generators + the `logger.warning` call |
| Diagnostic log block (comment + blank) | 2 | Block header + trailing blank |
| New test (assertions + setup code) | 21 | `test_keyword_filter_disabled_warning_logged` body |
| New test (def + docstring) | 2 | Method signature + 1-line docstring |
| New test (NOTE comment block) | 5 | Explains why test exists, when to delete |
| New test (blank separators) | 3 | Inside the test body |
| Weak-test marker (pre-existing test) | 3 | All comment, no code change |
| **Hash cascade (auto)** | **20** | `autonomy packages lock` — 0 authored |

Ratio: **1 line of actual change, 50 lines of scaffolding** (~98% scaffolding). Breakdown of the scaffolding:
- ~29% diagnostic log (15 lines) — chosen so the disable is observable in production
- ~61% test (31 lines, of which ~7 lines are explanatory comments)
- ~6% weak-test marker (3 lines) — pure comment, defensive
- ~2% override comment (1 line)

If we re-enable the filter later, the revert is also **1 line** (delete the `is_category_valid = True`); the rest of the scaffolding gets removed in the same commit.


packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
```diff
@@ -1597,6 +1600,37 @@ def test_invalid_category_market_blacklisted(self) -> None:
...
+
+        disabled_calls = [
+            c
+            for c in mock_warning.call_args_list
+            if "Keyword category filter DISABLED" in str(c)
+        ]
+        assert len(disabled_calls) == 1
+        msg = str(disabled_calls[0])
+        assert "passing all 2 markets" in msg
+        assert "1/2 would pass" in msg
```
bennyjo, line 1611: The test asserts the warning fires with correct counts but doesn't verify the actual disable. If `is_category_valid = True` were reverted in `polymarket_fetch_market.py:404`, this test would still pass as long as the diagnostic log block stayed.

Suggest capturing the return from `_exhaust_gen` and adding behavior assertions:

```python
with patch.object(behaviour.context.logger, "warning") as mock_warning:
    gen = behaviour._fetch_markets_from_polymarket()
    result = _exhaust_gen(gen)  # type: ignore[arg-type]
# ... existing log assertions ...
assert len(result) == 2
assert all(r["outcomes"] is not None for r in result)
assert all(r["queue_status"] == QueueStatus.FRESH for r in result)
```

That locks in the disable in addition to the diagnostic log.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
```diff
@@ -1581,6 +1581,9 @@ def test_closed_market_blacklisted(self) -> None:
 
     def test_invalid_category_market_blacklisted(self) -> None:
         """Test market that fails category validation gets blacklisted."""
+        # NOTE: Non-falsifiable — assertions below only check len(result) == 1,
+        # which holds regardless of is_market_valid. Compare with
+        # test_closed_market_blacklisted, which asserts queue_status == EXPIRED.
```
bennyjo: Delete this test — it isn't actually testing anything. `assert len(result) == 1` holds whether the market is blacklisted, passes normally, or the code does nothing, since even a blacklisted market produces one bet_dict (with `outcomes=None`). Coverage of the blacklist path is already in `test_closed_market_blacklisted` (1567-1580).

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -385,6 +400,8 @@ def _fetch_markets_from_polymarket(self) -> Generator[None, None, Optional[List]
             for market in markets:
                 market_id = market.get("id", "unknown")
                 is_category_valid = market.get("category_valid", False)
+                # keyword filter disabled — accept all categories
+                is_category_valid = True
```
bennyjo, line 404: Blocking rollout question: this PR doesn't actually deliver the lift to the users it's aimed at. In `bets.py:372-379`, `update_market_info` short-circuits to `self.blacklist_forever()` whenever the existing or incoming bet has `processed_timestamp == sys.maxsize`. Previously-rejected bets on disk carry `sys.maxsize`, so on the next fetch they're re-blacklisted and stay EXPIRED.

The 608/24% lift in the PR description is from a local run with a fresh-ish store. On a long-running agent — exactly the users with "too few markets to sample from" — most of those 608 are already locked in EXPIRED and won't recover via market churn for weeks. Waiting that long isn't acceptable.

Suggested fix: add a one-shot un-blacklist pass in `setup()` that resets bets matching `processed_timestamp == sys.maxsize` AND `not closed` AND `openingTimestamp > self.synced_time + self.params.opening_margin` back to `FRESH` with `processed_timestamp=0`. That's the population whose only blacklist reason was plausibly category-validity — `_blacklist_expired_bets` and the `closed` check will re-blacklist anything that has another legitimate reason on the next pass.

Also recommend adding a `blacklist_reason` enum to the `Bet` model (`category_invalid`, `closed`, `opening_margin_passed`, `extreme_price`, etc.) and writing it whenever `blacklist_forever()` is called. Without that, every future filter flip lands in the same trap and has to be patched with another heuristic migration. With it, a future PR like this is one line: re-queue all bets where `blacklist_reason == "category_invalid"`.

bennyjo (commented):
Three findings: tighten the new test, drop the non-falsifiable one, and address the previously-blacklisted-bet trap that blocks the lift from reaching long-running agents.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
jmoreira-valory: Independent verification: confirmed. `polymarket_fetch_market.py:404` sets `processed_timestamp = sys.maxsize` on the re-blacklisted bet and writes it via `update_market_info` (`bets.py:372-379`). On the next fetch cycle, `_blacklist_expired_bets` hits the `sys.maxsize` check on the stored bet and calls `blacklist_forever()` again — permanently, regardless of the category fix in this PR.

The PR description's 608-market lift is accurate for a fresh store. For any long-running agent that already accumulated EXPIRED bets before this deploy, the lift is ~0. Those agents are exactly the "too few markets to sample from" cohort the PR is targeting.

Agree with the proposed fix: a one-shot `setup()`-phase migration to reset bets where `processed_timestamp == sys.maxsize AND not closed AND openingTimestamp > synced_time + opening_margin` back to `processed_timestamp=0`. And +1 on the `blacklist_reason` enum — without it this class of breakage will recur.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
LOCKhart07: Deleted in d0f7ac39.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
LOCKhart07: Added the suggested behavior assertions in d0f7ac39 — `result` now captures the generator return, and the test asserts `len(result) == 2`, `outcomes is not None`, and `queue_status == FRESH`. If the `is_category_valid = True` override were reverted, these would fail.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
LOCKhart07: Per discussion with @bennyjo on Slack: measured market churn against the trader's exact fetch filter (9 tags, 4-day endDate window, Yes/No, tradeable, non-extreme-price) shows ~487 new bettable markets/day with pool half-life ~2 days. Stuck agents recover in days, not weeks for typical (cross-category) blacklists.

Decision: this PR doesn't touch existing blacklists. A proper `BlacklistReason` enum + targeted un-blacklist pass (~1-2 days of work) is the right longer-term fix so future filter flips become one-line reverts — tracking as a separate follow-up.

jmoreira-valory (approved):

bennyjo (approved):

PR #311, opened 2024-08-09
[WIP]: Remove hacky dictionary overrides

Adamantios:
This PR demonstrates the changes that https://github.com/valory-xyz/open-aea/pull/750 enables.
This has been tested and works when run as an agent.

Additionally, breaking changes will be necessary in the `open-autonomy` framework to implement the corresponding feature at the service-level overrides.

PR #979, opened 2026-05-20
chore(polystrat): add [POLYSTRAT] funnel logs at fetch/filter/sample

LOCKhart07:
## Summary
- Rewrites existing per-filter logs into a uniform `[POLYSTRAT] filter=X input=N dropped=M kept=K` shape so the Polystrat market-pipeline funnel reads top-to-bottom as a single greppable trace per cycle.
- Adds drop counters at sites that were silent today: `null_or_mismatch_blacklist` / `zero_liquidity_blacklist` (`Bet.__post_init__`), `update_propagated_blacklist` (`update_market_info`), `past_opening_margin_blacklist` / `extreme_outcome_price_blacklist` (`_blacklist_expired_bets`), plus a per-reason breakdown of `processable_bet` rejections and roll-ups of the in-loop sampling drops.
- No drop behavior changed. Counters are local; no signatures, FSM, payloads, or state touched. `bets.py` untouched — null/mismatch and zero-liquidity classification is done from outside `Bet` by inspecting `queue_status` post-construction.

### Stages

**FETCH** — `connections/polymarket_client/connection.py`
- `createdAt_cutoff`, `yes_no_outcomes`, `tradeable` (per category)

**FILTER** — `skills/market_manager_abci/behaviours/polymarket_fetch_market.py`
- `keyword_category` (annotate-only; filter is currently disabled)
- `cross_category_dedup`
- Per-cat parallel reason buckets (marked `stage=reason_bucket`): `closed_flag`, `parse_error`, `invalid_or_missing`, `unexpected_error`
- `per_market_loop` stage total
- `null_or_mismatch_blacklist`, `zero_liquidity_blacklist` (new bets via `_process_chunk`)
- `update_propagated_blacklist` (existing bets re-blacklisted via `update_market_info`)
- `past_opening_margin_blacklist`, `extreme_outcome_price_blacklist` (via `_blacklist_expired_bets`)

**SAMPLE** — `skills/decision_maker_abci/behaviours/sampling.py`
- `processable_bet` rollup (with `fallback_activated={True|False}` field) + `.breakdown.strict` (always) + `.breakdown.fallback` (only when fallback fires, post-filter classification)
- `disabled_tag` (level bumped DEBUG → INFO + uniform format)
- In-loop sequential funnel: `in_loop_zero_liq` → `in_loop_outcome_skew` → `in_loop_neg_risk`

### Alternative considered: simpler logging without helpers

A lighter version of this PR would skip the two new helper methods entirely:
- `_classify_processable_bet` — observability-only mirror of `processable_bet`
- `_is_null_or_mismatch_violation` — post-construction classifier for `Bet.__post_init__` blacklists

…and just emit a single aggregate `dropped=N` at each existing filter point. That would have been roughly half the diff and removed the ~14 tests dedicated to the helpers.

We picked the helper-based approach because the aggregate version drops two pieces of signal we explicitly want:

1. **`processable_bet` per-reason breakdown.** Without `_classify_processable_bet`, a sampling drop is just "this bet was filtered out" — no way to tell if it was `expired` vs `out_of_safe` vs `wrong_queue` etc. With the mirror, you can grep `filter=processable_bet.breakdown.strict` (and `.fallback` when fallback fires) and see exactly which sub-reason dominates a given cycle.
2. **`null_or_mismatch` vs `zero_liquidity` split.** Without `_is_null_or_mismatch_violation`, all `Bet.__post_init__` blacklists collapse into one bucket. With the classifier, "upstream feed sent malformed data" (null/mismatch) and "market has no liquidity yet" (zero-liq) become separately countable — they're different operational signals (one points at a feed problem, the other is normal early-market noise).

Cost: the helpers can drift if `processable_bet` / `Bet._validate` change in the future. Mitigated by (a) both helpers are short and explicitly documented as mirrors of specific upstream code with the file:line of what they mirror, and (b) both are covered to 100% (7 unit tests each), so a drift would either fail those tests or fail in a way visible from the funnel logs themselves.

## Live verification

Captured a real cycle on a local Polystrat agent (4 of 9 categories returned data; the rest were empty from the Gamma API). Capture is from the **final code state after all review fixes** — all renames + new fields verified live.

**Setup-time (existing-store hygiene, before fetch):**
```
past_opening_margin_blacklist    input=675  dropped=3  kept=672
extreme_outcome_price_blacklist  input=672  dropped=0  kept=672
```

**FETCH (per category, only non-zero shown):**
```
politics  createdAt 285→235  yes_no 235→234  tradeable 234→233
science   createdAt  39→39   yes_no  39→39   tradeable  39→39
weather   createdAt 1359→1359 yes_no 1359→1359 tradeable 1359→1205  (154 inactive/unpriced)
finance   createdAt 715→715  yes_no 715→662  tradeable 662→662    (53 non-binary)
```

**FILTER:**
```
keyword_category   stage=total       input=2139 dropped=629 kept=1510 mode=annotate
cross_category_dedup                 input=2139 dropped=28  kept=2111 categories=4
closed_flag        stage=reason_bucket  politics dropped=51
closed_flag        stage=reason_bucket  science  dropped=5
closed_flag        stage=reason_bucket  finance  dropped=70
invalid_or_missing stage=reason_bucket  finance  dropped=5
per_market_loop    stage=total       input=2111 dropped=5 kept=2106 at_construction_blacklisted=126
null_or_mismatch_blacklist           input=20   dropped=0 kept=20
zero_liquidity_blacklist             input=20   dropped=0 kept=20
update_propagated_blacklist          input=2086 dropped=0 kept=2086
past_opening_margin_blacklist        input=692  dropped=0 kept=692
extreme_outcome_price_blacklist      input=692  dropped=23 kept=669
```

**SAMPLE:**
```
processable_bet                  input=2109 dropped=1440 kept=669 fallback_activated=False
processable_bet.breakdown.strict expired=1440 wrong_mode=0 out_of_safe=0 out_of_open=0 wrong_queue=0 processable=669
disabled_tag                     input=669  dropped=602  kept=67  slugs=36
in_loop_zero_liq                 input=1 dropped=0 kept=1
in_loop_outcome_skew             input=1 dropped=0 kept=1
in_loop_neg_risk                 input=1 dropped=0 kept=1
→ sampled
```

**Observations from the captured cycle:**
- All four reviewer-flagged fixes are visible live: `stage=reason_bucket` on the four parallel reason rows (#3), `invalid_or_missing` rename (#2), `fallback_activated=False` field on the top sampling line (#4), `.breakdown.strict` rename (#4). The reworded `_classify_processable_bet` docstring (#1) is source-only.
- Every filter's `input` equals the previous filter's `kept` along its chain. ✓ The four `stage=reason_bucket` rows are explicitly off-chain (parallel, not serial — they all share `input=category_count` because they're alternative reason buckets for the same loop, hence the marker).
- `extreme_outcome_price_blacklist input=692 dropped=23 kept=669` — this cycle, RULE#4 caught 23 markets with prices ≥ 0.99 (vs. dropped=0 in the previous capture). Confirms the counter fires when there's real data to count.
- `processable_bet.breakdown.strict` confirms 1440/1440 sampling drops were `expired=` this cycle — the aggregate-only alternative would have hidden that. (Operationally: nothing was lost to safe-range / opening-range / wrong-queue / wrong-mode here.)
- `disabled_tag` remains the largest single non-expired drop: 602/669 (~90%) rejected by the 36-slug disable list.

**`.breakdown.fallback` not present in this capture** — multi-bet fallback didn't fire (strict mode found 669 candidates immediately, so the fallback branch was never entered). The contract that `.breakdown.fallback` is emitted iff `fallback_activated=True` is locked in by the two new TestSample assertions added in commit 4/4.

**Unrelated infra noise observed (not from this PR):** every `[POLYSTRAT]` info line appears twice in the agent log — once formatted `] [INFO]` (space after `]`) and once `][INFO]` (no space). Two log handlers attached to the same logger in the agent runner config; every other `[INFO]` line doubles too. Doesn't block this PR; can be tracked separately.

## Test plan
- [x] Tail journald on a live Polystrat agent and grep for `\[POLYSTRAT\]` — verified each category produces a contiguous funnel block and the `input` of each filter equals the `kept` of the previous one in its chain. Re-captured post review-fixes; all new labels + fields surface correctly. (See "Live verification" above.)
- [x] Confirmed no existing log strings that monitoring/alerts depend on were removed (only rewritten with new shape; the warnings / per-event lines / startup banners stay as-is). One existing test that asserted on the old `"skipped"` string was updated to assert on the new `filter=invalid_or_missing` shape.
- [x] `uv run tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint` — local pass
- [x] Hashes refreshed via `uv run autonomy packages lock`
- [x] Coverage: `connection.py` 100%, `polymarket_fetch_market.py` 100%, `sampling.py` 100% (418 tests pass across the three modules)
- [ ] CI green on this branch.

Branched off `chore/disable-polymarket-keyword-filter` so it stacks cleanly with the keyword-filter-disable change.

## LOC breakdown — what was actually needed vs. scaffolding

Total diff: **~860 / −130 lines across 17 files** (3 source, 2 tests, 12 auto-generated YAML/JSON hash bumps). Numbers below are approximate (block-level, not per-line accurate) but the proportions hold.

| Bucket | Lines | Note |
|---|---:|---|
| **New drop counters at silent sites** ✱ | ~130 | The actual observability win. `past_opening_margin_blacklist`, `extreme_outcome_price_blacklist` in `_blacklist_expired_bets`; `null_or_mismatch_blacklist` + `zero_liquidity_blacklist` + `update_propagated_blacklist` in `_process_chunk`; per-cat exception split (`closed_flag` / `parse_error` / `invalid_or_missing` / `unexpected_error`) in the fetch loop; in-loop `zero_liq` / `outcome_skew` / `neg_risk` roll-ups in `_sample`. |
| Existing log line rewrites (uniform `[POLYSTRAT]` shape) | ~30 | 9 pre-existing filter logs rewritten in-place across all 3 source files. Pure formatting — could have stayed as-is. |
| `_classify_processable_bet` helper (`sampling.py`) | ~55 | Powers the `processable_bet.breakdown.strict` and `.breakdown.fallback` lines. Cuttable if a single aggregate `dropped=N` is enough. |
| `_is_null_or_mismatch_violation` helper (`polymarket_fetch_market.py`) | 36 | Distinguishes `null_or_mismatch_blacklist` from `zero_liquidity_blacklist`. Cuttable if both buckets can stay merged. |
| Fallback-aware breakdown logic in `_sample` | ~35 | Pre-filter strict-mode snapshot + conditional post-filter fallback-mode snapshot + `fallback_activated` flag + two breakdown log lines. Strictly additive reads — no bet/list mutation. |
| **Tests — coverage for new counter sites** ✱ | ~140 | `was_expired` skip in `_blacklist_expired_bets` (×2), new-bet `null_or_mismatch` vs `zero_liquidity` (×2), `update_propagated_blacklist` count + no-recount-already-expired (×2). Required regardless of helpers. |
| Tests — helper unit tests (`TestClassifyProcessableBet` + `TestIsNullOrMismatchViolation`) | ~250 | 7 + 7 unit tests. Required by 100%-coverage CI gate. Cuttable iff the helpers are cut. |
| Tests — fallback / reason-bucket contract assertions | ~30 | Two new TestSample asserts on the `fallback_activated=True/False` paths + `stage=reason_bucket` assertions on closed_flag and invalid_or_missing rows. Locks in the contract so the markers can't silently regress. |
| Existing test assertion fix (`"skipped"` → `filter=invalid_or_missing`) | 3 | Forced by the log rewrite — a pre-existing test asserted on the old phrasing. |
| Self-doc rename (`rule1_…/rule4_…/update_poison` → self-documenting labels) | ~50 touched, ~0 net | Renames log labels, var names, helper name, and one docstring. |
| Hash cascade (auto-generated, `autonomy packages lock`) | ~50 | Mirrors ~50 deletions across 12 YAML/JSON hash files. Zero authored. |

✱ = the rows you couldn't drop without losing the actual observability win.

**Stripped-down counterfactual:** skip both helpers, skip the breakdown, skip the rename → roughly **~350 lines** vs. ~860 today. The extra ~510 lines bought the `processable_bet.breakdown.{strict,fallback}` per-reason split, the `null_or_mismatch` vs `zero_liquidity` split, self-documenting label names, the fallback-aware breakdown fix from review, and 100% coverage on the helpers.


packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -107,6 +107,54 @@ def processable_bet(
...
+    def _classify_processable_bet(
+        self, bet: Bet, now: int, multi_bets_active: bool
+    ) -> str:
+        """Classify the dominant reason ``processable_bet`` would reject the bet.
+
+        Observability-only mirror of ``processable_bet``: returns one of
+        ``expired``, ``wrong_mode``, ``out_of_safe``, ``out_of_open``,
+        ``wrong_queue``, ``processable``. Does NOT call ``blacklist_forever``
+        — the real filter ``processable_bet`` still runs and is the source of
+        truth. Reason precedence matches the order ``processable_bet``
```
bennyjo: Small ask: the docstring says "Reason precedence matches the order ``processable_bet`` short-circuits in," but `processable_bet` (lines 68–108) doesn't actually short-circuit — it computes `bet_mode_allowable`, `within_opening_range`, `within_safe_range`, and `bet_queue_processable` and returns their AND. The mirror, by contrast, returns the *first* tripping reason in fixed precedence (`expired > wrong_mode > out_of_safe > out_of_open > wrong_queue`).

The observable consequence: a bet that trips both `wrong_mode` AND `out_of_safe` (e.g. `n_bets > 0`, not multi-bets mode, AND opens within `opening_margin + safe_voting_range`) gets classified `wrong_mode` here, but `processable_bet` still calls `bet.blacklist_forever()` via the `not within_safe_range` branch at line 96. So `processable_bet.breakdown` will under-count `out_of_safe` vs. actual blacklists when both conditions overlap.

Two ways forward — either is fine, but worth picking one:
- (a) reword the docstring to say "dominant rejection reason in its own precedence; does not enumerate all tripping conditions" and add a test case for the `wrong_mode`/`out_of_safe` overlap so a future contributor doesn't "fix" the ordering, or
- (b) make the breakdown multi-label (count every condition that fires) so the sum reconciles with the rollup `dropped`.

Happy to leave as-is if (a) is deliberate — flagging because operators reading the breakdown probably expect the sub-counts to reconcile with the rollup.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -519,34 +616,59 @@ def _fetch_markets_from_polymarket(self) -> Generator[None, None, Optional[List]
...
-                except (json.JSONDecodeError, ValueError, TypeError) as e:
+                except json.JSONDecodeError as e:
                     self.context.logger.warning(
-                        f"Skipping market {market_id}: Invalid or missing required fields - {e}"
+                        f"Skipping market {market_id}: parse_error - {e}"
                     )
+                    parse_error_in_category += 1
+                    skipped_in_category += 1
+                    continue
+                except (ValueError, TypeError) as e:
```
bennyjo, line 628: The `missing_required` label is a bit misleading for this clause. Looking at the try-block above, `ValueError` is raised for:
- "Missing required fields (outcomes/prices/token IDs)" — truly missing ✓
- "Mismatched lengths" (line 521) — shape mismatch, not missing
- "Missing endDate" / "Missing conditionId" / "Missing question" — truly missing ✓
- "Negative liquidity" (line 536) — present but invalid
- "Invalid price range" (line 541) — present but invalid
- Plus any `float()`/`int()` cast failures, `date_parser.isoparse` errors, etc.

Plus `TypeError` from cast paths. An operator grepping `filter=missing_required` to diagnose a feed problem will think the upstream omitted fields when the field may be present but malformed (e.g. negative liquidity, NaN prices). Suggest renaming to `invalid_or_missing`, or splitting into a separate `invalid_value` bucket via an inner try around the cast/range checks.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -519,34 +616,59 @@ def _fetch_markets_from_polymarket(self) -> Generator[None, None, Optional[List]
...
                     continue
 
-            # Log summary for this category
-            processed_in_category = category_count - skipped_in_category
-            if skipped_in_category > 0:
+            # Per-reason filter logs for this category (skip zero-drop rows)
+            if closed_in_category:
+                self.context.logger.info(
+                    f"[POLYSTRAT] filter=closed_flag category={category} "
+                    f"input={category_count} dropped={closed_in_category}"
```
bennyjo: These four per-category exception rows (`closed_flag`, `parse_error`, `missing_required`, `unexpected_error`) break the `input/dropped/kept` invariant the rest of the funnel maintains:

1. **No `kept=` field.** Every other `[POLYSTRAT] filter=…` row emits `kept=`. A downstream parser keying on `kept=` will trip on these rows.
2. **`input=category_count` is the same for all four buckets** — these aren't truly chained filters, they're parallel reason buckets inside one loop. The PR description's "Every filter's `input` equals the previous filter's `kept` along its chain" claim doesn't hold here.

Either:
- Add `kept={category_count - dropped}` to each row for shape consistency, or
- Mark these as `stage=reason_bucket` (or similar) so funnel parsers know to treat them as parallel reason rows rather than serial filter rows.

bennyjo (commented):
Three observability-shape notes on the funnel — see inline.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
```diff
@@ -240,15 +301,27 @@ def _sample(self) -> Optional[int]:
...
                 )
 
+        self.context.logger.info(
+            f"[POLYSTRAT] filter=processable_bet "
+            f"input={len(self.bets)} "
+            f"dropped={len(self.bets) - len(available_bets)} "
+            f"kept={len(available_bets)}"
+        )
+        self.context.logger.info(
+            f"[POLYSTRAT] filter=processable_bet.breakdown "
```
jmoreira-valory: **MEDIUM — `processable_bet.breakdown` contradicts `kept` in fallback-activated cycles**

The classification snapshot is taken before the `processable_bet` filter runs (lines 271–300), which is correct for avoiding side-effects. But `kept` in the *preceding* log line (line 305–309) reflects the *post-fallback* count — i.e. bets that were admitted only because multi-bet fallback expanded the eligible set. `breakdown['processable']` was computed in strict mode and will always be ≤ the actual `kept` value whenever fallback fires.

Concrete case: 50 bets total, 30 expire, 10 are `wrong_mode` in strict single-bet mode, 10 are genuinely processable. Fallback fires and admits the 10 `wrong_mode` bets. The logs will show:

```
filter=processable_bet input=50 dropped=30 kept=20
filter=processable_bet.breakdown ... wrong_mode=10 processable=10
```

An operator reading this correctly concludes 10 bets are processable, but `kept=20` — the discrepancy silently hides the fallback expansion and will mislead anyone debugging under-betting or over-betting incidents.

**Suggested fixes:**

Option A — recompute the breakdown *after* the filter, using the actual `available_bets` list:
```python
post_breakdown: Dict[str, int] = defaultdict(int)
for bet in available_bets:
    post_breakdown[self._classify_processable_bet(bet, now=now, multi_bets_active=...)] += 1
```
Then log `post_breakdown['processable']` as `kept` so both lines are consistent. The pre-filter snapshot (for observability of safe-range vs queue rejections) can be kept as a second, separately-named log line.

Option B — add a `fallback_activated=True/False` field to the top-level `filter=processable_bet` line and replace `processable_bet.breakdown` with two distinct log lines: one for strict-mode candidates, one for fallback additions. This makes the expansion explicit rather than burying it in a count mismatch.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -155,6 +161,18 @@ def _blacklist_expired_bets(self) -> None:
...
                     bet.blacklist_forever()
+                    if not was_expired:
+                        extreme_outcome_price_drops += 1
+        self.context.logger.info(
+            f"[POLYSTRAT] filter=past_opening_margin_blacklist "
+            f"input={pre_non_expired} dropped={past_opening_margin_drops} "
+            f"kept={pre_non_expired - past_opening_margin_drops}"
+        )
+        self.context.logger.info(
+            f"[POLYSTRAT] filter=extreme_outcome_price_blacklist "
```
jmoreira-valory, line 172: **LOW — `extreme_outcome_price_blacklist` denominator inflated by pre-expired bets**

`pre_non_expired` is captured once at the top of the method and is used as the `input` denominator for *both* the `past_opening_margin_blacklist` filter and the `extreme_outcome_price_blacklist` filter. But between the two filters, `past_opening_margin_drops` bets have already been blacklisted. The `extreme_outcome_price_blacklist` log line (line 172–175) subtracts those in the `input` field (`pre_non_expired - past_opening_margin_drops`) — that part is correct.

The off-by-one is subtler: any bet that was *already expired before this call* (`was_expired == True`) passes through the `continue` branch without incrementing either drop counter, so it is silently excluded from all reported counts. If 5 bets were pre-expired, the actual pipeline input visible to operators will be `pre_non_expired` (already filtered), but the outer caller will still see those 5 pass through `_blacklist_expired_bets` with no log attribution.

```python
# current: pre_non_expired excludes already-expired bets — but those bets
# still enter the loop and hit  on the
# opening-margin branch without being counted anywhere.
was_expired = bet.queue_status.is_expired()
if self.synced_time >= bet.openingTimestamp - self.params.opening_margin:
    bet.blacklist_forever()
    if not was_expired:             # <-- pre-expired bets silently fall through
        past_opening_margin_drops += 1
    continue
```

Concretely: with 10 pre-expired + 5 past-opening-margin bets, the log reports `input=N dropped=5` but actually 15 bets were blacklisted. A operator debugging a sudden drop in `kept` counts would undercount the blacklist pressure by those 10.

**Suggested fix:** add a `pre_expired_drops` counter alongside the others and log it as a separate upstream stage, or change `pre_non_expired` to `len(self.bets)` and count all three categories independently.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -326,21 +349,91 @@ def get_bet_idx(self, bet_id: str) -> Optional[int]:
         """Get the index of the bet with the given id, if it exists, otherwise `None`."""
         return next((i for i, bet in enumerate(self.bets) if bet.id == bet_id), None)
 
+    @staticmethod
+    def _is_null_or_mismatch_violation(raw_bet: Dict[str, Any]) -> bool:
```
jmoreira-valory, line 353: **NIT — `_is_null_or_mismatch_violation` sync-sentinel comment is fragile**

The docstring says this mirrors `Bet._validate` at `bets.py:248`. That line number will silently drift the next time `bets.py` is edited, and nothing enforces that the two implementations stay in sync.

Two concrete failure modes:
1. A new required field is added to `Bet._validate` (e.g. `resolution`) — `_is_null_or_mismatch_violation` keeps the old `necessary_keys` tuple, drops are undercounted, and no test fails because both paths produce the same overall `blacklist_forever()` result.
2. A field is removed from `Bet._validate` — `necessary_keys` still checks it, inflating `null_or_mismatch_drops` and misleading incident triage.

**Suggested hardening:** add an import-time or test-time assertion that enumerates `Bet._validate`'s required keys and compares them to `necessary_keys`, or move the list into a class-level constant on `Bet` that both `_validate` and `_is_null_or_mismatch_violation` reference.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
LOCKhart07: Addressed in c9cb4d1c. Reworded the docstring to be explicit that `processable_bet` only short-circuits on its first two checks (the last three are computed independently and ANDed), and to call out the `blacklist_forever()` side-effect on `not within_safe_range` regardless of the other conditions — so when a bet trips both `wrong_mode` and `out_of_safe`, the classifier reports `wrong_mode` (its own precedence) while the actual side-effect blacklists the bet and it'll surface as `expired` in the next cycle's breakdown.

Went with your option (a) — kept the precedence-based classification and made the docstring honest about it. The fallback fix in #311 also moves the breakdown to post-filter for `.breakdown.fallback`, which separately addresses the overlap question by reporting the post-side-effect state.

Skipped the overlap test for now — happy to add it in a follow-up if you want it as a regression anchor.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
LOCKhart07: Good call. Addressed in c9cb4d1c: renamed `missing_required` → `invalid_or_missing` everywhere (log label, var name, warning string). Kept it as a single bucket rather than splitting into `missing_field` vs `invalid_value` to avoid restructuring the try block further; the bucket name now accurately covers all of: missing fields, shape mismatches, negative liquidity, invalid price range, cast failures. The existing `test_skipped_markets_logged_in_summary` assertion was updated to match.

Happy to revisit and split into two buckets in a follow-up if `invalid_or_missing` still feels too coarse once we see it in production logs.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
LOCKhart07: Right — these are parallel reason buckets, not chained filters. Addressed in c9cb4d1c by adding `stage=reason_bucket` to all four rows (matches the existing `stage=total` convention on the rollup line). Downstream parsers can branch on `stage=reason_bucket` and skip the chain reconciliation.

Updated the existing `test_closed_market_blacklisted` test to assert on the new marker so it can't silently disappear.

Did NOT add `kept=` to these rows — chose between your two options. The marker felt more honest than synthesizing a `kept=(category_count - dropped)` field that's only meaningful per-row, not as part of the chain.

packages/valory/skills/decision_maker_abci/behaviours/sampling.py
LOCKhart07: Confirmed and addressed in c9cb4d1c. Went with a variant of your Option B:

- Top filter line now carries `fallback_activated=True/False`.
- The previous `.breakdown` row is renamed `.breakdown.strict` (always emitted, captured pre-filter with strict-mode params — preserves the original observability of safe-range / queue rejections).
- A new `.breakdown.fallback` row is emitted **only when fallback fired**, computed post-filter with `multi_bets_active=True`. In a fallback-activated cycle, `breakdown.fallback.processable` reconciles with `kept`.

Important shape detail: `breakdown.fallback` reports post-filter state, so bets that strict's `blacklist_forever()` side-effect just killed appear there as `expired` rather than `out_of_safe`. That's by design — operators reading both rows can see (a) the pre-side-effect picture in `.strict` and (b) the actual state fallback ran against in `.fallback`.

Strictly additive reads: both breakdowns populate fresh local `defaultdict`s and never mutate any bet or list. Two new TestSample assertions lock in the contract on both branches (`fallback_activated=True` emits `.fallback`; `=False` does not).

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
LOCKhart07: Thanks for the careful read. After thinking it through I'd push back gently on this one and keep the current shape — open to reconsidering if you disagree.

The semantic claim of these two rows is "of N non-expired bets entering this function, M were newly blacklisted." That's what the code reports today, and it's accurate: the 10 pre-expired bets in your example were NOT newly blacklisted — `blacklist_forever()` is idempotent and runs as a no-op on them. So `dropped=5` reflects new blacklist pressure correctly. Operators looking at the funnel for "what just got killed" get the right number.

The framing "actually 15 bets were blacklisted" conflates "blacklist_forever() was called 15 times (with 10 of those being no-ops on already-dead bets)" with "15 new dead bets emerged." The latter is 5.

Two real concerns I'd acknowledge, both minor:
1. **Consistency with `processable_bet` shape.** `processable_bet` uses `input=len(self.bets)` and reports the already-expired population in `.breakdown.expired`. `_blacklist_expired_bets` uses `input=pre_non_expired` and silently excludes them. Different conventions in the same funnel.
2. **Operator double-take.** Someone tailing logs might wonder why `input=675` when the store holds e.g. 2089 bets.

But these are setup-only filter rows (called once from `setup()`, not chained after the main fetch funnel), so the inconsistency doesn't actually propagate downstream. And the funnel chains correctly within the function (`extreme_outcome_price_blacklist.input` = `past_opening_margin_blacklist.kept`).

If you'd still like the `pre_expired={N}` info field added for visibility, happy to do that as a small follow-up — but flagging that the suggested change to `len(self.bets)` as the denominator would actually break the chaining invariant within the function. Let me know.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
LOCKhart07: Real concern — agreed it's fragile. Deferring to a follow-up PR rather than bundling it here, for two reasons:

1. **The cleanest fix touches `bets.py`.** Extracting a `Bet.REQUIRED_FIELDS` class constant that both `_validate` and the mirror reference is the most robust option, but this PR's contract is explicitly that `bets.py` is untouched (called out in the summary). Worth doing as a focused change on its own with its own review.
2. **The brittle test-time alternative** (introspecting `Bet._validate`'s source via `inspect.getsource` / `ast` and comparing field tuples) would technically work but adds AST-walk complexity for marginal value — I'd rather not ship that as the "fix" if we're going to do (1) anyway.

Filed mentally as a follow-up. Both failure modes you describe are detectable in the funnel logs (a drift would show up as either an unexplained drop in `null_or_mismatch_blacklist` or an inflation; the helper is short enough that anyone editing `Bet._validate` is reasonably likely to grep for "necessary"), so the immediate operational risk is low. Will keep this on the to-do list.

bennyjo (approved):
Approved. All three of my prior comments addressed cleanly in c9cb4d1c — mirror docstring now honest about precedence + side-effects, `missing_required` renamed to `invalid_or_missing` end-to-end, `stage=reason_bucket` added to the four parallel reason rows. jmoreira's MEDIUM also resolved via the `fallback_activated` flag + split `.breakdown.strict` / `.breakdown.fallback` design, with tests locking both branches. No new substantive changes since c9cb4d1c (062fb312 is a clean main merge). Minor naming/coverage items noted in the re-review are non-blocking follow-ups.

jmoreira-valory (approved):
Three findings from the prior review are resolved at HEAD 062fb31:

- **#1 MEDIUM (processable_bet classification)**: The fix goes beyond the suggested Option B — a `fallback_activated` flag is introduced alongside separate `processable_bet.breakdown.strict` and `processable_bet.breakdown.fallback` branches, both populated via the read-only `_classify_processable_bet` helper. This is cleaner than Option B and correctly resolves the silent-reclassification risk.

- **#2 LOW (extreme_outcome_price_blacklist input)**: `pre_non_expired - past_opening_margin_drops` is now correctly threaded as the input to the blacklist filter, matching the documented intent.

- **#3 NIT (Bet.REQUIRED_FIELDS constant)**: Deferred with a credible follow-up note — acceptable for a chore/logging PR.

LGTM.

PR #981, opened 2026-05-20
fix(ci): stop release matrix from racing on shared artifact uploads

LOCKhart07:
## Summary

The `build-agent-runner` matrix in `release.yaml` has 5 cells (`windows-2025`, `macos-15`, `macos-15-large`, `ubuntu-24.04`, `ubuntu-22.04-arm`). Four of the upload steps (`hash_id`, `agent_id`, `agent.tar.gz`, `agent.zip`) used the **same artifact name across every cell** with `overwrite: true`, causing a race on the artifact finalize call.

The Windows cell finishes last (signing adds ~2 min), so it disproportionately lost the race and 403'd on `FinalizeArtifact` — see the failed run at https://github.com/valory-xyz/trader/actions/runs/26170206622/job/76985590289 where Windows uploaded the bytes, then:

> `Error: Failed to FinalizeArtifact: Received non-retryable error: Failed request: (403) Forbidden`

## Fix

These four artifacts are all OS-independent:

- `hash_id` / `agent_id` — pure `jq` extracts from `packages/packages.json` (Makefile:183-187).
- `agent.zip` / `agent.tar.gz` — archives of `./agent`, which is just `autonomy fetch --remote $(cat hash_id)` (Makefile:189-199) — the package source, not any built binary.

So restrict their uploads to one cell (`ubuntu-24.04`, the cheapest and fastest). Per-OS binary uploads stay per-cell — they're already uniquely named via `${{env.FILENAME}}`.

Other cleanups:

- Drop `overwrite: true` — only there to paper over the race; with one writer it's not just unnecessary, it would mask real collisions.
- Rename all four steps from the identical `"Upload Release Assets"` to distinct labels, so the failed-step in the log is unambiguous.
- Move `prepare archives` from `if: runner.os == 'macOS'` to the same `ubuntu-24.04` gate — fixes a latent bug where Linux/Windows cells were attempting to upload `agent.zip` / `agent.tar.gz` files that don't exist on those runners (saved only by `if-no-files-found: warn`'s default).

## Test plan

Validated end-to-end by the `v0.38.6-test` release (commit `4812d6ef`, which includes this fix): run [26226279440](https://github.com/valory-xyz/trader/actions/runs/26226279440) — all jobs green, Windows finalized with no 403.

- [x] Cut a test pre-release (or rerun this workflow) and confirm all 4 shared artifacts (`hash_id`, `agent_id`, `agent.zip`, `agent.tar.gz`) appear exactly once in the run's Artifacts list
- [x] Confirm all 5 per-OS binaries still upload (`agent_runner_windows_x64.exe`, `agent_runner_macos_x64`, `agent_runner_macos_arm64`, `agent_runner_linux_x64`, `agent_runner_linux_arm64`)
- [x] Confirm the downstream `upload-assets` job's `Download artifacts` steps still find all six artifact names
- [x] Confirm the published GitHub Release includes all expected assets


OjusWiZard (approved):

DhairyaPatel7 (commented):
Clean fix for the matrix upload race. Verified: hash_id/agent_id are OS-independent jq extracts; agent.zip/agent.tar.gz archive an autonomy-fetched ./agent dir (also OS-independent); per-OS binary uploads stay uniquely named via ${{env.FILENAME}}; the latent prepare-archives/upload mismatch (macOS-only prep vs. unconditional upload) is correctly fixed by co-gating both to ubuntu-24.04; downstream upload-assets Download steps still find all six artifact names. No issues.

DhairyaPatel7 (approved):

PR #982, opened 2026-05-21
feat(config): restore named env vars in aea-config.yaml

DhairyaPatel7:
## Summary

Re-applies the named `${VAR:type:default}` syntax to `packages/valory/agents/trader/aea-config.yaml`, **plus** names the new operator-tunable parameters added since the original work. Naming convention aligns with what `service.yaml` already uses, so the same knob has one name end-to-end.

The original named-env-var work (#892-era, commit `0ddd6581`) was reverted in commit `5ae2eb5f` because open-autonomy and the middleware lacked support at the time; both have since shipped that support, and `service.yaml` overrides have been carrying named vars in the meantime — only the agent-level defaults were still anonymous.

**Restored (originally named, then reverted):** 54 distinct vars across 67 occurrences — chain RPCs, subgraph URLs, Polymarket addresses, staking config, operator flags. Three names that diverged from the deployed services have been aligned to match `service.yaml` (`LOG_DIR`, `RESET_TM_AFTER`, `GNOSIS_STAKING_SUBGRAPH_URL`).

**Newly named (added to the agent after the original PR):**
- `DISABLED_POLYMARKET_TAGS`
- `WITHDRAWAL_SLIPPAGE`, `WITHDRAWAL_RETURN_BUFFER`
- `DUST_EPSILON_WXDAI`
- `VALID_MECHS`
- `MECH_MARKETPLACE_V1_SUITABLE_TOOLS`

**Intentionally left anonymous:** framework/protocol plumbing — Tendermint timeouts and retries (`finalize_timeout`, `history_check_timeout`, `keeper_*`, `max_attempts`), consensus settings, fixed protocol addresses, content-type headers, HTTP method strings, fixed subgraph URLs that aren't operator-tunable. The reset cadence (`reset_tendermint_after` → `RESET_TM_AFTER`) is named because it's a per-deploy operator knob, not a framework constant.

Hashes re-locked via `autonomy packages lock`.

## Test plan

- [x] `tomte tox -e check-hash`, `check-packages`, `check-doc-hashes`
- [x] `tomte tox -e check-dependencies`, `check-third-party-hashes`
- [x] `tomte tox -e check-abciapp-specs`, `check-abci-docstrings`, `check-handlers`
- [x] `tomte tox -e gitleaks`, `liccheck`, `safety`, `bandit`
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e darglint -e mypy`, `pylint`
- [x] `tomte tox -e py3.11-darwin` — 3898 passed
- [x] `uv lock --check` and `uv sync --all-groups`
- [ ] Smoke-test Pearl deploy (Omenstrat + Polystrat) to confirm middleware resolves the named vars correctly — this is the exact failure mode that caused the original revert, so it's a hard gate before merge


packages/valory/agents/trader/aea-config.yaml
```diff
@@ -451,7 +451,7 @@ models:
       error_index: ${int:0}
       error_type: ${str:dict}
       retries: ${int:5}
-      url: ${str:https://staking-gnosis.subgraph.autonolas.tech}
+      url: ${GNOSIS_STAKING_SUBGRAPH:str:https://staking-gnosis.subgraph.autonolas.tech}
```
LOCKhart07: `gnosis_staking_subgraph.url` is named `${GNOSIS_STAKING_SUBGRAPH:...}` — missing the `_URL` suffix that every sibling subgraph url in this file uses (`NETWORK_SUBGRAPH_URL`, `OMEN_SUBGRAPH_URL`, `TRADES_SUBGRAPH_URL`, …). Both `service.yaml` files use `GNOSIS_STAKING_SUBGRAPH_URL`, part of a consistent `GNOSIS_STAKING_SUBGRAPH_{METHOD,RETRIES,URL}` family. As-is, this name lives only in the agent config and diverges from the deployed services. Suggest `${GNOSIS_STAKING_SUBGRAPH_URL:str:...}`.

Not a runtime break (the service override wins in `autonomy deploy`), but it gives the same knob two names — exactly what this PR aims to eliminate.

packages/valory/agents/trader/aea-config.yaml
```diff
@@ -165,7 +165,7 @@ models:
       keeper_allowed_retries: ${int:3}
       keeper_timeout: ${float:30.0}
       max_attempts: ${int:10}
-      reset_tendermint_after: ${int:2}
+      reset_tendermint_after: ${RESET_TENDERMINT_AFTER:int:2}
```
LOCKhart07: `reset_tendermint_after` is named `${RESET_TENDERMINT_AFTER:...}`, but both `service.yaml` files use `${RESET_TM_AFTER:...}`. Two things:

1. The names diverge — the agent-level name exists nowhere else in the repo.
2. The PR description says Tendermint internals were *intentionally left anonymous* ("operators shouldn't touch"), yet this one was named. The description and the diff disagree.

Either align it to `RESET_TM_AFTER` (matches the services) or leave it anonymous per the stated policy.

packages/valory/agents/trader/aea-config.yaml
```diff
@@ -135,13 +135,13 @@ type: skill
 models:
   benchmark_tool:
     args:
-      log_dir: ${str:/benchmarks}
+      log_dir: ${BENCHMARKS_DIR:str:/benchmarks}
```
LOCKhart07: `log_dir` is named `${BENCHMARKS_DIR:...}`, but both `service.yaml` files override it with `${LOG_DIR:...}`. Same knob, two names. Suggest `${LOG_DIR:str:/benchmarks}` to match the deployed services.

LOCKhart07 (dismissed):

DIvyaNautiyal07 (dismissed):

packages/valory/agents/trader/aea-config.yaml
DhairyaPatel7: Done in 5d1e109f — renamed to `LOG_DIR` to match `service.yaml`.

packages/valory/agents/trader/aea-config.yaml
DhairyaPatel7: Good catch on the description/diff mismatch. Went with align-not-strip — the original PR (`0ddd6581`) treated `reset_tendermint_after` as an operator knob (how often to wipe Tendermint state, which is genuinely a per-deploy concern, not a framework constant), and `service.yaml` agrees with that read by exposing it as `RESET_TM_AFTER`. Renamed to `RESET_TM_AFTER` in 5d1e109f to match the services. Updated the PR body to be explicit that "Tendermint internals" means timeouts/retries, not the reset cadence.

packages/valory/agents/trader/aea-config.yaml
DhairyaPatel7: Done in 5d1e109f — renamed to `GNOSIS_STAKING_SUBGRAPH_URL` to match the `GNOSIS_STAKING_SUBGRAPH_{METHOD,RETRIES,RESPONSE_KEY,RESPONSE_TYPE,URL}` family in both services.

DIvyaNautiyal07 (dismissed):

LOCKhart07:
https://github.com/valory-xyz/optimus/pull/330

LOCKhart07 (approved):

OjusWiZard (approved):

PR #988, opened 2026-06-03
feat(polymarket): CLOB v2 DepositWallet migration (Path A)

OjusWiZard:
## Summary

Migrates the **Polystrat** trader (W2) from the legacy Safe-funder model to Polymarket's **CLOB v2 DepositWallet (DW)** flow.

CLOB v2 retired the Safe-based trading account — new Polystrat services get a Safe that Polymarket rejects with `400 "maker address not allowed, please use the deposit wallet flow"`. This implements **Path A**: a transient, agent-EOA-owned DepositWallet used as the CLOB funder, while the Olas service **Safe stays canonical** and holds all persistent assets (pUSD float + open CTF positions). The DW is empty at rest; funds sweep in just before a match and back to the Safe right after settlement.

Omenstrat (Gnosis/Omen) is unaffected — every new flow is gated on `IS_RUNNING_ON_POLYMARKET`.

## What changed

**Connection (`polymarket_client`)**
- Order signing switched to `signature_type=3` (POLY_1271) / `funder=DW`.
- New `relayer_proxy.py` client — `deploy_dw`, `exec_wallet_batch`, `sweep_dw`, `transaction`, `nonce` — signing the `wildcard-relayer:v2:…` challenge with the agent EOA. Replaces the Safe-only `RelayClient`.
- New request types + routes: `DEPLOY_DW`, `EXEC_WALLET_BATCH`, `SWEEP_DW`, `RELAYER_TX`, `FETCH_ALL_POSITIONS`, `SELL_POSITION`, `GET_ORDER`.

**Contract**
- New `valory/deposit_wallet` — `execute(...)` (sweep/approval batches) + `owner` read used for rotation / self-heal.

**Skill (`decision_maker_abci`)**
- DW provisioning folded into `PolymarketSetApproval` (deploy + DW trading approvals via `exec_wallet_batch`; Safe redemption approvals stay on the Safe). The setup gate is extended to "allowances recorded **AND** a DW recorded whose owner is the current agent EOA".
- New rounds/behaviours: `PolymarketTopUp` (Safe→DW), `PolymarketSweep` (DW→Safe, post-buy), `PolymarketWithdrawTopUp` (batched Safe→DW of all sellable CTF). The withdrawal sell-loop is now DW-funded, with a batched top-up before and a sweep after.
- Persistence: `deposit_wallet.json` `{dw_address, dw_owner, approvals_done}` drives the setup gate and self-heal / mnemonic-recovery EOA-rotation re-provisioning.

**Asset flow**
```
Setup (once / on EOA rotation): deploy_dw(owner=agentEOA) + DW trading approvals [proxy]
                                + Safe redemption approvals                       [Safe tx]
Buy:    top-up Safe→DW  →  FOK order (funder=DW, sigType=3)  →  poll→settle  →  sweep DW→Safe
Withdraw: batch Safe→DW all sellable CTF  →  per-position FAK sells (DW-funded)  →  sweep DW→Safe
Redeem: Safe.redeem (unchanged; DW untouched)
```

**Config default changes**
- `polymarket_builder_program_enabled` default flipped **`true → false`** (`connection.yaml`, `aea-config.yaml`). With the builder program off by default, DW trading approvals are set via the connection-side `exec_wallet_batch` relayer path as described above; order attribution (`builder_code`) is opt-in and can be re-enabled per-deploy via `POLYMARKET_BUILDER_PROGRAM_ENABLED` + `POLYMARKET_BUILDER_CODE`. Builder-budget capacity is the out-of-scope W5 workstream.

## Verification
- New contract/connection/skill unit tests; **100% coverage on every changed module**.
- `autonomy packages lock --check`, FSM specs, black/isort/flake8/mypy all green.
- Validated end-to-end on live Polygon: fresh-Safe setup → deploy DW + approvals → top-up → FOK buy funded by DW → sweep → withdrawal sell-loop → redeem; restart leaves setup a no-op.
- The diff went through several rounds of adversarial code review (security + behaviour), driven to zero materially-reachable findings.

## Known follow-up
- `polymarket_relayer_proxy_url` currently defaults to the **test proxy** (`mpp.valory.xyz`) rather than `predict-api.olas.network`. Left as-is per discussion; flip before production cutover.

## Out of scope (other workstreams)
W1 predict-api proxy generalization · W3 middleware cleanup (`POLY_SAFE_SERVICE_NAMES`) · W4 frontend "temporarily unavailable" gate · W5 Builder-budget capacity · W6 `predict-polymarket` subgraph fix (must land before cutover).


packages/valory/connections/polymarket_client/connection.py
```diff
@@ -55,7 +62,16 @@
 DATA_API_BASE_URL = "https://data-api.polymarket.com"
 GAMMA_API_BASE_URL = "https://gamma-api.polymarket.com"
 RELAYER_URL = "https://relayer-v2.polymarket.com/"
+# Default wildcard predict-api proxy that fronts the CLOB v2 DepositWallet
+# relayer (the deploy, exec-batch, deployed and transaction calls). Empirically
+# verified endpoint; overridable via ``polymarket_relayer_proxy_url``.
+DEFAULT_RELAYER_PROXY_URL = "https://mpp.valory.xyz"
```
jmoreira-valory, line 68: **MEDIUM — test proxy baked in as default with no service.yaml escape hatch**

`DEFAULT_RELAYER_PROXY_URL = "https://mpp.valory.xyz"` is correctly marked as a pre-production flip in the PR body, but `service.yaml` does not currently expose a `polymarket_relayer_proxy_url` env override. Any service deploy from this branch silently routes DW provisioning, approval batches, and sweeps through the test proxy — the param is overridable at the connection-config level but requires an operator to know to set it, with no env-var hook in the service manifest.

Either include the service.yaml override in this PR, or gate merge on adding it before cutover. Suggested entry under the connection's `config` block in `service.yaml`:

```yaml
polymarket_relayer_proxy_url: ${POLYMARKET_RELAYER_PROXY_URL:str:https://predict-api.olas.network}
```

A PR-body note alone is not an operational safeguard — if the cutover checklist is missed, production funds are DW-provisioned and swept through a test endpoint with no alert.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_deposit_wallet.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+        path = self.params.store_path / DEPOSIT_WALLET_STORE
+        data = {
+            "dw_address": dw_address,
+            "dw_owner": dw_owner,
+            "approvals_done": approvals_done,
+        }
+        try:
+            with open(path, "w") as f:
+                json.dump(data, f, indent=2)
```
jmoreira-valory, line 155: **LOW — non-atomic write of deposit_wallet.json**

`open(path, "w")` truncates the file before `json.dump` completes. A crash between truncation and the final flush leaves a zero-byte or partially-written file. On next boot, `_read_deposit_wallet_file` catches the resulting `ValueError` and returns `None`, which routes `_resolve_or_deploy_dw` back onto the deploy path.

Safe in practice (the relayer deploy is idempotent on duplicate deploys) but wastes a settlement cycle and a relayer round-trip on every crash-during-write. Standard atomic fix:

```python
import os
tmp = path.with_suffix(".tmp")
try:
    with open(tmp, "w") as f:
        json.dump(data, f, indent=2)
    os.replace(tmp, path)
except OSError as e:
    self.context.logger.error(f"Failed to write {DEPOSIT_WALLET_STORE}: {e}")
```

`os.replace` is atomic on POSIX (rename(2)) and atomic on Windows when `tmp` and `path` are on the same volume.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -1272,6 +1396,327 @@ def _encode_set_approval_for_all(self, operator: str, approved: bool) -> str:
...
+                                dw, self.safe_address, token_id, ctf_balance
+                            ),
+                        }
+                    )
+            if not calls:
+                return {"swept": False, "amount": 0, "transaction_id": None}, None
+            nonce = self._dw_nonce(dw)
+            tx_id = self.relayer_proxy.exec_wallet_batch(dw, nonce, calls)
+            return {
+                "swept": True,
```
jmoreira-valory, line 1649: **LOW — sweep submitted but not awaited; "swept": True is misleading on relay failure**

`_sweep_dw` returns `{"swept": True, "transaction_id": tx_id}` as soon as the relayer accepts the `exec_wallet_batch` POST. If the relayed transaction subsequently fails on-chain (out-of-gas, nonce race, network drop), the round exits `DONE` with the response already claiming `swept=True` while CTF may still be in the DW.

This is recoverable — the withdrawal path queries the DW's CTF balance and sells from there — but `"swept": True` paired with an unpolled `transaction_id` is misleading in logs and in any dashboard that reads this field as confirmation of settlement.

Two options:
1. Rename the field: `"submitted": True` + `"transaction_id": tx_id` (caller understands it is submission, not settlement).
2. Add a brief inline comment: `# swept=True means submission accepted; settlement is polled by the behaviour via RELAYER_TX`.

No correctness blocker — the withdrawal path handles unsettled CTF. Documentation/naming clarity only.

LOCKhart07:
### Review — CLOB v2 DepositWallet migration

Thorough pass over the connection, relayer-proxy client, the `deposit_wallet` contract, the new/changed `decision_maker_abci` behaviours + rounds, and the FSM wiring across `tx_settlement_multiplexer_abci` / `trader_abci`. The FSM transitions are internally consistent, the EIP-712 `Batch` signing is unit-tested with real signer recovery, and the three earlier review comments (test-proxy default, atomic write, misleading `swept=True`) are all addressed in `ec5589f`. The Snyk "hardcoded secret" hit is a false positive — a clearly-labelled deterministic test key.

**No blocking issues.** A few non-blocking items:

**1. Dead config: `polymarket_signature_type: 3` is never read** — `connections/polymarket_client/connection.yaml`
The new key is added to the connection config, but `connection.py` signs with the hardcoded constant `SIGNATURE_TYPE_POLY_1271 = 3` and never reads `self.configuration.config.get("polymarket_signature_type")`. It reads as operator-tunable but isn't — either wire it through `_build_clob_client`/`_ensure_dw_funder` or drop the key.

**2. `builder_code` lost its env-var override** — `connections/polymarket_client/connection.yaml`
Changed from `builder_code: ${POLYMARKET_BUILDER_CODE:str:}` to `builder_code: ''`. `_validate_builder_code` still references `POLYMARKET_BUILDER_CODE` in its log messages, but the value can no longer be supplied via env. If the builder program is re-enabled via `polymarket_builder_program_enabled` (now `false` by default), there's no way to provide the code without editing the manifest. Restore the env hook or confirm this is deliberate W5-scope removal.

**3. Opportunistic top-up sweep only reclaims the *current* bet's CTF** — `decision_maker_abci/behaviours/polymarket_top_up.py` (~L97-102)
The comment says it reclaims "pUSD AND CTF stranded in the DW by a prior crash," but it passes `self._position_token_ids()` — the *currently sampled* bet's outcome tokens. CTF from a **different** market stranded by a prior crashed cycle is not swept (only its pUSD is), and the canonical Safe — which withdrawal/redeem act on — never sees it. The post-buy `PolymarketSweepRound` loops-until-success, so this is only reachable on a crash/restart between buy-settlement and sweep; when it happens the position lingers in the transient DW. ERC-1155 has no cheap enumeration so a full fix is non-trivial, but at minimum the comment overstates the guarantee.

**4. Deprecated private API in the batch signer** — `connections/polymarket_client/relayer_proxy.py` (~L352)
`Account._sign_hash(...)` is a deprecated private method. It works on the pinned `eth_account 0.13.7`, but the public equivalent is `Account.unsafe_sign_hash(...)`. Swapping avoids a breakage when the dep is next bumped (already gated `<0.14.0`).

**5. Contract module docstring is inaccurate** — `contracts/deposit_wallet/contract.py` (L25-27)
Says the interface exposes "the `owner()` read ... and `execute` encoding (single-call sweep helper)." The class actually exposes `get_owner` **and** `get_nonce`, and there is no `execute`-encoding method (that lives in the connection / `relayer_proxy`). Update the docstring to match.

**6. PR description omits the `polymarket_builder_program_enabled` default flip**
`true → false` in `connection.yaml`, `aea-config.yaml` (connection + skill config). This materially changes the default approval path (Safe-prepared 8-approval tx instead of connection-side `_set_approval`). Plausibly intentional (builder program is out-of-scope W5), but it's an unmentioned behavioural default change worth a line in "What changed."

packages/valory/skills/decision_maker_abci/states/polymarket_withdraw_top_up.py
```diff
@@ -0,0 +1,43 @@
...
+
+"""This module contains the Polymarket (CLOB v2) DepositWallet withdrawal top-up round."""
+
+from packages.valory.skills.decision_maker_abci.states.base import Event
+from packages.valory.skills.decision_maker_abci.states.polymarket_top_up import (
+    PolymarketTopUpRound,
+)
+
+
+class PolymarketWithdrawTopUpRound(PolymarketTopUpRound):
```
LOCKhart07, line 28: Will this allow users who have stuck funds on current Polystrat to sell off their positions? As now their ctf tokens will be moved to a DW that works and can sell. Basically positions that were bought using agent safe that have not yet resolved which have been stuck due to EOA rotation

LOCKhart07:
### Asset-safety edge: DW pUSD orphaned on agent-EOA rotation

Following up on finding #3 with a narrower, concrete fund-loss case for **pUSD** (not just CTF).

Failed FOK buys are killed, not settled, so the topped-up pUSD stays in the DW. That's normally fine — it's reclaimed by the next cycle's opportunistic sweep (`polymarket_top_up.py` ~L97-102), which sweeps the DW's pUSD by **live balance** (`connection._sweep_dw` reads `_erc20_balance_of(collateral, dw)`), independent of token ids, and runs *before* the Safe-balance check. The withdrawal path's `_sweep_dw_to_safe` is a second backstop. So pUSD is conserved across `BET_PLACEMENT_FAILED` / `IMPOSSIBLE` / `INSUFFICIENT_BALANCE`.

The exception is **agent-EOA rotation while pUSD is stranded in the DW**. If a buy leaves residual pUSD in the DW and the agent EOA then rotates (mnemonic recovery) before the next sweep reclaims it, the setup gate (`polymarket_set_approval._provision_deposit_wallet` + `check_benchmarking._deposit_wallet_ready`) detects the owner mismatch, invalidates the persisted state, and provisions a **fresh DW under the new EOA** — explicitly treating the old DW as inaccessible. The new EOA cannot sign `execute` batches for the old DW, so any pUSD left there is **orphaned with no recovery path**.

The window is narrow (requires a stranded balance *and* a rotation before the next opportunistic sweep), but unlike every other failure path it is not self-healing. Worth either:
- sweeping the old DW back to the Safe *before* invalidating it on rotation (the old EOA key is still available at that point during mnemonic recovery, if recoverable), or
- at minimum logging a loud warning with the orphaned balance + old DW address so the residual is visible to the operator rather than silently abandoned.

packages/valory/skills/decision_maker_abci/payloads.py
```diff
@@ -127,6 +127,35 @@ class PolymarketBetPlacementPayload(MultisigTxPayload):
...
+
+    ``event`` carries the short-circuit signal: ``Event.PREPARE_TX.value``
+    when a pUSD transfer to the DW must settle (``tx_hash`` populated), or
+    ``Event.DONE.value`` when the DW is already funded (no tx settles; jump
+    straight to bet placement). ``dw_address`` propagates the DepositWallet
+    so the funder is available to the bet-placement / sweep rounds.
+    """
+
+    event: Optional[str] = None
+    dw_address: Optional[str] = None
```
LOCKhart07: Question, any reason we're preferring to store dw_address in the synchronized data as compared to shared state? We've previously had problems where we had to thread stuff like this through multiple rounds

LOCKhart07 (dismissed):

packages/valory/skills/decision_maker_abci/states/polymarket_withdraw_top_up.py
OjusWiZard: Yes — validated live on a real grandfathered Safe. The withdrawal flow batch-transfers the Safe's legacy positions to a DepositWallet (`safeBatchTransferFrom`) and sells them via the DW-funded `sigType=3` path the CLOB v2 maker accepts. EOA rotation doesn't block it: the Safe stays controllable via its recovery module, and setup re-provisions a fresh DW under the new EOA.

Proof on Safe `0xF97c70Cf2…`: moved a legacy position Safe→DW ([`0x27f9e7e2…`](https://polygonscan.com/tx/0x27f9e7e2bb1669e5901c5ccef641033ad98059a7a86be41f0161c2208fa129c1)) and sold it DW-funded — matched/settled ([`0xb169e3e9…`](https://polygonscan.com/tx/0xb169e3e9d02a378998fd19ae36252471d4784f03e7abc1429bcc77e7f58f985b)), pUSD swept back, then bought back to restore.

_Note: funds already orphaned in an **old** DW from a prior rotation aren't recoverable this way — separate from Safe-held positions, and covered by the new operator warning._

packages/valory/skills/decision_maker_abci/payloads.py
OjusWiZard: Good point. The durable source of truth is actually the persisted `deposit_wallet.json` (the setup gate and `_resolve_deposit_wallet()` read it) — synced-data was just a fast-path. But a few behaviours (sweep, bet-placement, withdraw-sell) were reading `synchronized_data.deposit_wallet_address` **directly with no fallback**, which is exactly the fragile cross-round threading you're flagging. I've routed those through `_resolve_deposit_wallet()` (synced-data fast-path, else the owner-checked store file), so the DW no longer depends on being threaded into each round's payload.

OjusWiZard: Followed through on this — rather than just adding the fallback, I removed the synced-data threading entirely (`ce4b6b0e`): dropped `SynchronizedData.deposit_wallet_address`, the `dw_address` field from the top-up/sweep payloads, and the round-decode that wrote it. The DW is now resolved solely from the persisted, owner-checked `deposit_wallet.json` via `_resolve_deposit_wallet()` — single source of truth, no cross-round threading. (Polystrat is single-agent, so the local store is authoritative.) 100% coverage on the changed modules.

LOCKhart07 (dismissed):

jmoreira-valory (approved):
Recheck at HEAD fb57798c3f6c6aa3ec6fd10bf612d2a45d993b1e. CI 28/28 green, MERGEABLE.

**Changes since last review (ec5589fb)**

Four commits landed after my prior sign-off:

- `ce4b6b0` — removes synced-data threading for `dw_address` entirely; all behaviours now resolve the DW exclusively from the file-backed `_resolve_deposit_wallet()`. Eliminates the fragile cross-round state pattern flagged in the inline thread. Confirmed: `SynchronizedData.deposit_wallet_address` property gone, `dw_address` removed from `PolymarketTopUpPayload`/`PolymarketSweepPayload`, tests updated at 100% coverage.
- `8d370bb` — agent/skill reference bump (hash lock follow-through).
- `3d769c5` — doc comment noting the benign CLOB api-key 400 in `_build_clob_client`; no logic change.
- `fb57798` — Polystrat UI bundle bump to v0.1.17 (metrics-unavailable card); no service logic.

No new CRITICAL or HIGH findings introduced by any of the four commits.

**Prior findings (jmoreira-valory, addressed in ec5589fb — still in effect)**

- MEDIUM `connection.py:68`: `polymarket_relayer_proxy_url` env override exposed in `service.yaml`. Confirmed present at HEAD (`service.yaml` +1/-1 diff in `ce4b6b0` relock, override intact).
- LOW `polymarket_deposit_wallet.py:155`: atomic `deposit_wallet.json` write (temp + `os.replace`). Confirmed at HEAD.
- LOW `connection.py:1649`: sweep-doc comment clarified (`swept=True` = relayer accepted, not on-chain settled). Confirmed at HEAD.

**LOCKhart07's open items — status at HEAD**

- LOW `builder_code` lost env-var form: **ADDRESSED** — `connection.yaml:L29` restored to `${POLYMARKET_BUILDER_CODE:str:}` at HEAD.
- MEDIUM pUSD orphaned on EOA rotation (`polymarket_set_approval.py:177-195`): **Documented as accepted-risk with loud operator warning.** `polymarket_set_approval.py:187-194` emits a `WARNING`-level log that names the orphaned balance, old DW address, and old owner, directing the operator to manually sweep with the previous key. Auto-sweep is correctly noted as impossible here (the relayer `execute` must be owner-signed by the old EOA, which the rotating process no longer holds). The window is narrow (stranded pUSD AND rotation before the next opportunistic sweep) and the old EOA key remains available to the operator at rotation time. Accepted-risk with visibility — matches LOCKhart07's framing.
- LOW CTF sweep comment overstates scope (`polymarket_top_up.py`): not re-verified individually; low severity, non-blocking.
- NIT `Account._sign_hash` deprecated: not re-verified; NIT, non-blocking.
- NIT `contract.py` docstring: not re-verified; NIT, non-blocking.

**Follow-up items (non-blocking, tracked for a future PR)**

- Wire `polymarket_signature_type` config key through `_build_clob_client`/`_ensure_dw_funder` or drop the dead key from `connection.yaml`.
- Flip `polymarket_relayer_proxy_url` default from `mpp.valory.xyz` to `predict-api.olas.network` before production cutover (already noted in PR body).
- `Account.unsafe_sign_hash` migration when `eth_account` is next bumped.

LGTM.

LOCKhart07 (approved):

PR #989, opened 2026-06-08
feat(polystrat): send Polymarket market description in mech request_context

LOCKhart07:
## What

Captures the Polymarket market **description** (the long-form resolution rules from the Gamma API) and forwards it to the Mech inside the request's `request_context`, for Polystrat.

## Why

The mech request already carries structured `request_context` (market_id, prob, liquidity, spread, …). The market description contains the full resolution criteria, which is useful context for prediction tools. This wires the data through end-to-end on the trader side. (Tool-side consumption of the field is a separate change.)

## Changes

- **`bets.py`** — add optional `description: Optional[str]` field to `Bet` (Omen bets leave it `None`).
- **`bets.py` `to_request_context()`** — emit `description`; the existing `None`-filter drops it for Omen, so only Polymarket requests carry it.
- **`bets.py` `update_market_info()`** — copy `description` on each refresh, so already-queued bets pick up edited resolution rules (Polymarket can change rules mid-life).
- **`polymarket_fetch_market.py`** — read `market["description"]` from the Gamma response into the bet dict.
- Hash cascade (`packages.json`, affected `skill.yaml`/`aea-config.yaml`/`service.yaml`) from `autonomy packages lock`.

## Backward compatibility

Persisted bets without the field decode to `None` via the lenient `BetsDecoder` branch — no crash, and they upgrade to `description: null` on next store. Already-queued Polymarket markets backfill the description on their next refresh.

## Testing

- New unit tests (TDD, red→green): `to_request_context` inclusion/omission, Polymarket fetch mapping (present + missing), `update_market_info` overwrite, and decoder backward-compat for legacy JSON.
- Full `market_manager_abci` suite + `decision_request` + `mech_interact` base tests pass.
- **100% coverage** on both modified source modules.
- Lint clean (black, isort, flake8, mypy, pylint, darglint).
- Verified live on a local Polystrat run: a real mech request went out with the full resolution-rules text populated in `request_context.description`.

---

## LOC breakdown

The actual feature is **8 lines**; everything else is tests and auto-generated hash pointers.

| Category | +Lines | Share | Files |
|---|---:|---:|---|
| **Core logic (required)** | **8** | **7%** | `bets.py` (+7), `polymarket_fetch_market.py` (+1) |
| Tests | 88 | 74% | `test_bets.py` (+53), `test_behaviours_polymarket.py` (+34), `test_utils.py` (±1 fixture) |
| Generated hashes / config cascade | 23 | 19% | `packages.json`, `aea-config.yaml`, 2× `service.yaml`, 4× `skill.yaml` |
| **Total** | **119** | 100% | 13 files |


packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -474,6 +480,7 @@ def to_request_context(self) -> Optional[Dict[str, Any]]:
             "market_close_at": market_close_at,
             "amm_fee": amm_fee,
             "market_spread": spread,
+            "description": self.description,
```
jmoreira-valory: **MEDIUM — Uncapped `description` field flows through consensus payload**

`description` is sourced directly from the Gamma API (`market.get("description")`) with no length bound — Gamma descriptions commonly run 500–2000+ chars. `to_request_context` serializes it into the mech `request_context` dict that gets packed into `SynchronizedData` and travels through Tendermint consensus.

The PR body notes tool-side consumption is a separate change, meaning the full (and unbounded) description starts flowing through consensus on this PR alone. Tendermint's default `max_tx_bytes` is 1 MB, but the concern is a single unguarded field with no ceiling: an unusually large (or adversarially large) Gamma description causes unbounded payload growth with no visible error until a consensus timeout.

Suggested defensive cap before insert:

```python
MAX_DESCRIPTION_LEN = 2000  # chars; above all observed Gamma markets
desc = self.description
if desc is not None and len(desc) > MAX_DESCRIPTION_LEN:
    desc = desc[:MAX_DESCRIPTION_LEN]
```

Alternatively, document the max observed Gamma description length and confirm it's safely below the payload budget — but a defensive cap is cheaper than an incident.

LOCKhart07: Thanks — verified both mechanics, you're right on the structure and I added a cap. Detail for the record:

**Flows through consensus, not just IPFS — confirmed.** `request_context` is serialized into the `MechRequestPayload` consensus payload, not only the IPFS metadata: `mech_interact_abci/behaviours/request.py:1029` (`serialized_requests = json.dumps(self._mech_requests, …)`) → `request.py:1046` packs it into `MechRequestPayload.mech_requests`. So the full text does ride the Tendermint payload (subject to `max_tx_bytes`).

**Length is unbounded at the source — confirmed.** I sampled 2,100 live Gamma markets:

| stat | chars |
|---|---|
| median | 784 |
| p95 | 1,603 |
| p99 | 2,005 |
| max | **2,519** |
| >2000 | 29 (1.4%) |
| >3000 | 0 |

There's no hard cap — descriptions exceed 2000 (~1.4%), so the field is genuinely unbounded in principle (practical ceiling ~2.5 KB; max ~0.25% of the 1 MB `max_tx_bytes`, single-agent deployment, content is author/Gamma-controlled not attacker-injectable — hence low-severity rather than a timeout risk, but a backstop is cheap and correct).

One caveat on the suggested `2000`: that would truncate ~1.4% of real markets mid-rule — and those long ones are the detailed multi-condition resolution texts (e.g. the "Will Russia capture X by date" set) that are exactly the signal we want to send. So I capped at **5000** instead — ~2× the observed max, so it never fires on real data and only bounds a pathological/future outlier:

```python
MAX_REQUEST_CONTEXT_DESCRIPTION_LEN = 5000
...
description = self.description
if description is not None:
    description = description[:MAX_REQUEST_CONTEXT_DESCRIPTION_LEN]
```

Applied in `to_request_context()` (the consensus-bound copy); the full description is retained on the `Bet` on disk. Fix in 1eb5609b, with tests for over-limit truncation and at-limit passthrough.

bennyjo (dismissed):
Approve — tightly-scoped end-to-end plumbing of the Polymarket description into request_context; the 5000-char cap correctly bounds the consensus payload, backward-compat is handled via the lenient decoder, and tests are discriminating. Non-blocking follow-ups noted in review chat (empty-string guard symmetry; treat description as untrusted in the tool-side consumer).

jmoreira-valory (dismissed):
Re-reviewed at \`1eb5609b\`. The MEDIUM finding on the unbounded \`description\` flowing through the consensus payload is addressed:

- \`MAX_REQUEST_CONTEXT_DESCRIPTION_LEN = 5000\` declared at module level with a comment explaining why (consensus-bound, caps an unbounded external string).
- Truncation applied in \`to_request_context()\` only — the consensus-bound copy. Full description retained on the \`Bet\` (disk only).
- Empirical justification on the cap value: 5000 ≈ 2× observed Gamma max (~2.5k chars across a 2.1k-market sample), so real markets are never truncated mid-rule and only pathological/future outliers are bounded.
- Two tests pin the contract: \`test_description_truncated_when_over_limit\` and \`test_description_at_limit_not_truncated\`.

LGTM.

bennyjo (approved):

DhairyaPatel7 (approved):

DIvyaNautiyal07 (approved):

jmoreira-valory (approved):

PR #977, opened 2026-05-19
feat(decision_maker_abci): replace operator allowlist with tool-suitability classifier

jmoreira-valory:
> Depends on https://github.com/valory-xyz/mech-interact/pull/89 -- merge that first; the trader CID bump in this PR points at the new mech-interact CID.

## Summary

- Adds `utils/tool_suitability.py` with a deterministic 5-check `is_prediction_tool` classifier that reads a tool's IPFS manifest blob and decides whether the tool is suitable for binary prediction markets. The classifier hits **100% accuracy** against a 32-tool labelled corpus.
- `ToolSelectionBehaviour` fetches the full IPFS manifest per unique mech CID once per cycle and applies the classifier as the primary tool filter, before the existing ChatUI `selected_mechs` and `allowed_tools` narrowing layers.
- Per-CID manifest fetch uses the canonical `wait_for_condition_with_sleep` retry pattern (mirrors `mech_interact_abci/behaviours/mech_info.py::populate_tools`) so the `mech_tools_api` retry budget is actually engaged on transient IPFS gateway failures.
- The dropped-tool log line is partitioned by reason: classifier rejections are grouped by firing rule via `explain_prediction_tool`; missing-manifest drops emit a separate WARNING cross-referenceable with the per-CID extraction WARNINGs.
- **V1 marketplace path**: the operator allowlist that previously lived on `mech_interact_abci.params.valid_tools` (removed upstream in mech-interact v0.31.0) is restored on the trader side as `decision_maker_abci.params.mech_marketplace_v1_suitable_tools` (env override `MECH_MARKETPLACE_V1_SUITABLE_TOOLS`). It applies only to the V1 marketplace branch in `storage_manager._get_mech_tools`; V2 short-circuits this method upstream and runs the suitability classifier instead. A redundant `not is_marketplace_v2` guard inside `_get_mech_tools` exists belt-and-suspenders so a future refactor can't accidentally route V2 through the V1 allowlist.

## Why

The mech-interact upstream change makes per-mech `relevant_tools` carry the full IPFS manifest tool set rather than the operator-pre-filtered subset. This PR moves tool-suitability decisions to the consumer where they belong: the classifier reads the actual manifest `toolMetadata` and applies content-based vetoes (description-negation, resolver-shape, input-domain mismatch, schema-prose corroboration, passthrough-shape).

[...]

PR #990, opened 2026-06-09
fix(chatui): show and validate tool pins against the suitable set

LOCKhart07:
## Summary

The ChatUI's "Available tools" list and `allowed_tools` validation read the raw `available_mech_tools` synced-data set, which advertises every tool the mechs expose — **including non-prediction tools the suitability classifier rejects** (e.g. `resolve-market-reasoning-gpt-4.1`). This lets a user pin a tool the policy can never select.

Observed live (single agent run):
- tool-selection round: `Tool-suitability classifier narrowed 2 -> 1 tools; rejected by classifier: {'schema_resolver_shape': ['resolve-market-reasoning-gpt-4.1']}`
- chatui `configure_strategies` prompt, same run: `Available tools: {'resolve-market-reasoning-gpt-4.1', 'superforcaster-polymarket-v1'}`

Pinning the resolver passed validation (it's in the raw set), then intersected to an **empty** candidate set at selection → `Event.NONE` → `(ToolSelectionRound, NONE): ToolSelectionRound` self-loop, silently halting trading while the UI reported the pin as successful.

## Root cause

PR #977 moved tool-suitability filtering from the feed point to the selection stage (`_candidate_tools`), leaving the published `available_mech_tools` raw. The ChatUI consumer was not updated to match — the same class of gap #977 fixed on the V1 policy side (the `mech_marketplace_v1_suitable_tools` restore), one layer over on the UI side. The raw set **must** stay raw: the policy learns across all tools and `redeem`/`polymarket_reedem` restore it into the policy, so a misclassification or transient manifest failure must not prune learned history.

## Fix

Two complementary layers, because write-side validation alone can't catch a pin that was valid when made but drifts later (raised by @jmoreira-valory):

**1. Write side (good UX).** `ToolSelectionBehaviour._candidate_tools` publishes the **post-suitability, pre-pin** set on shared state (`available_prediction_tools`). The ChatUI displays and validates `allowed_tools` against it, falling back to `available_mech_tools` when unpublished **or empty** (cold start before the first `ToolSelectionRound`, a round where every manifest fetch failed, or the V1 path that has no classifier). A resolver pin now returns *"Unknown tool(s) ignored"* at pin time instead of silently stalling later.

**2. Read side (robustness).** `_candidate_tools` re-validates the pin every round: it intersects only with the pinned tools that are **actually selectable this round**, and if every pinned tool has dropped out — drifted out of suitability, pinned through a cold-start/IPFS-outage fallback window, or no longer served — it **ignores the pin for the round** (proceeds on the full selectable set) instead of emptying `candidate` and self-looping on `Event.NONE`. This is **non-destructive**: the stored pin is left intact (it re-applies automatically once a tool is selectable again) and the **policy/`accuracy_store` is never touched**. This closes the three persistence/drift paths that write-side validation alone left open (stale pin after restart, resolver pinned in the cold-start window, resolver pinned during a total IPFS outage).

- `available_mech_tools`, the policy, and the redeem paths are **unchanged** — pins stay non-destructive at the selection layer.

Verified live with the patched build: the `configure_strategies` prompt now shows `Available tools: {'superforcaster-polymarket-v1'}`, matching the classifier's selectable set.

## Changes

- `chatui_abci/models.py` — `ChatUISharedState.available_prediction_tools` (shared-instance channel, `None` until published).
- `decision_maker_abci/behaviours/tool_selection.py` — publish the suitable, pre-pin set; **non-destructive read-side revalidation** of `allowed_tools`; document the no-publish-on-setup-failure path (leaving the last-known-good set is safer than clearing to raw).
- `chatui_abci/handlers.py` — `_available_tools()` resolves filtered-or-raw (empty published set falls back to raw, so it can't brick `configure_strategies`); used by display + `allowed_tools` validation.
- Tests + package hash lock cascade.

## Test plan

- [x] `chatui_abci` + `decision_maker_abci` suites pass (new cases: read-side relax + warn, partial-stale keeps valid subset, pin & `accuracy_store` non-mutation, `selected_mechs`-collapse keeps its cause, empty-published-set fallback)
- [x] Coverage-neutral on modified modules (all added statements covered; remaining gaps pre-existing)
- [x] `black`, `isort`, `flake8`, `mypy`, `pylint`, `darglint` green
- [x] End-to-end: live agent prompt shows the filtered set

---

## Scope / what's deferred

This PR closes the `allowed_tools`-vs-**suitability** stalls completely (write-side + read-side). What remains, tracked in **#991**:

- The `allowed_tools`-vs-**`selected_mechs`** (mech×tool) conflict — a tool that is suitable but served by no pinned mech empties `candidate` at the mech-pin step, *after* the read-side revalidation, and keeps its `selected_mechs` cause. Pruning there would silently pick a winner between two valid pins, which deserves UI surfacing rather than a quiet relax.
- Surfacing the stuck state (and the deferred halt-vs-relax decision for the mech-pin case) to the ChatUI.

## Alternatives considered

- **Filter `available_mech_tools` in place at the feed point (restore the pre-#977 invariant for V2).** Tempting because V1 already filters there (`mech_marketplace_v1_suitable_tools`) and it needs no ChatUI change. Rejected: that set seeds the e-greedy policy and is restored into the policy by the redeem paths, so filtering it **prunes the policy's tool universe**. The suitability classifier is a heuristic over per-round IPFS manifests — a false negative or a transient manifest-fetch failure would then permanently destroy a tool's learned accuracy history. The raw set must stay raw; that's why #977 filters non-destructively at selection.

- **Publish via a new `synchronized_data` key (a `ToolSelectionPayload` field).** Works and survives restarts, but needs payload-shape/consensus plumbing, a `selection_key` change, a new `SynchronizedData` property, and hash churn — and it's inconsistent with how the ChatUI reads its *other* config (all via `shared_state`). Since the decision-maker and ChatUI share one `SharedState` instance, a plain shared-state attribute is lighter and self-consistent. Kept the synced `available_mech_tools` as the fallback, so the restart-ephemerality of shared state is covered.

- **Have the ChatUI run the classifier itself.** Rejected: the ChatUI skill has no access to the IPFS tool-manifest metadata (`_tool_metadata`), which is fetched per-round only in `ToolSelectionBehaviour`. It would also duplicate classifier work.

- **Publish the post-pin *candidate* set instead of the pre-pin *suitable* set.** Rejected: the ChatUI must validate pins against the full selectable universe. The candidate set is already narrowed by `selected_mechs`/`allowed_tools`, so reusing it would make otherwise-valid tools appear "unknown" the moment a pin is active.

- **Destructive read-side prune** (rewrite the stored `allowed_tools`, dropping stale entries). Considered for the read-side layer; chose the **non-destructive** variant instead — it self-heals when a tool becomes selectable again, doesn't have the decision-maker behaviour mutate ChatUI's stored config, and never risks the policy store. Cost: the UI keeps showing a temporarily-unavailable pinned tool until #991 surfaces "temporarily unavailable", which is arguably honest (the user *did* pin it).

---

## LOC breakdown (needed vs fluff)

| Category | Lines | Notes |
|---|--:|---|
| **Core logic** | **~16** | the actual functional change |
| Docstrings / comments (source) | ~61 | the why-this-set, fallback, and non-destructive rationale |
| Tests | 229 | read-side relax, fallback, pin/policy non-mutation, etc. |
| Hash-lock cascade | 28 | `packages.json` + `aea-config`/`service`/`skill.yaml` CID bumps |
| Blank (source) | 4 | |
| **Total** | **~338** | |

The functional change is **~16 lines** across 3 files:
- `decision_maker_abci/behaviours/tool_selection.py` (~8) — the publish line + the non-destructive read-side revalidation block
- `chatui_abci/handlers.py` (~6) — `_available_tools()` resolver, the two read-site swaps, and the truthy/empty-set fallback
- `chatui_abci/models.py` (2) — the shared-state field + its `FrozenSet` import

The rest is tests (229), regenerated package hashes (28), and explanatory comments (~61) — i.e. ~5% logic, ~95% supporting.


bennyjo (dismissed):
Tightly-scoped, correct fix: ChatUI now validates/displays `allowed_tools` against the post-suitability set, closing the silent ToolSelectionRound stall. Verified the cross-skill shared-state channel resolves to one instance; both FSM and resilience framework audits ran clean. LGTM.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
```diff
@@ -204,6 +204,16 @@ def _candidate_tools(self) -> Tuple[set, Optional[str]]:
...
 
+        # Publish the post-suitability, pre-pin set for the ChatUI to validate
+        # and display tool pins against (see chatui handler ``_available_tools``).
+        # ``candidate`` here is the suitability-filtered universe with the
+        # raw-set fallback already applied above, so a manifest outage degrades
+        # to the full set rather than blocking the user from pinning anything.
+        # The user pins (selected_mechs / allowed_tools) applied below stay
+        # non-destructive at the selection layer, so the policy keeps learning
+        # across the full ``mech_tools`` set regardless of this published view.
+        self.shared_state.available_prediction_tools = frozenset(candidate)
```
OjusWiZard, line 219: **[Important] The same "valid-in-UI but stalls the round" bug class survives via the `selected_mechs` pin.**

This publishes the *pre-mech-pin* set (deliberate, so the policy validates against the full universe). But the `selected_mechs` intersection at line ~226 runs *after* this publish. So a user can pin an `allowed_tools` tool that passes UI validation (it's in `available_prediction_tools`) yet is served by no pinned mech → `candidate` empties → `cause="selected_mechs"` → `_select_tool` returns `None` → `Event.NONE` self-loop. It's logged per-round (line ~255, "skipping this round so the user can adjust the pin") so not *fully* silent like the original — but the UI still reports both pins as accepted with no conflict surfaced.

This is **pre-existing and arguably out of this PR's stated scope** (the suitability/`allowed_tools` mismatch). Flagging it for a conscious call: either document it as a known limitation, or follow up by surfacing the mech-vs-tool-pin conflict back to the ChatUI. _(Converged across 3 reviewers.)_

packages/valory/skills/chatui_abci/handlers.py
```diff
@@ -329,12 +329,32 @@ def _get_available_valid_mechs(self) -> Set[str]:
...
+        when the filtered set has not been published yet: cold start before the
+        first ``ToolSelectionRound``, a round where every manifest fetch failed,
+        or the V1 marketplace path (which has no classifier and already filters
+        at the feed point).
+
+        :return: the set of tool names the user may pin / that are advertised.
+        """
+        prediction_tools = self.shared_state.available_prediction_tools
+        if prediction_tools is not None:
+            return set(prediction_tools)
```
OjusWiZard, line 355: **[Advisory] Unnecessary copy.** Both return branches are read-only at all call sites (`in`, `join`, list-comp), and `FrozenSet[str]` already satisfies `Set[str]` for reads. Widen the annotation on line 332 to `FrozenSet[str]` (or `AbstractSet[str]`) and `return prediction_tools` directly — drop the `set(...)` materialization. _(type-design + simplifier converged.)_

packages/valory/skills/chatui_abci/handlers.py
```diff
@@ -329,12 +329,32 @@ def _get_available_valid_mechs(self) -> Set[str]:
...
+        return self.synchronized_data.available_mech_tools
+
     def _get_available_tools(
         self, http_msg: HttpMessage, http_dialogue: HttpDialogue
     ) -> Optional[Set[str]]:
         """Get available mech tools, handle errors if not available."""
         try:
-            return self.synchronized_data.available_mech_tools
+            return self._available_tools()
```
OjusWiZard, line 363: **[Advisory] `except TypeError` only.** No regression here — the new `available_prediction_tools` access can't raise (always initialized to `None`), and the `TypeError` risk lives on the `available_mech_tools` fallback. Since you're already touching this site, consider broadening to `(TypeError, AttributeError)` to defend against a future SharedState/synced-data shape mismatch returning a 500 instead of the intended 425.

packages/valory/skills/chatui_abci/models.py
```diff
@@ -91,6 +91,14 @@ def __init__(self, *args: Any, skill_context: SkillContext, **kwargs: Any) -> No
 
         self._chatui_config: Optional[ChatuiConfig] = None
 
+        # Suitability-filtered tool set published by the decision-maker's
+        # ToolSelectionBehaviour each round (the post-classifier, pre-pin
+        # universe). The ChatUI validates and displays tool pins against this
+        # so a user cannot pin a tool the policy would never select. It stays
+        # None until the first ToolSelectionRound runs; consumers fall back to
+        # the raw `available_mech_tools` synced-data set in that window.
+        self.available_prediction_tools: Optional[FrozenSet[str]] = None
```
OjusWiZard, line 100: **[Advisory] `None` vs empty-`frozenset` is an overloaded encoding.** `None`="not published, fall back"; empty `frozenset`="zero selectable tools". The suitability path is guarded against publishing empty (tool_selection.py ~200-205 keeps the raw set), so the invariant holds *today* — but `frozenset(candidate)` would still publish empty if `mech_tools` itself is empty, which the UI reads as "nothing pinnable." Not worth a wrapper type here; a one-line note at the write site ("candidate is never empty unless mech_tools is genuinely empty") would lock in the reasoning.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
```diff
@@ -204,6 +204,16 @@ def _candidate_tools(self) -> Tuple[set, Optional[str]]:
                     "can proceed."
                 )
 
+        # Publish the post-suitability, pre-pin set for the ChatUI to validate
+        # and display tool pins against (see chatui handler ``_available_tools``).
+        # ``candidate`` here is the suitability-filtered universe with the
+        # raw-set fallback already applied above, so a manifest outage degrades
+        # to the full set rather than blocking the user from pinning anything.
+        # The user pins (selected_mechs / allowed_tools) applied below stay
+        # non-destructive at the selection layer, so the policy keeps learning
```
OjusWiZard: **[Advisory] Comment slightly overstates the guarantee.** Accurate for `self.policy` (pins only restrict an *ephemeral deepcopy* in `_select_tool`), but the phrasing reads as a stronger claim under an active pin. Suggest tightening to: "the main policy object is not mutated by pinning — only an ephemeral deepcopy is restricted — so accuracy accumulates across all tools between rounds."

packages/valory/skills/chatui_abci/tests/test_handlers.py
```diff
@@ -106,6 +107,12 @@ def _make_handler(
 
     shared_state = MagicMock()
     shared_state.chatui_config = current_config or ChatuiConfig()
+    # None (default) makes _available_tools fall back to the raw
```
OjusWiZard: **[Advisory] Redundant comment.** The `prediction_tools` param name and the two named test methods below (`test_falls_back_to_mech_tools_when_unpublished`, `test_prefers_filtered_set_when_published`) already say this. Deletable — it only adds rot surface if the fallback logic changes.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_tool_selection.py
```diff
@@ -1139,6 +1139,54 @@ def test_suitability_skipped_when_no_metadata(self) -> None:
...
+        )
+        behaviour._tool_metadata = {
+            "good": _PredictionMetadata.predictor(),
+            "also-good": _PredictionMetadata.predictor(),
+        }
+
+        candidate, _ = behaviour._candidate_tools()
+
+        assert candidate == {"good"}  # pin applied to the returned set
+        # published set is pre-pin (full suitable universe, not the pinned subset)
```
OjusWiZard: **[Advisory] Redundant comment** — restates the test's own docstring + the assertion on the next line; deletable.

**[Advisory] Worthwhile test to add** (documents the Important finding on the publish line): no test currently shows that a UI-valid tool can still stall via the mech-pin interaction. Adding one pins the publish point's intent:
```python
# allowed_tools=["good"], selected_mechs=["0xa"], mech 0xa serves only {"also-good"}
candidate, cause = behaviour._candidate_tools()
assert candidate == set() and cause == "selected_mechs"
assert behaviour.shared_state.available_prediction_tools == frozenset({"good", "also-good"})  # UI still shows "good" pinnable
```

OjusWiZard (commented):
Multi-perspective review (6 specialized agents, findings deduped & ranked). Net: the core fix is correct — publish point is post-suitability/pre-pin, the `None`→fallback semantics are sound, and the single-threaded behaviour/handler model rules out any race on the shared field. One scope-adjacent gap (Important) + advisories inline below; none block merge.

jmoreira-valory (changes requested):
Second-pass review with adversarial verification. The fix is in the right direction — correctly identifies the wrong reader of `available_mech_tools` and publishes the suitable set for the handler. However the deep dive surfaced one P1 that I think blocks merge plus a few smaller items worth landing in the same PR.

## 1. Persistence revalidation gap reproduces the NONE-loop (P1, request-changes)

The fix closes the write-side validation gap (handler accepts only pins in the suitable set) but does NOT re-validate persisted pins at the read side. Three concrete paths still reproduce the original silent-halt bug:

**Path A — Restart with a stale persisted pin**
1. User pins `allowed_tools=[X]` while X is in the suitable set. Stored to `chatui_param_store.json`.
2. Agent restarts → `ChatUISharedState.__init__` sets `available_prediction_tools = None` at `chatui_abci/models.py:100`. Pin survives on disk.
3. Mech republishes metadata OR classifier output drifts → X no longer in `suitable`.
4. First `ToolSelectionRound`: `_candidate_tools` reads the persisted pin directly via `self.shared_state.chatui_config.allowed_tools` at `tool_selection.py:230` — no revalidation against the new `candidate`.
5. `candidate &= {X}` = ∅ → `cause = "allowed_tools"` → `Event.NONE` → self-loop. Original bug, reproduced.

**Path B — Cold-start race**
Fresh boot, `available_prediction_tools is None`, handler falls back to raw `available_mech_tools` (`handlers.py:347-350`). User pins a resolver via `configure_strategies` in the pre-first-round window → validation accepts → persisted. First `ToolSelectionRound` → same NONE-loop.

**Path C — Total IPFS outage**
`_fetch_mech_manifests` fails every CID, `_tool_metadata` stays `{}`, safety-net keeps `candidate = mech_tools` (raw, includes resolvers). Publish at `tool_selection.py:215` writes the RAW set as `available_prediction_tools`. User pins a resolver, validation accepts. Next round (after IPFS recovers) → NONE-loop.

**Suggested fix** — single ~5-line addition at the top of the pin block in `_candidate_tools` (after the suitability filter, before the intersection at line 230):

```python
allowed_tools = self.shared_state.chatui_config.allowed_tools
if allowed_tools:
    stale = [t for t in allowed_tools if t not in candidate]
    if stale:
        self.context.logger.warning(
            f"Pruning stale allowed_tools entries {stale} no longer in the "
            f"suitable set; user must re-pin if needed."
        )
        self.shared_state.chatui_config.allowed_tools = [
            t for t in allowed_tools if t in candidate
        ] or None
```

Closes all three paths in one site. The PR description's framing ("non-destructive at the selection layer") describes the policy/learning side correctly, but the SELECTION intersection at line 232 IS destructive — that's how the bug manifests.

## 2. `available_prediction_tools = frozenset()` semantics (P1)

`_available_tools()` in `chatui_abci/handlers.py:347-350` triggers the filtered branch on `prediction_tools is not None`. An empty `frozenset()` is therefore "no tools selectable" — display shows empty, validation rejects every pin. No new test covers the empty case in either reader. If the publish in `_candidate_tools` ever writes an empty frozenset (corner cases in the safety-net branch or where `mech_tools` happens to be empty), the handler silently rejects every future `configure_strategies` POST.

Either:
- Switch the fallback trigger to `if prediction_tools:` (truthy) instead of `is not None`, treating empty as "fall back to raw", or
- Add explicit writer-side and reader-side tests pinning the current semantic.

## 3. Publish is not "every round" as the PR description implies (P2)

`_select_tool` returns early when `_setup_policy_and_tools` returns False (`tool_selection.py:240-242`). That happens when `_get_tools()` leaves `self._mech_tools` empty (transient mech-info outage, V2 cold start before MechInformationRound, `selected_mechs` pinning a mech with zero `relevant_tools`). In that path `_candidate_tools` never runs → `available_prediction_tools` stays at its prior value (None on cold start, stale frozenset later) → handler falls back to raw `available_mech_tools`, reopening the validation gap.

Either move the publish earlier (defensive publish in `_select_tool` immediately after a successful `_setup_policy_and_tools`) or document the no-publish path explicitly.

## 4. "Non-destructive at the selection layer" comment is misleading (P2 / docstring)

The comment at `tool_selection.py:212-214` reads:

> The user pins (selected_mechs / allowed_tools) applied below stay non-destructive at the selection layer, so the policy keeps learning across the full mech_tools set regardless of this published view.

That describes the POLICY/accuracy_store side correctly. But the SELECTION result IS destructive — the pin intersection at line 232 can empty `candidate`, which is exactly the mechanism producing the bug this PR is supposed to fix. Suggest tightening the comment to "policy/accuracy state stays non-destructive" so the next reader doesn't conflate the two.

## 5. No test asserts composed SharedState instance identity (nit)

The whole fix rests on the MRO: `decision_maker_abci.SharedState(ChatUISharedState, MechInteractSharedState)` and the assumption that `self.shared_state` resolves to the SAME composed instance from both `ToolSelectionBehaviour` (writer) and `HttpHandler` (reader). The 11 new tests mock these separately. A single test pinning `behaviour.shared_state is handler.shared_state` in the composed context (e.g. via a minimal `_TestableComposedSharedState` setup) would catch a future refactor that accidentally splits them.

---

Posting as request-changes since #1 is blocking. Happy to discuss scope — if you'd prefer to ship the smaller fix now and track persistence revalidation as a follow-up, an issue is fine, but the PR description should disclose the gap explicitly (currently it doesn't note that the persistence path still reproduces the bug under the three scenarios above).

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
LOCKhart07: Good catch, and agreed it's reachable — I confirmed the mechs serve identical tool sets on *today's* Polystrat (all three point to the same service metadata, so the `selected_mechs` intersection is currently a no-op), but that's a data coincidence, not a guarantee: heterogeneous-mech deployments are supported, so this path is real.

Marking it **out of scope for this PR**, which is deliberately scoped to the one *structural, preventable* case (a resolver can never become a useful predictor → never pinnable). This finding is the broader **runtime-drift** class — a pin valid at pin time becomes unsatisfiable later — which no amount of pin-time validation can fully catch (mechs dropping a tool, a tool being reclassified, etc., all hit the same empty-candidate `Event.NONE` self-loop). The durable fix is cause-agnostic: surface/handle the stall at the selection point rather than validating each cause.

Tracked in #991. I've left the publish point pre-pin (deliberate) and skipped the suggested fabricated-heterogeneous-mech test, since it would assert a path that can't occur in the current deployment and reads as if it can — #991 will cover it with the real fix.

packages/valory/skills/chatui_abci/handlers.py
LOCKhart07: Kept the `set()` here intentionally — this return value is str-formatted straight into the LLM prompt (`-- Available tools: {available_tools}`). A `set` renders `{'x'}`; a `frozenset` renders `frozenset({'x'})`, which would both look wrong and diverge from the `available_mech_tools` fallback (already a plain `set`). Returning the `frozenset` directly would regress the prompt. Added a comment at the site explaining why so a future simplifier doesn't strip it. The other advisories (broadened `except`, comment tightening, redundant-comment removal) are applied.

packages/valory/skills/chatui_abci/handlers.py
LOCKhart07: Applied in `edd60c93` — broadened to `(TypeError, AttributeError)` so a future SharedState/synced-data shape mismatch returns the intended 425 rather than a 500.

packages/valory/skills/chatui_abci/models.py
LOCKhart07: Agreed a wrapper type isn't worth it. Added the note at the write site (`tool_selection`) in `edd60c93`: `candidate` is only ever empty if `mech_tools` itself is empty (no tools discovered), whereas the `None` default means "not published yet, fall back to `available_mech_tools`". That locks in the reasoning you flagged.

packages/valory/skills/decision_maker_abci/behaviours/tool_selection.py
LOCKhart07: Applied in `edd60c93`, using your phrasing — the publish-site comment now reads that the main policy object is not mutated by pinning (only an ephemeral deepcopy is restricted in `_select_tool`), so accuracy accumulates across all tools between rounds.

packages/valory/skills/chatui_abci/tests/test_handlers.py
LOCKhart07: Removed in `edd60c93` — the `prediction_tools` param name and the two named tests (`test_prefers_filtered_set_when_published`, `test_falls_back_to_mech_tools_when_unpublished`) already document it.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_tool_selection.py
LOCKhart07: Removed the redundant comment in `edd60c93`. On the suggested test for the mech-pin interaction: skipping it here — it asserts an empty-candidate stall that can't occur in the current homogeneous-mech deployment, so it would read as if the path were reachable. The mech×tool stall is tracked in #991, where the real fix and its test will live.

OjusWiZard (dismissed):
Approving. The core fix is correct — `ToolSelectionBehaviour` publishes the post-suitability, pre-pin set and the ChatUI validates/displays `allowed_tools` against it, closing the resolver-pin stall at the UI layer. The `None`→`available_mech_tools` fallback is sound and the single-threaded behaviour/handler model rules out a race on the shared field. Follow-up `edd60c93` addressed the advisory comments (broadened except, documented the kept copy + None/empty encoding, tightened the policy-learning comment, dropped redundant comments).

Two non-blocking items remain, tracked in open review threads:
- The `selected_mechs` mech-pin can still stall a round for a tool that passes UI validation (pre-existing, out of this PR's stated scope).
- Minor: the published set can lag `available_mech_tools` between tool-selection rounds; cross-skill shared-state coupling is implicit.

None block merge.

LOCKhart07:
Thanks @jmoreira-valory — the adversarial pass was spot-on. Addressed in `128b399d` (description updated to match):

**1. Persistence/read-side gap (P1) — fixed.** Added non-destructive read-side revalidation in `_candidate_tools`: it intersects `allowed_tools` only with the tools actually selectable this round, and if all pinned tools have dropped out it ignores the pin for the round (relaxes to the full selectable set) instead of emptying `candidate` → `Event.NONE`. Closes all three paths (stale-on-restart, cold-start-window resolver, IPFS-outage resolver). Non-destructive on purpose: the stored pin is left intact (re-applies once a tool is selectable again) and the policy/`accuracy_store` is never touched. A genuine `selected_mechs` collapse keeps its own cause (guarded on a non-empty `candidate`).

**2. Empty-`frozenset` semantics (P1) — fixed.** `_available_tools` now uses a truthy check, so an empty published set falls back to raw rather than reading as "reject every pin." Test added.

**3. Publish not every round (P2) — documented.** When `_setup_policy_and_tools` fails, `_candidate_tools` doesn't run and `available_prediction_tools` is intentionally left at its last-known-good value — clearing it would make the ChatUI fall back to the raw set (re-exposing unsuitable tools), which is worse; the read-side revalidation (#1) covers any resulting staleness. Comment added at the early-return.

**4. Misleading "non-destructive at the selection layer" comment (P2) — already fixed** in `edd60c93` (predates this review): it now states the *main policy object* isn't mutated (only an ephemeral deepcopy is restricted).

**5. SharedState instance-identity test (nit) — skipped.** A unit test asserting `behaviour.shared_state is handler.shared_state` would have to mock both contexts to the same object (testing the mock, not the wiring); genuine instance-sharing is a framework property, verified end-to-end by the live run (the prompt shows the set `ToolSelectionBehaviour` published). The `decision_maker.SharedState(ChatUISharedState, …)` MRO is the structural guarantee.

On scope: this PR now fully closes the `allowed_tools`-vs-suitability class. The `selected_mechs` (mech×tool) conflict you'd reach via a different pin combination keeps its `selected_mechs` cause and is tracked in #991 (re-scoped accordingly), since relaxing there silently picks a winner between two valid pins and warrants UI surfacing instead. Re-requesting review.

OjusWiZard (approved):

jmoreira-valory (approved):
Re-reviewed at `128b399d`. The new commit addresses all five findings from my earlier `CHANGES_REQUESTED` review, plus the polish items from the parallel review. Clearing the block.

Mapping:

**#1 — Persistence revalidation gap (P1)** ✅ Addressed via non-destructive read-side revalidation in `_candidate_tools` (`tool_selection.py:233-256`). The chosen invariant is *better* than my suggested 5-line prune:
- Non-destructive: the persisted pin is left intact on disk and re-applies automatically once a pinned tool becomes selectable again. My prune would have silently lost user state.
- Cause-agnostic: catches all the runtime-drift triggers (restart with drifted pin, cold-start race, IPFS outage publishing the raw set), not just the persistence path I anchored on.
- Doesn't touch the policy / accuracy_store.
- Genuine `selected_mechs` collapse keeps its own `cause` for #991 to handle separately. Correct scoping.

**#2 — Empty-frozenset semantics (P1)** ✅ `_available_tools` switched from `is not None` to truthy check, so an empty published set falls back to raw `available_mech_tools` rather than locking out every pin. Docstring updated to spell out the empty-set case. `test_falls_back_to_mech_tools_when_published_set_empty` locks it in.

**#3 — Publish not every round (P2)** ✅ Documented at the `_select_tool` early-return path (`tool_selection.py:268-275`) — leaving the last-known-good `available_prediction_tools` is safer than clearing it (clearing re-exposes the raw set in the UI), and the new read-side revalidation absorbs any staleness.

**#4 — "Non-destructive" docstring overstates (P2)** ✅ Rewritten in `edd60c93` using the suggested phrasing about the ephemeral deepcopy.

**#5 — Composed SharedState instance-identity test (nit)** Not added — acceptable. The instance identity was independently verified by an earlier reviewer, so the current correctness is anchored; adding a test is regression-protection rather than correctness coverage.

Plus three new behaviour-level tests anchor the read-side revalidation contract:
- `test_all_stale_returns_full_set_no_cause`
- `test_partial_stale_keeps_valid_subset_silently`
- `test_selected_mechs_collapse_keeps_its_cause`

And `test_pin_left_intact_and_policy_untouched` locks in the non-destructive invariant.

Issue [#991](https://github.com/valory-xyz/trader/issues/991) tracks the broader cause-agnostic stall handling — the residual `selected_mechs × allowed_tools` interaction is in scope there.

LGTM.

PR #992, opened 2026-06-10
fix(decision_maker): surface suitable tool set to ChatUI under stop-trading

LOCKhart07:
## Problem

`ToolSelectionRound` publishes the suitability-filtered tool set to `shared_state.available_prediction_tools`, which the ChatUI displays and validates tool pins against (PR #990). But `ToolSelectionRound` **only runs on the trading path** (`CheckStopTrading(DONE) → Randomness → Sampling → ToolSelection`).

Agents in **stop-trading mode** (no funds, unstaked, KPI paused → `Event.SKIP_TRADING`) never reach it. `available_prediction_tools` stays `None`, so the ChatUI handler falls back to the raw `available_mech_tools` set and re-exposes non-prediction tools — e.g. `resolve-market-reasoning-gpt-4.1` — which a user can then pin into a silent `ToolSelectionRound` self-loop.

Observed in a real Polystrat log: agent stop-trading from boot, ChatUI prompt showing `Available tools: {'superforcaster-polymarket-v1', 'resolve-market-reasoning-gpt-4.1'}`.

### Why #990 didn't cover this
#990 made the ChatUI *read* the filtered set, but the set is only *produced* on the trading path. This is the structural root: in #977 the suitability filter moved from the every-period feed point onto the trading-only selection path. For the live (marketplace-v2 / Polymarket) path there was never a stop-trading-safe filtered view.

## Fix

Publish the suitable set from the **shared setup path** instead of selection-only:

- Move the manifest fetch/classify helpers (`_fetch_mech_manifests`, `_fetch_one_manifest`, `_extract_tool_metadata`) down from `ToolSelectionBehaviour` into the shared `StorageManagerBehaviour` — pure relocation.
- Add `_maybe_publish_suitable_tools()` and call it from `_setup_policy_and_tools`. The **redeem path** runs `_setup_policy_and_tools` every period (before `CheckStopTrading`), regardless of trading state.

Properties:
- Runs **once per boot** (gated on `available_prediction_tools is None`; re-runs each boot because `is_policy_set` resets — no disk store, shared-state only).
- **v2-only** and skipped in benchmarking mode (the classifier needs the mech manifest).
- On a manifest fetch failure or an all-unsuitable verdict it leaves the field **unpublished**, so the ChatUI degrades to the raw set rather than hiding every tool.
- `available_mech_tools` is **never narrowed** — the e-greedy policy keeps learning across all tools.
- `ToolSelection`'s richer per-round publish (with mech/tool-pin narrowing) is **unchanged** and still takes precedence when trading.

`mech_interact_abci` (which owns `MechInformationRound`) is **not touched** — the fix stays entirely within `decision_maker_abci`.

## Verification (live run)

After deploying, a fresh run shows:
- `polymarket_redeem_behaviour` enters → `_setup_policy_and_tools` logs `Tool-suitability classifier fetched 1 unique mech manifest CID(s)` then `Published 1 suitable prediction tool(s) for the ChatUI from the setup path` — **6 seconds before `tool_selection_round` is even entered**.
- All ChatUI prompts now read `Available tools: {'superforcaster-polymarket-v1'}` — `resolve-market-reasoning-gpt-4.1` gone.

## Tests & checks

- 6 new `TestMaybePublishSuitableTools` cases: publishes-when-unset, skips-when-already-set (no re-fetch), skips-v1, skips-benchmarking, no-publish-on-fetch-failure, no-publish-when-all-unsuitable.
- Full `decision_maker_abci` suite: 1392 passed; redeem/blacklisting/decision_receive: 237 passed.
- New method fully covered; remaining per-module coverage misses are pre-existing (identical in baseline, covered by the full CI suite) — moved code only.
- black / isort / flake8 / mypy / pylint / darglint all green; hashes re-locked.

## Scope of the diff
Hand-written: `storage_manager.py`, `tool_selection.py`, `test_tool_selection.py`. The remaining files (`packages.json`, agent `aea-config`, two `service.yaml`s, three `skill.yaml`s) are auto-generated content-address hash bumps from `autonomy packages lock`.


packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
```diff
@@ -472,13 +590,59 @@ def _try_recover_mech_tools(self) -> Optional[List[str]]:
...
+        """Publish the suitability-filtered tool set for the ChatUI, once per boot.
+
+        Runs from ``_setup_policy_and_tools``, so under stop-trading -- when
+        ``ToolSelectionRound`` never executes -- the redeem path still populates
+        ``shared_state.available_prediction_tools`` and the ChatUI shows the
+        post-classifier universe instead of the raw ``available_mech_tools``.
+
+        Gated on ``is None`` so ``ToolSelectionRound``'s richer per-round publish
+        (with its mech/tool-pin narrowing) wins whenever the agent is actually
+        trading; this only fills the stop-trading gap and re-runs each boot
```
OjusWiZard: **Important — the redeem `is_policy_set` short-circuit can bypass this publish entirely.**

Both `RedeemBehaviour._setup_policy_and_tools` (`reedem.py:881-888`) and `PolymarketRedeemBehaviour._setup_policy_and_tools` (`polymarket_reedem.py:161-168`) `return True` early when `is_policy_set` is true, never reaching `super()` — which is where this method is wired in.

Fine on a fresh boot (TM reset → empty db → `is_policy_set` false → super path runs, matching the live verification). But on any restart where the consensus db is rebuilt with `policy` already present (e.g. crash-recovery via Tendermint block replay instead of a reset), `is_policy_set` is true at boot while `available_prediction_tools` (in-memory) is back to `None` — a stop-trading agent in that state regresses to the exact bug this PR fixes.

Cheap hardening: also `yield from self._maybe_publish_suitable_tools()` inside the short-circuit branch of both overrides — the `is None` gate makes it a no-op everywhere else. If Pearl restarts always TM-reset, waiving this is defensible, but then this docstring line ("re-runs each boot because `is_policy_set` resets") should state that assumption explicitly: under a db-replay restart `is_policy_set` does *not* reset, and the re-run gate is really the `is None` check.

packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
```diff
@@ -472,13 +590,59 @@ def _try_recover_mech_tools(self) -> Optional[List[str]]:
...
+        :yield: None
+        """
+        if self.shared_state.available_prediction_tools is not None:
+            return
+        if not self.synchronized_data.is_marketplace_v2:
+            return
+        if self.benchmarking_mode.enabled:
+            return
+
+        yield from self._fetch_mech_manifests()
```
OjusWiZard, line 634: **Advisory — one redundant manifest fetch per boot on the trading path, and worst-case boot stall.**

On the first `ToolSelectionRound` of a boot, `_select_tool` → `_setup_policy_and_tools` runs this fetch, then `_select_tool` unconditionally calls `_fetch_mech_manifests()` again for `_candidate_tools` (it resets `_tool_metadata` first). One extra IPFS round-trip per CID, once per boot — bounded and correct, just noting it.

Also: under persistent IPFS 5xx each CID costs up to ~6 × `rpc_sleep_time` (~60 s) before retry exhaustion lets the redeem behaviour proceed. Same bounded-retry pattern as the existing `_get_tools`, so not a regression, but it now also sits on the redeem path — maybe worth a sentence in this docstring so operators can tune `retries`.

packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
```diff
@@ -472,13 +590,59 @@ def _try_recover_mech_tools(self) -> Optional[List[str]]:
...
+        if not self._tool_metadata:
+            return
+
+        suitable = {
+            tool
+            for tool in self.mech_tools
+            if is_prediction_tool(self._tool_metadata.get(tool))
+        }
+        if not suitable:
```
OjusWiZard, line 643: **Important — the all-unsuitable branch is the one silent degradation path.**

The fetch-failure path is observable (the no-metadata warning in `_fetch_mech_manifests`), and the trading path logs "classifier marked every candidate as unsuitable" in `_candidate_tools` — but this branch returns with no log. An operator debugging "why does the ChatUI show raw tools under stop-trading" can't distinguish IPFS outage from an all-unsuitable verdict.

```python
if not suitable:
    self.context.logger.warning(
        f"Tool-suitability classifier marked all {len(self.mech_tools)} "
        "tool(s) as unsuitable during setup; the ChatUI will fall back "
        "to the raw mech_tools set."
    )
    return
```

packages/valory/skills/decision_maker_abci/tests/behaviours/test_tool_selection.py
```diff
@@ -1282,3 +1282,124 @@ def test_drop_partition_separates_classifier_from_missing_manifest(
...
+        called[0] = True
+        behaviour._tool_metadata = dict(metadata)  # type: ignore[attr-defined]
+        return
+        yield  # make it a generator function
+
+    behaviour._fetch_mech_manifests = _gen  # type: ignore[assignment,method-assign]
+    return called
+
+
+class TestMaybePublishSuitableTools:
```
OjusWiZard, line 1327: **Important — the integration through `_setup_policy_and_tools` is effectively untested.**

These six cases drive `_maybe_publish_suitable_tools()` directly, but the production entry point — the new `yield from` in `_setup_policy_and_tools` — is never exercised. `test_storage_manager.py:1521` (`TestSetupPolicyAndTools.test_returns_true_when_tools_and_policy_set`) reaches it but only passes by accident: the MagicMock `shared_state.available_prediction_tools` is not `None`, so the method exits at the first guard. Inverting the `is None` gate would not fail that test.

Relatedly, the redeem-path delegation to `super()` is mocked out in `test_reedem.py:3149` / `test_polymarket_reedem.py:394`, so the stop-trading scenario itself has no end-to-end test — which is also how the `is_policy_set` short-circuit gap (see comment on `storage_manager.py`) went unnoticed.

A small test stubbing `_maybe_publish_suitable_tools` and asserting `_setup_policy_and_tools` invokes it would lock in the wiring.

OjusWiZard (commented):
Multi-perspective review pass (correctness, silent failures, tests, comments, simplification). Core change is sound: the moved helpers are verbatim-identical, imports clean, no circular imports, and the six new tests are well-targeted. Four inline notes — one conditional behavioral gap worth an explicit decision, two cheap hardening items, one advisory.

Checked and dismissed: the "(before CheckStopTrading)" docstring claim is correct per the FSM composition; `mech_tools` vs lowercased `_tool_metadata` keys is not a case bug (lowercased at collection on both v1/v2 paths); the double v2/benchmarking guards are load-bearing; no thread-safety concern on the shared-state write (single-threaded FSM).

packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
LOCKhart07: Fixed in c9271d69. Both `RedeemBehaviour` and `PolymarketRedeemBehaviour` `_setup_policy_and_tools` overrides now `yield from self._maybe_publish_suitable_tools()` inside the `is_policy_set` short-circuit branch, so the db-replay restart case is covered; the `is None` gate keeps it a no-op on the fresh-boot/super path. Added an empty-`mech_tools` guard in the method so the short-circuit call is safe even when `available_mech_tools` is empty. Also updated this docstring to state the `is None` vs `is_policy_set` gating explicitly, as you suggested. Wiring is now asserted by new tests in `test_reedem`/`test_polymarket_reedem`.

packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
LOCKhart07: Fixed in c9271d69 — added the warning (`marked all N tool(s) as unsuitable during setup; the ChatUI will fall back to the raw mech_tools set`), so this path is now observable and distinguishable from an IPFS outage. Covered by `test_no_publish_when_all_unsuitable`, which asserts the warning is logged.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_tool_selection.py
LOCKhart07: Fixed in c9271d69. Added wiring tests asserting the production entry point invokes the publish: `test_storage_manager.py::test_invokes_maybe_publish_suitable_tools` (base setup, also pins ordering — publish after `_set_policy`, so inverting/removing the call fails), plus `test_short_circuit_publishes_suitable_tools` in both `test_reedem` and `test_polymarket_reedem` for the redeem short-circuit path you flagged above.

packages/valory/skills/decision_maker_abci/behaviours/storage_manager.py
LOCKhart07: Documented in c9271d69 — the method docstring now notes the one redundant per-boot manifest fetch on the trading path, and that under a persistent IPFS outage each CID costs up to `retries` × `rpc_sleep_time` before the redeem behaviour proceeds, so operators can tune `retries`. Kept as-is otherwise (bounded, same retry pattern as `_get_tools`).

OjusWiZard (approved):
Reviewed across two passes (correctness, silent failures, tests, comments, type design, simplification — details in the inline threads, all resolved).

The follow-up commit c9271d69 addressed everything: the redeem `is_policy_set` short-circuits now publish the suitable set (covering db-replay restarts), the all-unsuitable verdict is logged, the `_setup_policy_and_tools` wiring and both redeem overrides are pinned by spy tests, and the docstring now states the actual gating semantics. Second pass verified the delta clean: `available_prediction_tools` is never reset mid-run so the `is None` gate is a true one-shot-per-boot guard, the empty-tools guard ordering is correct, and the new `yield from` is transparent to all five call sites. CI fully green including the coverage gate. LGTM.

jmoreira-valory (approved):
Re-reviewed at \`c9271d69\` (was reading \`2398e962\` when the follow-up landed).

**The fix is structurally sound.** Verified independently:

- **FSM placement**: \`trader_abci/composition.py\` routes every cycle through \`RedeemRouterRound → RedeemRound\` ("always-redeem-first"), so \`RedeemBehaviour.async_act\` runs unconditionally and reaches \`_setup_policy_and_tools()\` regardless of stop-trading state. Confirms the PR's central claim.
- **Relocation purity**: \`_fetch_mech_manifests\`, \`_fetch_one_manifest\`, \`_extract_tool_metadata\` plus \`_tool_metadata\` / \`_pending_cid\` init move from \`ToolSelectionBehaviour\` to \`StorageManagerBehaviour\` with no logic change. \`ToolSelectionBehaviour._candidate_tools\`'s per-round publish is preserved and still wins during trading.
- **Gating**: \`is None\` (already-published) + \`not is_marketplace_v2\` + \`benchmarking_mode.enabled\` + empty \`_tool_metadata\` (fetch failure) + empty \`suitable\` (all-unsuitable) — all five short-circuits covered by \`TestMaybePublishSuitableTools\` (6 cases).
- **db-replay hardening (\`c9271d69\`)**: nice catch — \`is_policy_set\` is true at boot on a Tendermint db-replay while in-memory \`available_prediction_tools\` is back to \`None\`, so without the direct call from the \`is_policy_set\` short-circuit a stop-trading agent would regress to the raw set. Both redeem overrides now invoke \`_maybe_publish_suitable_tools()\` directly, and the new \`_mech_tools\` empty-guard makes the call safe.
- **Observability (\`c9271d69\`)**: the all-unsuitable path now logs a WARNING; closes the only previously-silent degradation path. The new \`test_no_publish_when_all_unsuitable\` asserts the log fires.
- **Wiring tests (\`c9271d69\`)**: \`test_invokes_maybe_publish_suitable_tools\` (base), \`test_short_circuit_publishes_suitable_tools\` × 2 (both redeem overrides), \`test_skips_when_no_mech_tools\` (empty-guard). Locks the wiring the original 6 direct-call tests left unexercised — exactly the regression-protection gap I was going to flag.

One residual observation, not blocking: in pure stop-trading mode (no Tendermint restart between mech metadata republishes), the published set is the first-boot snapshot until trading resumes. The docstring already implies this via the "in-memory shared state, so it is None again after any restart" framing — bounded impact, since the read-side revalidation in \`_candidate_tools\` (PR #990's \`128b399\`) catches stale pins at selection time and selection only runs once trading resumes.

LGTM.

PR #993, opened 2026-06-17
feat(staking): decouple staking KPI from off-chain activity target

LOCKhart07:
## What & why

Today a single on-chain liveness threshold does two unrelated jobs at once: it unlocks staking rewards **and** signals "epoch work done" (the cue Pearl auto-run uses to rotate agents). On the new staking contracts the on-chain requirement is normalised to ~1 request/epoch, so reusing it as the "work done" cue would let the agent stop after a single request — the farming behaviour this removes.

This PR **splits that one number into two**:

1. **On-chain KPI** (`is_staking_kpi_met`) — unchanged meaning, gates staking, ~1 on the new contracts. Healthcheck-only.
2. **Off-chain activity target** (`activity_target`, default **8**) — governs when the agent considers its epoch work done and drives the stop vote + auto-run rotation signal (`is_activity_target_met`).

New behaviour only engages when staked on a **new (decoupled-activity) contract**, detected on-chain — old contracts behave exactly as before.

## How regime detection works

The agent only receives the staking-contract address from Pearl, so it detects the regime itself: read `activityChecker()` on the staking contract, then `VERSION()` on the checker. `VERSION() == "0.2.0"` ⇒ NEW regime; absent / unexpected value / no checker ⇒ OLD (fail-safe). A *transient* RPC error is retried, never silently classified OLD. The verdict is cached once per process on the composed agent's shared state.

## Changes

- **`staking_abci/behaviours.py`** — `_is_new_staking_regime()` seam + `_read_optional_contract_value` helper (accepts a resolved `None` as a valid answer, distinguishes it from transient errors).
- **`check_stop_trading_abci`** — `_compute_activity_status()` reads liveness/service inputs once and derives both signals coherently; `_compute_stop_trading` returns a `StopTradingResult`; the stop vote keys off `activity_target_met`. Payload, round `end_block`, and synced-data carry `is_staking_kpi_met` / `is_activity_target_met` / `activity_target` / `activity_completed` — `is_staking_kpi_met` now sourced from the payload (decoupled from the vote).
- **`decision_maker_abci`** — `/healthcheck` exposes the three new fields (round-owned, never recomputed); `sampling.kpi_is_met` retargets to `is_activity_target_met`; shared-state cache field added to `models.SharedState`.
- **Contracts** — `mech_activity` gains `get_activity_checker` + `version` (and `activityChecker`/`VERSION` ABI entries), both degrading gracefully on a genuinely missing function.
- **Config** — `activity_target` (`ACTIVITY_TARGET`, default 8) wired into `skill.yaml`, both service YAMLs, and `aea-config.yaml`.

## Notes / deviations from the design spec

1. The spec placed `get_activity_checker` on the `staking_token` contract, but that package is **third-party / gitignored** in this repo (its hash can't be locked here). The read is routed through the tracked **`mech_activity`** package instead — functionally identical; `staking_token` is untouched.
2. `get_activity_checker` degrades gracefully too (not just `version()`), because old `ServiceStakingToken` contracts have no `activityChecker()` and would otherwise make the seam retry forever.
3. Pre-existing latent issue flagged (not fixed, out of scope): `staking_token/contract.py` has a `liveness_ratio()` method calling `livenessRatio()`, which isn't in that contract's ABI — dead code on the v2 path.

## Testing

- 1606 tests pass; **100% coverage** on every touched module.
- `black` / `isort` / `flake8` / `mypy` / `pylint` / `darglint` all green.
- `autonomy packages lock` idempotent; no FSM-spec changes.

## LOC breakdown (added lines, excludes Makefile/spec)

| Category | Lines |
|---|---:|
| **Production Python** | **423** |
| &nbsp;&nbsp;– code | 216 |
| &nbsp;&nbsp;– docstrings | 144 |
| &nbsp;&nbsp;– comments | 36 |
| &nbsp;&nbsp;– blank | 27 |
| **Test Python** | **655** |
| &nbsp;&nbsp;– code | 516 |
| &nbsp;&nbsp;– docstrings | 55 |
| &nbsp;&nbsp;– comments | 13 |
| &nbsp;&nbsp;– blank | 71 |
| ABI JSON (`MechActivity.json`) | 26 |
| Hashes / fingerprints (`autonomy packages lock`) | 46 |
| Config YAML (`activity_target` wiring) | 4 |
| **Total added** | **1154** |

Hand-written code is **216 prod + 516 test = 732 lines**; the remaining ~422 are docstrings/comments (~248), generated hashes/ABI (72), config wiring (4) and blanks (98). Test-to-prod code ratio ≈ 2.4×.


packages/valory/contracts/mech_activity/contract.py
```diff
@@ -23,6 +23,14 @@
...
 from aea.contracts.base import Contract
 from aea.crypto.base import LedgerApi
+from web3.exceptions import BadFunctionCallOutput, ContractLogicError
+
+# "No such function" surfaces as one of these: an absent ``VERSION`` getter
+# either reverts (``ContractLogicError``) or returns empty data that fails ABI
+# decoding (``BadFunctionCallOutput`` / ``ValueError``). A connection / RPC /
+# timeout error is NOT in this set and must propagate so the caller retries
+# instead of mis-classifying a transient blip as "old checker".
+_NO_SUCH_FUNCTION = (BadFunctionCallOutput, ContractLogicError, ValueError)
```
OjusWiZard: **🔴 Important — narrow this catch tuple.** `Web3ValueError` subclasses the built-in `ValueError`, so transient RPC failures that surface as a plain `ValueError` (e.g. `aea_ledger_ethereum.ethereum` raising `ValueError("Skipping the filtering operation as the RPC is misbehaving.")`) get swallowed here → `data=None` → regime classified **OLD** and cached for the whole process lifetime. On a new-regime contract that reverts the agent to the ~1-request on-chain target and stops after a single mech request — the farming behaviour this PR exists to remove.

Suggest:
```python
_NO_SUCH_FUNCTION = (BadFunctionCallOutput, ContractLogicError)
```
(add `ABIFunctionNotFound` if a local-ABI miss is a concern — it's an `AttributeError`, not a `ValueError`, and isn't reachable since the new ABI includes both selectors). Plus a regression test asserting a transient `ValueError` **propagates** rather than degrading to `data=None`.

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -159,8 +174,19 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
...
-        is_staking_kpi_met = self.positive_vote_threshold_reached
-        self.synchronized_data.update(is_staking_kpi_met=is_staking_kpi_met)
+        # The persisted ``is_staking_kpi_met`` must come from the payload, not
+        # the vote: in the new regime the vote tracks the off-chain activity
+        # target while ``is_staking_kpi_met`` stays the on-chain KPI, so the two
+        # can legitimately disagree. Single-agent ⇒ a representative payload is
+        # authoritative.
+        payload = cast(CheckStopTradingPayload, next(iter(self.collection.values())))
+        is_activity_target_met = bool(payload.is_activity_target_met)
+        self.synchronized_data.update(
```
OjusWiZard, line 200: **🟡 Important — move this update below the `NONE`/`NO_MAJORITY` guard (line 196).** These four fields are written to the DB *before* the round confirms convergence, so on a non-converged round (`Event.NONE`/`NO_MAJORITY`) they persist anyway and the healthcheck can surface partial/incomplete data. Pre-existing structural pattern, but now more observable with the new `activity_target`/`activity_completed` fields carrying real semantic weight.

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -98,78 +114,143 @@ def params(self) -> CheckStopTradingParams:
...
-        liveness_period = self.liveness_period
-        self.context.logger.debug(f"{liveness_period=}")
-
         yield from self.wait_for_condition_with_sleep(self._get_liveness_ratio)
-        liveness_ratio = self.liveness_ratio
-        self.context.logger.debug(f"{liveness_ratio=}")
 
-        mech_requests_since_last_cp = (
-            staking_kpi_request_count - mech_request_count_on_last_checkpoint
+        # length-2 ``getMultisigNonces`` on both V1 and V2 ⇒ index 1 is requests
```
OjusWiZard: **🟢 Advisory — inaccurate comment.** The data is sourced from `getServiceInfo()[2]` (the `nonces` field), not from any function named `getMultisigNonces` (which isn't in the ABI). Rename the reference to avoid confusing future readers.

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -98,78 +114,143 @@ def params(self) -> CheckStopTradingParams:
...
+                completed=0,
+            )
 
+        (
+            staking_kpi_met,
+            activity_target_met,
+            target,
+            completed,
+        ) = yield from self._compute_activity_status()
         self.context.logger.debug(f"{self.params.stop_trading_if_staking_kpi_met=}")
```
OjusWiZard, line 223: **🟢 Advisory — name now misleading after the decoupling.** The next line gates `stop` on `activity_target_met`, so `stop_trading_if_staking_kpi_met` no longer means "stop when the staking KPI is met" — it's effectively "stop when the activity target is met". Renaming the param is config-breaking, but a short comment noting the semantic shift would prevent confusion.

OjusWiZard (commented):
Automated multi-perspective review (code / silent-failure / tests / types / comments). Inline comments below cover the actionable, line-anchored findings. Verdict: approve with changes — the regime-split design and the transient-vs-absent error boundary are sound; one inline finding (contract.py `ValueError`) genuinely undermines that fail-safe and should be fixed before merge.

Non-line-anchored items (test-assertion gaps despite 100% line coverage):
- `test_behaviours.py` `async_act`: payload contents never asserted (stubbed `send_a2a_transaction`) — a swapped kwarg would pass.
- `test_rounds.py` `end_block`: no test sets non-default payload fields and asserts the `synchronized_data.update(...)`; the vote/KPI decoupling has no round-layer assertion.
- `staking_abci/test_behaviours.py`: `test_absent_version_is_old` doesn't assert the cache write; `cached=False` shortcut path untested.
- `decision_maker_abci/test_models.py`: `SharedState.__init__` doesn't assert `staking_regime_is_new is None` (the caching contract depends on `None`, not `False`).

packages/valory/skills/check_stop_trading_abci/behaviours.py
```diff
@@ -98,78 +114,143 @@ def params(self) -> CheckStopTradingParams:
...
+        self.context.logger.debug(
+            f"{staking_kpi_met=} {activity_target_met=} {target=}"
         )
-        self.context.logger.debug(f"{required_mech_requests=}")
+        return staking_kpi_met, activity_target_met, target, completed
 
-        if mech_requests_since_last_cp >= required_mech_requests:
-            return True
-        return False
+    def is_staking_kpi_met(self) -> Generator[None, None, bool]:
```
DhairyaPatel7: 🟢 **Minor — no remaining production callers.** Grepping the skills (non-test) shows `is_staking_kpi_met()` is no longer called anywhere: `sampling.kpi_is_met` now reads `synchronized_data.is_activity_target_met`, and the healthcheck reads the round-owned synced-data fields. So the "retained for existing callers" rationale no longer holds — it's exercised only by tests. Consider dropping it (folding its assertion into a `_compute_activity_status` test) or updating the docstring to call it a test-only shim.

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -77,9 +77,24 @@ def last_review_timestamp(self) -> int:
...
 
+    @property
+    def is_activity_target_met(self) -> bool:
+        """Get whether the off-chain activity target is met (the rotation signal)."""
+        return bool(self.db.get("is_activity_target_met", False))
+
+    @property
+    def activity_target(self) -> int:
+        """Get the per-epoch activity target."""
+        return cast(int, self.db.get("activity_target", 0))
```
DhairyaPatel7: 🟢 **Minor — inconsistent with the sibling `SynchronizedData`.** These two properties use `cast(int, self.db.get(..., 0))` (a no-op hint), whereas the same-named properties in `decision_maker_abci/states/base.py` additionally guard `None → 0`. Harmless today because `end_block` always writes `int(payload.activity_target or 0)`, so the DB never holds `None` — but two divergent impls of the same logical field are a latent foot-gun if anything ever writes `None`. Suggest mirroring the None-safe version here for consistency.

packages/valory/contracts/mech_activity/contract.py
LOCKhart07: Fixed in 3e888937. Dropped `ValueError` → `_NO_SUCH_FUNCTION = (BadFunctionCallOutput, ContractLogicError)`, and rewrote the comment to spell out why (web3 raises a `ValueError` subclass for transient RPC blips, which must propagate so the read retries rather than caching an OLD verdict for the process lifetime). The existing parametrized tests were inverted accordingly: a transient `ValueError("Skipping the filtering operation as the RPC is misbehaving.")` is now asserted to **propagate** for both `get_activity_checker` and `version`, and removed from the absent→`None` cases. `ABIFunctionNotFound` is unreachable since the ABI carries both selectors, so it was not added.

packages/valory/skills/check_stop_trading_abci/rounds.py
LOCKhart07: Fixed in 3e888937. The four-field `synchronized_data.update(...)` now runs **after** the `Event.NONE / NO_MAJORITY` guard, so a non-converged round no longer persists partial activity data. Added `TestEndBlockPersistsActivityFields` with two cases: a converged event writes the payload fields verbatim, and a `NO_MAJORITY` event leaves the accessors at their defaults.

packages/valory/skills/check_stop_trading_abci/behaviours.py
LOCKhart07: Fixed in 3e888937. Comment now reads: "``getServiceInfo()[2]`` is the length-2 ``nonces`` field on both V1 and V2 ⇒ index 1 is the mech-requests count." No more reference to a non-existent `getMultisigNonces`.

packages/valory/skills/check_stop_trading_abci/behaviours.py
LOCKhart07: Addressed in 3e888937. Kept the config key for back-compat but added a `NOTE` above the gate making the semantic shift explicit: post-decoupling the flag gates on `activity_target_met` (the regime-aware target), not the on-chain staking KPI; the docstring of `StopTradingResult` carries the KPI-vs-target distinction.

packages/valory/skills/check_stop_trading_abci/behaviours.py
LOCKhart07: Good catch — removed in 3e888937. `is_staking_kpi_met()` had no production callers left (`sampling.kpi_is_met` reads `synchronized_data.is_activity_target_met`, the healthcheck reads the round-owned synced-data fields), so I dropped the method and its wrapper test. The `staking_kpi_met` arithmetic remains fully covered by `TestComputeActivityStatus`, and `async_act` now asserts the emitted payload fields directly.

packages/valory/skills/check_stop_trading_abci/rounds.py
LOCKhart07: Fixed in 3e888937. Both `activity_target` and `activity_completed` now use the None-safe guard mirroring `decision_maker_abci/states/base.py` (`if value is None: return 0`), removing the divergence. Added `test_activity_target_none_coerced_to_zero` / `test_activity_completed_none_coerced_to_zero` to lock the behaviour in.

LOCKhart07:
Thanks for the thorough multi-perspective review — all six inline findings plus the four non-line-anchored test gaps are addressed in 3e888937.

**Inline:** replied in-thread. Summary: 🔴 `ValueError` dropped from the contract catch tuple (transient RPC blips now propagate, not cached as OLD); 🟡 activity-field persistence moved below the `NONE/NO_MAJORITY` guard; 🟢 `getServiceInfo()[2]` comment fixed; 🟢 `stop_trading_if_staking_kpi_met` semantic-shift note added; 🟢 dead `is_staking_kpi_met()` removed; 🟢 None-safe getters mirrored from `states/base.py`.

**Test-assertion gaps (100% line coverage was masking these):**
- `async_act` — now captures the emitted payload and asserts all six fields (sender, vote, `is_staking_kpi_met`, `is_activity_target_met`, `activity_target`, `activity_completed`); a swapped kwarg would now fail.
- `end_block` — new `TestEndBlockPersistsActivityFields`: a converged event asserts the `synchronized_data.update(...)` (the vote/KPI decoupling is asserted at the round layer), and a `NO_MAJORITY` event asserts nothing is persisted.
- `staking_abci` — `test_absent_version_is_old` now asserts the cache write (`staking_regime_is_new is False`); added `test_cache_hit_false_skips_reads` covering the `cached is not None` shortcut for a `False` verdict.
- `decision_maker_abci` — `SharedState.__init__` now asserts `staking_regime_is_new is None` (the caching contract depends on `None`, not `False`).

All affected modules remain at 100% coverage; black/isort/flake8/darglint/mypy green; lock idempotent.

packages/valory/skills/decision_maker_abci/handlers.py
```diff
@@ -422,6 +425,9 @@ def _handle_get_health(
                 "is_making_on_chain_transactions": is_mech_reliable,
                 "is_mech_reliable": is_mech_reliable,
                 "is_staking_kpi_met": is_staking_kpi_met,
+                "is_activity_target_met": is_activity_target_met,
+                "activity_target": activity_target,
+                "activity_completed": activity_completed,
```
OjusWiZard, line 430: they aren't really the "health" metrics, right? Should they be outside of "agent_health"?

OjusWiZard (approved):

packages/valory/skills/check_stop_trading_abci/rounds.py
```diff
@@ -159,17 +180,30 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Enum]]:
...
 
+        # The persisted ``is_staking_kpi_met`` must come from the payload, not
+        # the vote: in the new regime the vote tracks the off-chain activity
+        # target while ``is_staking_kpi_met`` stays the on-chain KPI, so the two
+        # can legitimately disagree. Single-agent ⇒ a representative payload is
+        # authoritative.
+        payload = cast(CheckStopTradingPayload, next(iter(self.collection.values())))
+        is_activity_target_met = bool(payload.is_activity_target_met)
+        self.synchronized_data.update(
+            is_staking_kpi_met=bool(payload.is_staking_kpi_met),
```
bennyjo, line 201: Question on old-regime back-compat: here `is_staking_kpi_met` is sourced from the payload's on-chain read, whereas pre-PR it came from `positive_vote_threshold_reached`. Those agree only when `stop_trading_if_staking_kpi_met=True` — with the switch *off* on an old contract, the synced value flips from always-`False` to the real KPI. That in turn changes `sampling`'s multi-bets-fallback / sell-review gating (both now key off `is_activity_target_met`, which in the old regime equals `staking_kpi_met`) and the `/healthcheck` `is_staking_kpi_met` field. The stop *vote* itself stays byte-for-byte, so this only bites the switch-off config. Is that divergence intended? If old contracts should stay identical regardless of the switch, would it be worth sourcing these signals from the vote in the old-regime branch and only using the decoupled payload values when new-regime is detected?

packages/valory/skills/decision_maker_abci/models.py
```diff
@@ -213,6 +213,16 @@ def __init__(self, *args: Any, skill_context: SkillContext, **kwargs: Any) -> No
...
+        # Process-lifetime cache for the staking-regime detection seam
+        # (``staking_abci.behaviours.StakingInteractBaseBehaviour._is_new_staking_regime``).
+        # Lives here, NOT on ``staking_abci``'s SharedState: in the composed agent
+        # there is exactly one live ``context.state`` instance shared by every
+        # sub-skill — ``trader_abci.SharedState`` (MRO includes
+        # ``decision_maker_abci.models.SharedState``). The sub-skill SharedState
+        # subclasses are never instantiated as the live state, so a field placed
+        # there would ``AttributeError`` on the live instance. ``None`` = not yet
+        # detected; set once per process (Pearl restarts on a contract switch).
+        self.staking_regime_is_new: Optional[bool] = None
```
bennyjo, line 225: Question on the cache lifetime: this caches the regime verdict for the whole process, relying on "Pearl restarts on a contract switch." If that ever doesn't hold (a hot-swap, or a non-Pearl deployment reusing this skill), a stale OLD verdict would silently re-enable farming with no signal. Would it be worth keying the cache on the staking-contract address (or invalidating it when the address changes) rather than a bare process-lifetime bool? Genuinely open here — happy to be told the restart guarantee is solid enough that this path can't occur.

bennyjo (commented):
Two follow-up questions on the activity-target decoupling (old-regime signal sourcing + regime-cache lifetime). Both non-blocking. Also: thanks for already narrowing the `_NO_SUCH_FUNCTION` catch in the latest commit — that resolves the earlier `ValueError`/transient-misclassification concern, so no comment there.

packages/valory/skills/decision_maker_abci/handlers.py
LOCKhart07: Fair point — `activity_target` / `activity_completed` / `is_activity_target_met` are operational/rotation signals rather than liveness "health" per se, so `agent_health` isn't a perfect home for them. Leaving them there for now to avoid churning the healthcheck schema in this PR (and `is_staking_kpi_met` already lived under `agent_health`, so they sit next to their sibling). Happy to lift them to a dedicated sibling key (e.g. `staking` / `activity`) in a follow-up if consumers would rather have them out of `agent_health`.

packages/valory/skills/check_stop_trading_abci/rounds.py
LOCKhart07: Intended, and the scope is tighter than it first looks.

The stop **vote is byte-identical in every config**: on a switch-off old contract `stop = stop_trading_if_staking_kpi_met(False) and activity_target_met` = `False`, exactly as before — the agent still never halts on switch-off. The divergence is confined to **old contract + `STOP_TRADING_IF_STAKING_KPI_MET=false`** (non-default; default is `true` in `skill.yaml` and both `service.yaml`), and only in the secondary `sampling` gates that key off `kpi_is_met`:

- `_multi_bets_fallback_allowed`: was always-on (if enabled), now switches off once the real on-chain KPI is met;
- sell-review (`selling_specific` / `should_review_bets`): was a dead path on switch-off (gated on a value pinned to `False`), now can activate once KPI met + position-review enabled.

Those were pinned pre-PR only because a single bool did double duty as both the stop vote *and* the KPI signal — exactly the coupling this PR set out to remove. Re-sourcing them from the vote in the old-regime branch would reintroduce that coupling conditionally, which I'd rather not do. The switch-off semantics ("don't *halt* when the KPI is met") are preserved; what changes is that multi-bets/sell-review now track the actual KPI instead of inheriting the halt-suppression flag — which is the intended decoupling. Calling it intended.

packages/valory/skills/decision_maker_abci/models.py
LOCKhart07: The restart invariant holds for this deployment. `staking_contract_address` is a config param, and changing it requires a process restart — Pearl tears down and re-launches the agent on a contract switch, which clears the process-lifetime cache. There's no path that mutates `staking_contract_address` in a live process, so a stale OLD verdict can't outlive a contract change here.

Agreed it'd be more defensive to key the cache on the contract address for a hypothetical non-Pearl reuse, but given the restart guarantee I'll leave it as the simpler process-lifetime bool for now.

bennyjo (approved):
Approving. The KPI/activity-target decoupling is well-structured and thoroughly tested, and the hardening commit (3e888937) resolved the review feedback — the `ValueError` catch is correctly narrowed (transient RPC blips now propagate), the activity-field persistence is gated behind the NONE/NO_MAJORITY convergence check, the dead `is_staking_kpi_met()` wrapper is gone, and the None-safe getters are now consistent with `states/base.py`.

My two open questions are answered and I'm satisfied:
- Old-regime signal sourcing: confirmed intended — the stop vote is byte-identical in every config; the only divergence is the secondary `sampling` gates under the non-default `STOP_TRADING_IF_STAKING_KPI_MET=false` on old contracts, which is the intended decoupling. (Worth keeping in mind that this rests on no old-contract deployment shipping with that switch off.)
- Regime-cache lifetime: confirmed the Pearl restart invariant holds (`staking_contract_address` is config-only with no live-mutation path), so a stale OLD verdict can't outlive a contract switch.

Nice work.

PR #996, opened 2026-06-19
fix: set the default empty builder code

OjusWiZard:
## What

Set the Polymarket `builder_code` default to the all-zero bytes32
(`0x0000…0000`) instead of an empty string, across the connection,
agent, and service configs.

## Why

open-autonomy's env-var substitution mishandles an **empty** default in
`${POLYMARKET_BUILDER_CODE:str:}` — the empty value is mis-parsed (it
resolves to a malformed literal rather than `""`). Providing an explicit
zero bytes32 default sidesteps that parsing bug while keeping the
"no attribution" semantics: the all-zero code is the SDK's no-attribution
sentinel.

## Note on diagnostics

So the operator-facing diagnostics don't regress, `_validate_builder_code`
now treats an all-zero bytes32 as the empty/"disabled" case and returns
`""`. This preserves the existing
`Builder program enabled but builder_code is empty; orders will be posted
without attribution.` warning path instead of logging a misleading
`Using builder_code=0x00000000…` line when no real code is configured.


iasonrovis (dismissed):

DhairyaPatel7 (dismissed):

packages/valory/connections/polymarket_client/connection.yaml
```diff
@@ -26,7 +26,7 @@ config:
   safe_contract_addresses:
     polygon: 0x
   polymarket_builder_program_enabled: false
-  builder_code: ${POLYMARKET_BUILDER_CODE:str:}
+  builder_code: ${POLYMARKET_BUILDER_CODE:str:0x0000000000000000000000000000000000000000000000000000000000000000}
```
DhairyaPatel7, line 29: Changing this default from "" to the all-zero bytes32 silently changes the diagnostic for a misconfigured operator. In `connection.py::__init__`, `_validate_builder_code` accepts `0x0000...0000` as a valid bytes32 and returns it unchanged, so `if builder_program_enabled and builder_code:` now evaluates True even when the operator has not set POLYMARKET_BUILDER_CODE. The path previously logged 'Builder program enabled but builder_code is empty; orders will be posted without attribution' and now logs 'Builder program enabled. Using builder_code=0x00000000...'  — making it sound like attribution is configured when it is not. Suggested fix: in `_validate_builder_code`, after the hex check, also return "" when the decoded bytes are all zero (e.g. `if int(code, 16) == 0: return ""`), so the existing 'builder_code is empty' warning path is preserved. Same default change is applied in `aea-config.yaml:590` and `service.yaml:400`, and the fix in `_validate_builder_code` would cover all three.

DhairyaPatel7, line 29: PR body is empty. The fix's motivation is not captured anywhere (which downstream consumer was rejecting the empty default? a CI/schema validation? a SDK call?). Please add a one-line rationale to the PR description so reviewers and `git blame` archaeology can answer 'why zero bytes32 instead of ""' without re-deriving the failure mode.

OjusWiZard: Good catch — fixed. `_validate_builder_code` now returns `""` when the decoded bytes32 is all-zero (`if not any(decoded): return ""`), so the all-zero default falls into the existing `builder_program_enabled and builder_code` → False branch and the "builder_code is empty; orders will be posted without attribution" warning is preserved. Since the validator covers all three config defaults, no change needed in the YAMLs. Added a unit test (`test_all_zero_bytes32_returns_empty_no_warning`).

OjusWiZard: Added a "Why" section to the PR description: open-autonomy mis-parses the empty default in `${POLYMARKET_BUILDER_CODE:str:}`, so an explicit zero bytes32 sidesteps the parsing bug while keeping no-attribution semantics.

DhairyaPatel7 (approved):

iasonrovis (approved):

PR #998, opened 2026-06-25
fix: align funds_manager rpc_urls and mechs subgraph with agent-runner env vars

OjusWiZard:
Follow-up to #997. Two more agent-runner env-var mismatches from the named-env-vars refactor (#982), same root cause as the polymarket Safe fix, surfaced by the Polystrat middleware QA run.

## Root cause

On the **agent-runner** path open-aea resolves the agent-level `aea-config.yaml` (not `service.yaml`). open-aea's `replace_with_env_var` only applies the auto-derived component-path env var (e.g. `SKILL_..._ARGS_URL`) for **anonymous** `${type:default}` placeholders; giving a placeholder an explicit **name** (`${NAME:type:default}`) disables that path-key fallback, so it resolves *only* against a bare env var of that exact name and otherwise silently uses its default. #982 named several fields whose names don't match what the runtime exports, dropping them to gnosis/`0x` defaults on Polystrat.

## Fixes (`aea-config.yaml`)

- **`funds_manager.rpc_urls`** used `${RPC_URLS:dict:{"gnosis":...}}`, but the middleware exports per-chain `POLYGON_LEDGER_RPC` / `GNOSIS_LEDGER_RPC` scalars — never a bare `RPC_URLS` dict. On Polystrat it fell back to gnosis-only and `_validate_chain_keys` raised:
  ```
  ValueError: rpc_urls is missing chains declared in fund_requirements: ['polygon']
  ```
  Rebuilt from the per-chain scalars (both chains), matching `service.yaml`.
- **mechs marketplace subgraph `url`** was over-named to `${MECHS_SUBGRAPH_URL:...}` (no such bare var exists), pinning it to `marketplace-gnosis` on every chain. Reverted to the anonymous `${str:...}` form so it resolves via the path-key the middleware provides (`marketplace-polygon` on Polystrat), consistent with its already-anonymous block siblings.

## Regression test

`packages/valory/skills/trader_abci/tests/test_agent_config_resolution.py` resolves the real `aea-config.yaml` with open-aea's own resolver against a representative per-chain env and asserts the result is internally consistent for **both** flavors: `funds_manager` validation passes, the polymarket Safe resolves (never `0x`), and the mechs subgraph follows the active chain. It fails on the pre-fix config and guards against future #982-style name mismatches.

## Verification

- New test passes (polygon + gnosis); fails on the pre-fix config (reproduces the exact `agent_runner.log` error).
- Resolved against the actual middleware `agent.json`: `rpc_urls` now `{gnosis, polygon: <provisioned rpc>}`, mechs `marketplace-polygon`, Safe `0x318b…`.
- All 165 `trader_abci` tests collect; `flake8`/`black`/`isort` clean.
- Hashes refreshed via `autonomy packages lock` (idempotent).

### DIFF
--- a/packages/valory/agents/trader/aea-config.yaml
+++ b/packages/valory/agents/trader/aea-config.yaml
@@ -69,7 +69,7 @@ skills:
 - valory/tx_settlement_multiplexer_abci:0.1.0:bafybeidj2k4sgfgrz6tzaslvygawqfmnop7sqcsboaedqnmyzu7qqs4nkm
 - valory/market_manager_abci:0.1.0:bafybeic67hypwetzltnuy3a6y3o2o332lxyvn34buq57gy4n6bfqpbpxjm
 - valory/decision_maker_abci:0.1.0:bafybeicedj5dtoheq4y5ez4ttw72vzdrpouhwxjgth42eyerpneykkejka
-- valory/trader_abci:0.1.0:bafybeihzu6zmblk4c77dv3z53pz3lyawwdpm3i7yu3czrhr6hdzeowzzyq
+- valory/trader_abci:0.1.0:bafybeidyiphgc3u4x6fn3ghq2fatwgu5khtwnfch5hz6wnlbodn3jdz2rm
 - valory/staking_abci:0.1.0:bafybeihwyron2bv7ierrgxke7xofwjy7x6wtglg5vopon7i7lfdiivweaq
 - valory/check_stop_trading_abci:0.1.0:bafybeifbiko6p3sqqg2q6qemmcig432s5djcsxm44j35ejittlpyr5f6z4
 - valory/mech_interact_abci:0.1.0:bafybeidnmszufszq3ro4ebgrqzmhqgctrsc6sykeanfsbvotghvt2cgfsy
@@ -474,7 +474,7 @@ models:
       error_index: ${int:0}
       error_type: ${str:dict}
       retries: ${int:5}
-      url: ${MECHS_SUBGRAPH_URL:str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-gnosis}
+      url: ${str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-gnosis}
       delivery_rate_cap: ${int:10000000000000000}
   polymarket_agents_subgraph:
     args:
@@ -576,7 +576,9 @@ models:
   params:
     args:
       fund_requirements: ${FUND_REQUIREMENTS:dict:{"gnosis":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":2000000000000000000,"threshold":210000000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":5000000000000000000,"threshold":1000000000000000000},"0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d":{"topup":0,"threshold":0}}}}}
-      rpc_urls: ${RPC_URLS:dict:{"gnosis":"https://rpc.gnosischain.com"}}
+      rpc_urls:
+        gnosis: ${GNOSIS_LEDGER_RPC:str:https://rpc.gnosischain.com}
+        polygon: ${POLYGON_LEDGER_RPC:str:https://1rpc.io/matic}
       safe_contract_addresses: ${SAFE_CONTRACT_ADDRESSES:dict:{"gnosis":"0x0000000000000000000000000000000000000000"}}
 ---
 public_id: valory/polymarket_client:0.1.0:bafybeicblltx7ha3ulthg7bzfccuqqyjmihhrvfeztlgrlcoxhr7kf6nbq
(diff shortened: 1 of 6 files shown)


iasonrovis (approved):

packages/valory/agents/trader/aea-config.yaml
```diff
@@ -576,7 +576,9 @@ models:
   params:
     args:
       fund_requirements: ${FUND_REQUIREMENTS:dict:{"gnosis":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":2000000000000000000,"threshold":210000000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":5000000000000000000,"threshold":1000000000000000000},"0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d":{"topup":0,"threshold":0}}}}}
-      rpc_urls: ${RPC_URLS:dict:{"gnosis":"https://rpc.gnosischain.com"}}
+      rpc_urls:
+        gnosis: ${GNOSIS_LEDGER_RPC:str:https://rpc.gnosischain.com}
+        polygon: ${POLYGON_LEDGER_RPC:str:https://1rpc.io/matic}
```
Tanya-atatakai: why not https://polygon.drpc.org

Tanya-atatakai (dismissed):

DIvyaNautiyal07 (dismissed):

packages/valory/agents/trader/aea-config.yaml
```diff
@@ -474,7 +474,7 @@ models:
       error_index: ${int:0}
       error_type: ${str:dict}
       retries: ${int:5}
-      url: ${MECHS_SUBGRAPH_URL:str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-gnosis}
+      url: ${str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-gnosis}
```
Tanya-atatakai, line 477: not sure exactly how this works/what it breaks (I did read the PR description) but
1) why don't we have the same issue with `gnosis_staking_subgraph`, because we mention chain there in the name?
2) why do we also have `olas_mech_subgraph` variable? looks confusing to have two; maybe three actually, I see there's `polygon_mech_subgraph`

packages/valory/agents/trader/aea-config.yaml
OjusWiZard: it's mostly just a placeholder here, it's overwritten by the `service.yaml`

packages/valory/agents/trader/aea-config.yaml
OjusWiZard: Good questions — both are fine here:

**1) Why `gnosis_staking_subgraph` doesn't break despite naming the chain:** the bug needs two conditions together, and staking has neither.
- It's not chain-varying: Olas staking is on **Gnosis for both** Omenstrat and Polystrat (we stake OLAS on Gnosis regardless of trading venue), so `staking-gnosis…` is correct on both flavors — there's no wrong-chain default to fall back to. `mechs_subgraph` is the mech *marketplace*, which genuinely differs per chain (`marketplace-gnosis` vs `marketplace-polygon`).
- Its name matches a provided var: the middleware exports `GNOSIS_STAKING_SUBGRAPH_URL`, so the named placeholder resolves. The mechs one broke because `MECHS_SUBGRAPH_URL` is never exported — only the auto-derived path-key `SKILL_TRADER_ABCI_MODELS_MECHS_SUBGRAPH_ARGS_URL` is, and giving the placeholder an explicit name disables that path-key fallback (so it silently used the gnosis default). Reverting to the anonymous `${str:...}` form re-enables the path-key, which the middleware provides as `marketplace-polygon` on Polystrat.

**2) The three mech subgraphs are pre-existing and distinct** (added in df8d09804 / 80980b1f1, not this PR) — different response schemas and consumers:
| model | response_key | endpoint | used by |
|---|---|---|---|
| `mechs_subgraph` | `data:meches` | marketplace-<chain> | `mech_interact` — mech **discovery** |
| `olas_mech_subgraph` | `data` | `/proxy/mech` | agent-performance ROI on **Gnosis** |
| `polygon_mech_subgraph` | `data:sender` | marketplace-polygon | agent-performance ROI on **Polygon** |

So `olas_mech_subgraph` / `polygon_mech_subgraph` are per-chain counterparts for the ROI lookups (the code picks one by venue, and they query structurally different subgraphs — `data` vs `data:sender`), while `mechs_subgraph` is the separate discovery one. This PR only reverts `mechs_subgraph`'s placeholder to the anonymous form; consolidating the `olas_`/`polygon_` naming would be a fair separate cleanup if we want it.

packages/valory/agents/trader/aea-config.yaml
Tanya-atatakai: I understand, it's just not the official one for polygon https://docs.polygon.technology/pos/reference/rpc-endpoints

iasonrovis (dismissed):

iasonrovis (approved):

Tanya-atatakai (approved):

PR #999, opened 2026-06-26
feat(polymarket): edge-band + spread-band bet-selection gates (default no-op)

jmoreira-valory:
## Summary

Adds two **optional, default-no-op, Polymarket-only** bet-selection gates to the trader, so an operator can restrict betting to the market conditions an out-of-sample ROI analysis found least-bad. Both gates are inert until explicitly configured via env, so behaviour on `main` is unchanged. The Omen trader (`trader_pearl`) is behaviourally untouched.

### Gates shipped

1. **Edge band** (`customs/kelly_criterion/kelly_criterion.py`) — the CLOB pre-filter is widened from `edge < min_edge` to a band `edge outside [min_edge, max_edge]`. `max_edge` rides inside the existing `strategies_kwargs` dict and defaults to `1.0` (edge = `p − best_ask` is always ≤ 1.0, so `1.0` = no-op). **The FPMM (Omen) edge gate is untouched** — the band is CLOB-only.
2. **Spread band** (`decision_maker_abci/behaviours/decision_receive.py`) — skips a bet when the chosen side's **live-CLOB** bid-ask spread (`best_ask − best_bid`) is outside `[polymarket_spread_min, polymarket_spread_max]`. The gate keeps the CLOB `bids` (previously discarded), fires **after** `strategy_vote` is known (so it reads the correct side's book), and guards a missing/empty book (passes through silently). Defaults `0.0 / 1.0` widen the band to `[0, 1]` = no-op.

New params (`polymarket_spread_min`, `polymarket_spread_max`) are plumbed end-to-end (`decision_maker_abci/skill.yaml`, `trader_abci/skill.yaml`, agent `aea-config.yaml`, `polymarket_trader/service.yaml`, `models.py`). `max_edge` needs no plumbing — it lives in the `STRATEGIES_KWARGS` JSON blob.

### Activation (operator config, Polymarket service only)

```yaml
# spread band, e.g. 0.02–0.05
polymarket_spread_min: ${POLYMARKET_SPREAD_MIN:float:0.02}   # default 0.0
polymarket_spread_max: ${POLYMARKET_SPREAD_MAX:float:0.05}   # default 1.0
# edge band, e.g. 0.05–0.10 — inside STRATEGIES_KWARGS JSON:
#   set "min_edge": 0.05 and add "max_edge": 0.10
```

Defaults stay at the no-op values; operators flip the knobs via env when intentionally experimenting.

## Scope & invariants

- **Polymarket only.** Activation config is in `polymarket_trader/service.yaml`. `trader_pearl/service.yaml` is content byte-identical (the only diff is the cascaded `agent:` CID from `autonomy packages lock`, since both services share the `valory/trader` agent).
- **No FSM change.** Each gate reuses an existing skip path — edge/spread route through `_is_profitable` → `Event.UNPROFITABLE` → `BlacklistingRound`. `autonomy analyse fsm-specs` reports no diff. No new rounds/events/states.
- **Minimum diff.** ~50 LOC of real code across 2 behaviour/strategy files + param plumbing; the rest is tests and lock-cascade fingerprints.

## A note on the dropped horizon gate

This work originally explored a **third** gate — a 5–8-day time-to-resolution horizon floor. It was **removed before this PR** after a close-vs-resolution reconciliation: the +20pp ROI the analysis found lives in *oracle-resolution* time (`res_ts − bet_ts`), which is not observable at bet time, whereas a live gate can only filter on *market close* (`openingTimestamp`). The close-time 5–8d band had ~10 historical bets with 0.000 Jaccard overlap against the resolution-time winners, so a close-time horizon gate would not reproduce the ROI. The horizon machinery is fully reverted here (sampling stage unchanged from `main`); the investigation is retained outside the PR for a possible future revisit.

## Testing

- **3824 / 3824** dev-package unit tests pass (includes new edge-band tests in `test_kelly_criterion.py` and new spread-gate tests in `test_decision_receive.py`, covering on/off and the empty-book guard).
- No-op proof: with defaults, both gates are inert (`max_edge=1.0` never trips the new `or` branch; any spread in `[0,1]` passes).
- Lint pipeline green: `isort`, `black`, `flake8`, `mypy`, `pylint`, `darglint`.
- `autonomy analyse fsm-specs` no diff; `autonomy analyse docstrings` no update; `autonomy packages lock` clean; `check-hash` OK.

## Forward-test caveat (pre-activation, not a merge blocker)

The spread lift in the backing analysis was measured on the **Gamma snapshot** `market_spread`; this gate uses the **live-CLOB** chosen-side `best_ask − best_bid`. The two measures may differ, so the expected lift is not guaranteed to transfer — confirm on a live forward-test before any operator activation of the spread band. Both gates default to no-op, so this is a pre-activation check.

## Live mainnet smoke test

Beyond the unit tests, both gates were validated **end-to-end against live Polygon mainnet** — a real `superforcaster-polymarket-v1` mech prediction + the live CLOB order book — by running the agent via `make run-agent` with bet/collateral txs blocked by guards and a halt that prints the gate verdict (no bet tx is ever sent). This directly exercises the forward-test caveat above: it confirms the spread gate reads and gates on the **live-CLOB** chosen-side `best_ask − best_bid`. Full methodology + reproduction (guards patch, env, EOA-gas gotcha, knob matrix): [`.claude/testing/polymarket-bet-selection-gates-smoke-test.md`](.claude/testing/polymarket-bet-selection-gates-smoke-test.md).

**Method:** one live market sampled (`2363872`), three runs changing **one knob at a time** (the other left at its no-op default), reading the `BET DECISION` verdict each run.

| Run | `min_edge` | spread band | mech p_yes | measured edge | measured spread | Edge gate | Spread gate | Verdict |
|---|---|---|---|---|---|---|---|---|
| A — baseline | 0.01 | [0, 1] | 0.98 | **+0.66** | 0.17 | PASS | PASS | **WOULD PLACE BET (YES)** |
| B — edge-tight | **0.99** | [0, 1] | 0.97 | +0.54 | — | **REJECT** | n/a | **WOULD NOT PLACE A BET** |
| C — spread-tight | 0.01 | **[0, 0.0001]** | 0.97 | pass | **0.17** | PASS | **REJECT** | **WOULD NOT PLACE A BET** |

Each gate **independently flips `bet → no-bet`** when its own knob is tightened (other knob off), confirming both bands genuinely gate against the live book and that the defaults are inert. Guards held every run: zero bet/collateral txs, only the authorized mech requests settled. The gate code needed no change.


packages/valory/skills/decision_maker_abci/models.py
```diff
@@ -431,6 +431,15 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             msg = "The number of days to sample bets from must be positive!"
             raise ValueError(msg)
 
+        # Optional, Polymarket-only live-CLOB bid-ask spread band. Defaults
+        # 0.0 / 1.0 widen the band to the full [0, 1] range = no-op.
+        self.polymarket_spread_min: float = self._ensure(
+            "polymarket_spread_min", kwargs, float
+        )
+        self.polymarket_spread_max: float = self._ensure(
+            "polymarket_spread_max", kwargs, float
```
OjusWiZard, line 440: **[Important] Inverted spread band silently rejects every bet.** If an operator sets `POLYMARKET_SPREAD_MIN=0.05, POLYMARKET_SPREAD_MAX=0.02`, the gate's `not lo <= spread <= hi` is unsatisfiable → every CLOB bet skipped, only a per-bet `info` log, no startup error. The agent looks healthy and never trades. The class already has the precedent (`sample_bets_closing_days` raises `ValueError`). Add after the two `_ensure` calls:

```python
if self.polymarket_spread_min > self.polymarket_spread_max:
    msg = (
        f"polymarket_spread_min ({self.polymarket_spread_min}) must be "
        f"<= polymarket_spread_max ({self.polymarket_spread_max})"
    )
    raise ValueError(msg)
```

(Optionally also range-check both into `[0, 1]`, mirroring the `slippage` setter.)

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -347,12 +352,14 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
             b_min_side = min_bet
             x_native, y_native = 0.0, 0.0
 
-            # CLOB pre-filter: quick edge check against best ask
+            # CLOB pre-filter: quick edge check against best ask. Edge must
+            # fall inside the [min_edge, max_edge] band; defaults
+            # (0.03, 1.0) reduce to the original floor-only check.
             edge_best_ask = p - best_ask_price
-            if edge_best_ask < min_edge:
+            if edge_best_ask < min_edge or edge_best_ask > max_edge:
```
OjusWiZard: **[Important] Inverted edge band silently rejects every CLOB bet** (same algebra as the spread gate). `min_edge=0.10, max_edge=0.05` via `STRATEGIES_KWARGS` makes this condition always true; the rejection only lands in the per-side `info` list. Validate right after `max_edge` is read (line 280) and surface via the existing `_no_trade` helper so it logs at `error` level:

```python
if min_edge > max_edge:
    error.append(f"min_edge ({min_edge}) > max_edge ({max_edge}): no valid edge band")
    return _no_trade(info, error)
```

**[Advisory / simplify]** consider the chained form to match the spread gate's idiom: `if not min_edge <= edge_best_ask <= max_edge:`. Also the inline comment above hardcodes `(0.03, 1.0)` — reference `DEFAULT_MIN_EDGE`/`DEFAULT_MAX_EDGE` to avoid drift.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -518,6 +522,31 @@ def _is_profitable(
...
+        if market_type == "clob":
+            spread_asks = (
+                orderbook_asks_yes if strategy_vote == 0 else orderbook_asks_no
+            )
+            spread_bids = (
+                orderbook_bids_yes if strategy_vote == 0 else orderbook_bids_no
+            )
+            if spread_asks and spread_bids:
+                best_ask = min(float(a["price"]) for a in spread_asks)
+                best_bid = max(float(b["price"]) for b in spread_bids)
```
OjusWiZard: **[Important] Uncaught `ValueError` can stall the round.** `connection.py:2010` builds levels as `"price": str(level.get("price"))`, so a null price becomes the string `"None"` and `float("None")` raises `ValueError` here — uncaught up the chain, which blocks payload submission and times out the round. `KeyError` cannot occur (key is always present). The asks side mirrors pre-existing kelly code (`float(level["price"])`), but **bids are newly parsed here**. Low probability, but a `try/except` around these two comprehensions (log + pass-through) closes it cheaply and symmetrically.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -518,6 +522,31 @@ def _is_profitable(
             )
             return False, 0, None
 
+        # Optional spread gate (Polymarket / CLOB only): skip if the chosen
+        # side's live bid-ask spread is outside
+        # ``[polymarket_spread_min, polymarket_spread_max]``. Defaults
+        # 0.0 / 1.0 widen the band to [0, 1] = no-op. A missing/empty book
+        # passes through silently (the strategy already approved the bet).
+        if market_type == "clob":
+            spread_asks = (
```
OjusWiZard: **[Advisory] "no-op" comment is inaccurate for crossed books.** With defaults `[0.0, 1.0]`, a crossed book (`best_bid > best_ask` → `spread < 0`) fails `0.0 <= spread`, so the default *does* reject it — not strictly a no-op. Rejecting a broken market may be intended, but the comment should say so: e.g. "passes well-formed books (spread in [0,1]); crossed books (spread < 0) are still rejected." Decide intent, then make code and comment agree.

**[Advisory / simplify]** extract the repeated condition once: `is_yes = strategy_vote == 0`, matching the existing `side_name` idiom at line 554.

**[Advisory / visibility]** when the band is non-default but the book is missing, the gate is skipped silently — exactly the illiquid market an operator wants caught. Consider a `warning` log in the `else` of `if spread_asks and spread_bids:`.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
```diff
@@ -1105,6 +1107,129 @@ def test_fpmm_no_orderbook_fetch(self) -> None:
...
+        behaviour, pred = self._setup_clob_with_book(
+            ob, spread_min=0.02, spread_max=0.05
+        )
+        result = self._run_is_profitable(behaviour, pred)
+        patch.stopall()
+
+        is_profitable, _, _ = result
+        assert is_profitable is True
+
+    def test_spread_gate_no_side_no_book(self) -> None:
```
OjusWiZard, line 1237: **[Important] Side-selection by `strategy_vote` is never semantically tested.** This is the only `strategy_vote=1` case, but it uses an empty-bids book so the `if spread_asks and spread_bids:` guard short-circuits *before* the side ternary matters — and `_setup_clob_with_book` returns the same book for both tokens anyway. A regression swapping `strategy_vote == 0` → `== 1` in the two ternaries would route the gate to the wrong side's book on every real trade and **all tests would still pass**. Add a test with distinct YES/NO books (one narrow, one wide spread) + a tight band, asserting the verdict flips with the vote.

packages/valory/skills/decision_maker_abci/tests/test_models.py
```diff
@@ -580,6 +580,8 @@ def _build_decision_maker_params_kwargs() -> dict:
         "agent_registry_address": "0xaddr",
         "mech_marketplace_v1_suitable_tools": [],
         "sample_bets_closing_days": 7,
+        "polymarket_spread_min": 0.0,
+        "polymarket_spread_max": 1.0,
```
OjusWiZard, line 584: **[Important] New params are supplied but never asserted in `test_full_init`.** A typo'd attribute name in `models.py` (e.g. `polymarket_spread_minimum`) would pass 100% coverage and only surface as an `AttributeError` at runtime. Add two assertions to `test_full_init`:

```python
assert params.polymarket_spread_min == 0.0
assert params.polymarket_spread_max == 1.0
```

OjusWiZard (commented):
Automated multi-perspective review (code / tests / comments / silent-failure / type-design / simplify). No critical bugs — defaults are inert for well-formed books, `vote==0 → YES` side selection is correct, and plumbing is consistent. Two themes worth fixing before merge: (1) an inverted band on **either** gate silently turns the agent into a no-trade zombie with no startup error, and (2) two test gaps let a side-selection regression ship green. Inline comments below.

packages/valory/skills/decision_maker_abci/models.py
```diff
@@ -431,6 +431,15 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        )
```
DIvyaNautiyal07, line 441: **P1 — Inverted band silently rejects every bet.**

`_ensure` validates that both values are floats but not that `polymarket_spread_min <= polymarket_spread_max`. If an operator sets `POLYMARKET_SPREAD_MIN=0.5` and `POLYMARKET_SPREAD_MAX=0.1`, then in `decision_receive.py:543` the check `not lo <= spread <= hi` is `True` for every real spread — every bet is silently skipped, with no startup error and no log distinguishing 'band inverted' from 'spread outside band'. Ops would only notice when they see zero bets being placed.

Suggested fix: after line 441, add
```python
if self.polymarket_spread_min > self.polymarket_spread_max:
    raise ValueError(
        f"polymarket_spread_min ({self.polymarket_spread_min}) must "
        f"be <= polymarket_spread_max ({self.polymarket_spread_max})"
    )
```

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -273,6 +277,7 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
     fee_per_trade_wei: int = kwargs.get("fee_per_trade", default_fee)
     n_bets: int = kwargs.get("n_bets", DEFAULT_N_BETS)
     min_edge: float = kwargs.get("min_edge", DEFAULT_MIN_EDGE)
+    max_edge: float = kwargs.get("max_edge", DEFAULT_MAX_EDGE)
```
DIvyaNautiyal07, line 280: **P1 — Same inverted-band trap for the edge band.**

No check that `max_edge >= min_edge`. If an operator passes `max_edge < min_edge` (e.g. swapped values in `STRATEGIES_KWARGS`), the condition at line 359 (`edge_best_ask < min_edge or edge_best_ask > max_edge`) is `True` for every finite edge and every CLOB bet is silently rejected.

Suggested fix: short-circuit via the existing `_no_trade` helper (mirrors the early-return at line 306):
```python
if max_edge < min_edge:
    return _no_trade(info, error,
        f"max_edge ({max_edge}) < min_edge ({min_edge})")
```

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
```diff
@@ -1105,6 +1107,129 @@ def test_fpmm_no_orderbook_fetch(self) -> None:
...
+            polymarket_spread_max=spread_max,
+        )
+        patch.object(behaviour, "convert_to_native", return_value=0.001).start()
+        patch.object(behaviour, "get_token_name", return_value="USDC").start()
+        patch.object(behaviour, "rebet_allowed", return_value=True).start()
+        patch.object(
+            behaviour,
+            "_fetch_orderbook",
+            side_effect=lambda _tid: _return_gen(ob),
+        ).start()
```
DIvyaNautiyal07: **P1 — YES/NO side-selection is not actually tested.**

`_fetch_orderbook` is patched with `side_effect=lambda _tid: _return_gen(ob)` — the same `ob` is returned regardless of which token id is passed. That means a refactor that swapped `strategy_vote == 0` to `strategy_vote == 1` in `decision_receive.py:532, 535` (reading the wrong side's book) would pass every spread-gate test in this file, including `test_spread_gate_no_side_no_book` (line 1220) which superficially looks like a side-selection test but only asserts the gate is skipped — not that the *NO* book was the one consulted.

Suggested fix: keep distinct `ob_yes` and `ob_no` dicts (with different spreads), key the `side_effect` on the token id, and assert that the gate fires based on the chosen side's book. Optional helper:
```python
books = {yes_token_id: ob_yes, no_token_id: ob_no}
patch.object(behaviour, "_fetch_orderbook",
    side_effect=lambda tid: _return_gen(books[tid])).start()
```

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
```diff
@@ -1105,6 +1107,129 @@ def test_fpmm_no_orderbook_fetch(self) -> None:
...
+        behaviour, pred = self._setup_clob_with_book(
+            ob, spread_min=0.02, spread_max=0.05
+        )
+        result = self._run_is_profitable(behaviour, pred)
+        patch.stopall()
+
+        is_profitable, bet_amount, strategy_vote = result
+        assert is_profitable is False
+        assert bet_amount == 0
+        assert strategy_vote is None
```
DIvyaNautiyal07, line 1207: **P1 — Lower-bound `spread < spread_min` is not covered.**

Only the above-band case is asserted here (`spread=0.10` with `spread_max=0.05`). The production condition at `decision_receive.py:543` is `not lo <= spread <= hi`, which also rejects when `spread < lo`. A refactor that accidentally dropped the floor check (e.g. `lo <= spread <= hi` → `spread <= hi`) would pass every test in this file.

Add a below-band case, e.g. `spread=0.01` with `spread_min=0.02, spread_max=0.05`, asserting `(False, 0, None)`.

DIvyaNautiyal07 (commented):
Two findings below sit on lines outside any diff hunk, so GitHub can't anchor them inline. Captured here so they don't get lost.

---

**P2 — New knobs are undocumented for operators** (`README.md` line ~253, `STRATEGIES_KWARGS` table)

The `STRATEGIES_KWARGS` key table omits `max_edge`. The new `POLYMARKET_SPREAD_MIN` / `POLYMARKET_SPREAD_MAX` env vars aren't documented anywhere operator-facing either. Since both gates are default-no-op and only useful when operators flip them, the value of the PR depends on operators knowing the knobs exist. Add `max_edge` to the `STRATEGIES_KWARGS` table and add the spread env vars to the env-var section.

---

**P2 — `max_edge` missing from the diagnostic dump** (`kelly_criterion.py` line ~293)

The startup `info.append` line `f"max_bet: {max_bet}, n_bets: {n_bets}, min_edge: {min_edge}"` was not updated to include `max_edge`. An operator configuring a narrow band (e.g. `max_edge=0.15`) won't see it confirmed in the run info until the rejection message fires. One-token fix:

```diff
- info.append(f"max_bet: {max_bet}, n_bets: {n_bets}, min_edge: {min_edge}")
+ info.append(f"max_bet: {max_bet}, n_bets: {n_bets}, "
+     f"min_edge: {min_edge}, max_edge: {max_edge}")
```

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -347,12 +352,14 @@ def run(**kwargs: Any) -> Dict[str, Any]:  # pylint: disable=too-many-locals
...
+            # CLOB pre-filter: quick edge check against best ask. Edge must
+            # fall inside the [min_edge, max_edge] band; defaults
+            # (0.03, 1.0) reduce to the original floor-only check.
             edge_best_ask = p - best_ask_price
-            if edge_best_ask < min_edge:
+            if edge_best_ask < min_edge or edge_best_ask > max_edge:
                 msg = (
                     f"{label}: edge vs best_ask "
-                    f"{edge_best_ask:+.4f} < min_edge {min_edge}"
+                    f"{edge_best_ask:+.4f} outside [{min_edge}, {max_edge}]"
```
DIvyaNautiyal07: **P2 — Rejection message loses the floor/ceiling distinction.**

Pre-PR: `"edge {edge} < min_edge {min_edge}"`. New: the same `outside [{min_edge}, {max_edge}]` string fires whether the floor was tripped or the ceiling was tripped. With default `max_edge=1.0`, an operator who only tightened `min_edge` will see `+0.0200 outside [0.03, 1.0]` instead of the directional `< min_edge`. The FPMM path at line 381 still uses the original specific phrasing — divergent diagnostics for the same logical event. Any external grep/alerting that matches `< min_edge` silently breaks.

Branch the message by which bound was crossed.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -518,6 +522,31 @@ def _is_profitable(
...
+                best_bid = max(float(b["price"]) for b in spread_bids)
+                spread = best_ask - best_bid
+                lo = self.params.polymarket_spread_min
+                hi = self.params.polymarket_spread_max
+                if not lo <= spread <= hi:
+                    self.context.logger.info(
+                        f"Spread gate: {spread:.4f} outside "
+                        f"[{lo}, {hi}] — skipping bet."
+                    )
+                    return False, 0, None
```
DIvyaNautiyal07: **P2 — Crossed-book under the default `[0.0, 1.0]` band is not a no-op.**

A crossed book (`best_bid > best_ask`) produces `spread < 0`, and `not 0.0 <= -0.001 <= 1.0` is `True` → the bet is silently rejected even with the documented "no-op" defaults. Crossed books are rare on Polymarket but can occur (e.g. near resolution or briefly during liquidity flux). An operator running on the default config will quietly drop those bets without realising the gate is filtering anything.

Options: (a) set the default `polymarket_spread_min=-1.0` so any real spread passes; (b) special-case `spread < 0` with a distinct log line; or (c) update the comment / PR description to state explicitly that the gate intentionally treats crossed books as bad data.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -518,6 +522,31 @@ def _is_profitable(
...
+        # 0.0 / 1.0 widen the band to [0, 1] = no-op. A missing/empty book
+        # passes through silently (the strategy already approved the bet).
+        if market_type == "clob":
+            spread_asks = (
+                orderbook_asks_yes if strategy_vote == 0 else orderbook_asks_no
+            )
+            spread_bids = (
+                orderbook_bids_yes if strategy_vote == 0 else orderbook_bids_no
+            )
+            if spread_asks and spread_bids:
```
DIvyaNautiyal07, line 544: **P2 — Falsy guard collapses three different failure modes into one silent skip.**

`if spread_asks and spread_bids:` treats (a) `_fetch_orderbook` returning `None` on a network failure, (b) the API returning an empty `bids: []` for a legitimately one-sided book, and (c) the side never being fetched (missing token id) identically — gate silently skipped, no log. `_fetch_orderbook` already logs a WARNING on (a) at line 430, so the signal isn't entirely lost, but the gate's silent-skip doesn't link the two events.

If an operator tightens the spread band expecting protection, a network outage silently lifts the gate (the opposite of the operator's intent). Consider an INFO log on the bypass with the reason, e.g. "spread gate bypassed: chosen-side book unavailable". See `trace-falsy-signal-semantics`: a single falsy guard shouldn't collapse semantically distinct upstream signals.

packages/valory/customs/kelly_criterion/kelly_criterion.py
```diff
@@ -64,6 +65,9 @@
 DEFAULT_MIN_BET = 1
 DEFAULT_N_BETS = 1
 DEFAULT_MIN_EDGE = 0.03
+# Optional CLOB-only upper bound on edge. edge = p_oracle - best_ask is
+# always <= 1.0, so 1.0 widens the band to the full range = no-op.
+DEFAULT_MAX_EDGE = 1.0
```
DIvyaNautiyal07, line 70: **P2 — `max_edge` plumbing is asymmetric with `polymarket_spread_*`.**

The spread band gets first-class typed model fields validated via `_ensure` at `models.py:436-441`. `max_edge` rides inside the opaque `STRATEGIES_KWARGS` dict: no type check, no `_ensure`, silent fallback to `1.0` on typo (`max_egde: 0.1` is dropped without error). Operators tuning the band have asymmetric safety nets depending on which gate they're touching.

Options: (a) hoist `max_edge` to a model param parallel to the spread band, or (b) at minimum add a runtime assertion in `run()` that `max_edge >= min_edge` (covered by the separate P1 above, but worth pairing here).

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
```diff
@@ -1105,6 +1107,129 @@ def test_fpmm_no_orderbook_fetch(self) -> None:
...
+        ).start()
+
+        def _mock_get_bet_amount(*_a: Any, **_k: Any) -> Generator:
+            yield
+            return 500  # type: ignore[return-value]
+
+        patch.object(
+            behaviour, "get_bet_amount", side_effect=_mock_get_bet_amount
+        ).start()
+        return behaviour, pred
```
DIvyaNautiyal07, line 1177: **Nit — `_setup_clob_with_book` duplicates `_setup_behaviour` and they've already silently diverged.**

This helper repeats ~10 `patch.object` calls from `_setup_behaviour` (line 808-864): `benchmarking_mode`, `sampled_bet`, `params`, `convert_to_native`, `get_token_name`, `rebet_allowed`, and the `get_bet_amount` generator. The two helpers have already silently diverged — `_setup_behaviour` sets `g_improvement` in `_last_strategy_result` (line 828) but `_setup_clob_with_book` doesn't (line 1125).

Delegate to `_setup_behaviour(is_polymarket=True, strategy_vote=strategy_vote)` and only override what's CLOB-specific (the orderbook patch + spread params).

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -518,6 +522,31 @@ def _is_profitable(
...
+        # ``[polymarket_spread_min, polymarket_spread_max]``. Defaults
+        # 0.0 / 1.0 widen the band to [0, 1] = no-op. A missing/empty book
+        # passes through silently (the strategy already approved the bet).
+        if market_type == "clob":
+            spread_asks = (
+                orderbook_asks_yes if strategy_vote == 0 else orderbook_asks_no
+            )
+            spread_bids = (
+                orderbook_bids_yes if strategy_vote == 0 else orderbook_bids_no
+            )
```
DIvyaNautiyal07, line 540: **Nit — collapse the two parallel ternaries into one tuple unpack.**

```python
spread_asks, spread_bids = (
    (orderbook_asks_yes, orderbook_bids_yes)
    if strategy_vote == 0
    else (orderbook_asks_no, orderbook_bids_no)
)
```
Makes the side-paired structure explicit and avoids the risk of the two ternaries drifting in a future edit.

dagacha:
Reviewed the branch locally (ran the suite — 202 passing, 100% coverage on the 3 changed modules; `packages lock` and `analyse fsm-specs` both clean). Nice, tightly-scoped change. Three things worth addressing before activation:

**1. (medium) Crossed/locked book trips the spread gate even at defaults — breaks the no-op invariant.**
In `decision_receive.py` the gate is `if not lo <= spread <= hi:` with `spread = best_ask − best_bid` and defaults `lo=0.0, hi=1.0`. On a crossed/locked book (`best_bid > best_ask`, a real transient CLOB condition) `spread < 0`, so `0.0 <= spread` is false and the bet is skipped **even with the default `polymarket_spread_min=0.0`**. That contradicts the stated "any spread in `[0,1]` passes" guarantee.
Impact is limited (the skip routes to `BlacklistingRound`, which only soft-bumps `queue_status` — recoverable, re-queued next cycle), so it's a transient missed-bet, not fund risk. Suggest skipping the gate when `spread < 0` (a crossed book is a buy opportunity, not a wide one), e.g.:

```python
if spread >= 0 and not lo <= spread <= hi:
    ...
```

plus a dedicated test alongside the empty-book one.

**2. (low) No ordering validation on the bands.** `models.py` ensures both spread params are floats but doesn't check `spread_min <= spread_max` (unlike the adjacent `sample_bets_closing_days`, which raises). A misconfigured `MIN=0.5 / MAX=0.1` silently rejects every bet; same for `min_edge > max_edge` in kelly. A guard + log would prevent a silent "agent stopped betting" incident. Not a blocker since defaults are safe.

**3. (nit) Broken doc link.** The description links to `.claude/testing/polymarket-bet-selection-gates-smoke-test.md`, but that file isn't included in the PR (and `.claude/` isn't tracked), so the methodology/repro link is dead for reviewers. Either add the file or drop the link.

packages/valory/skills/decision_maker_abci/models.py
jmoreira-valory: Done in e0100470e — added the `ValueError` guard in `DecisionMakerParams.__init__` right after the two `_ensure` calls, exactly as suggested. Covered by a new `test_inverted_spread_band_raises`.

packages/valory/customs/kelly_criterion/kelly_criterion.py
jmoreira-valory: Done in e0100470e — early `_no_trade` when `min_edge > max_edge`, appending to the `error` list so it surfaces at error level. Also adopted the chained form `not min_edge <= edge_best_ask <= max_edge` and switched the comment to reference `DEFAULT_MIN_EDGE`/`DEFAULT_MAX_EDGE`. New `test_inverted_edge_band_rejected`.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
jmoreira-valory: Done in e0100470e — wrapped both `best_ask`/`best_bid` comprehensions in `try/except (ValueError, TypeError)`; a null/unparseable price now logs a warning and bypasses the gate (pass-through) instead of stalling the round. New `test_spread_gate_unparseable_price_bypasses`.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
jmoreira-valory: Done in e0100470e — (1) the comment now states crossed books (spread < 0) are intentionally rejected as bad data; (2) extracted `is_yes = strategy_vote == 0` and a tuple-unpack; (3) added a `warning` when the gate is bypassed (missing book) under a non-default band.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
jmoreira-valory: Done in e0100470e — `test_spread_gate_side_selection_follows_vote` uses distinct YES/NO books keyed on the token id (one wide, one narrow spread) with a tight band, asserting the verdict flips with `strategy_vote`. The helper now keys the `_fetch_orderbook` mock on the token id.

packages/valory/skills/decision_maker_abci/tests/test_models.py
jmoreira-valory: Done in e0100470e — added `assert params.polymarket_spread_min == 0.0` / `== 1.0` to `test_full_init`, plus `test_inverted_spread_band_raises` for the new guard.

packages/valory/skills/decision_maker_abci/models.py
jmoreira-valory: Done in e0100470e — `ValueError` raised in `Params.__init__` when `polymarket_spread_min > polymarket_spread_max`, as suggested.

packages/valory/customs/kelly_criterion/kelly_criterion.py
jmoreira-valory: Done in e0100470e — early `_no_trade` (via the `error` list) when `min_edge > max_edge`.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
jmoreira-valory: Done in e0100470e — the orderbook mock is now keyed on the token id (`{yes_token_id: ob_yes, no_token_id: ob_no}`) with distinct spreads, and the test asserts the verdict flips with the vote.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
jmoreira-valory: Done in e0100470e — added `test_spread_gate_below_band_skips` (spread 0.01, band [0.02, 0.05]) asserting `(False, 0, None)`.

packages/valory/customs/kelly_criterion/kelly_criterion.py
jmoreira-valory: Done in e0100470e — the CLOB rejection message now branches by bound (`< min_edge {min_edge}` / `> max_edge {max_edge}`), so the floor case matches the FPMM phrasing again.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
jmoreira-valory: Resolved by clarifying intent (your option c): crossed books are deliberately rejected as bad data. Updated the code comment and the README to say so; behaviour unchanged. (e0100470e)

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
jmoreira-valory: Done in e0100470e — when the chosen-side book is unavailable under a non-default band, the gate now logs a `warning` ("Spread gate bypassed: chosen-side order book unavailable …") so the bypass is no longer silent.

packages/valory/customs/kelly_criterion/kelly_criterion.py
jmoreira-valory: The runtime-safety half (`max_edge >= min_edge`) is covered by the inverted-band `_no_trade` guard. Hoisting `max_edge` to a first-class model param parallel to the spread band is deferred to the ChatUI-homogenization follow-up (#1000), which reworks both knobs together — keeping #999 scoped.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
jmoreira-valory: Partially — I extended `_setup_clob_with_book` to support distinct YES/NO books (needed for the side-selection test). I kept it standalone rather than delegating to `_setup_behaviour` for now, to avoid reworking the patching mid-coverage. The divergence you flagged is fair and a good follow-up cleanup.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
jmoreira-valory: Done in e0100470e — collapsed to a single tuple-unpack: `spread_asks, spread_bids = (… if is_yes else …)`.

jmoreira-valory:
Thanks for the thorough review — all findings are addressed in `e0100470e` (lint follow-up in `b44cf78f3`), and I've replied inline on each thread. Summary:

**Inverted-band guards (both gates)** — the spread band raises `ValueError` at param init; the edge band returns an error-level `_no_trade`. No more silent no-trade with a healthy-looking agent.

**Diagnostics** — edge-rejection message is directional again (`< min_edge` / `> max_edge`, matching the FPMM path); `max_edge` added to the startup dump; the comment references `DEFAULT_MIN_EDGE`/`DEFAULT_MAX_EDGE`.

**Spread gate** — `try/except` around price parsing (a null price no longer stalls the round); a `warning` is logged when the gate is bypassed under a non-default band; `is_yes` + tuple-unpack; crossed-book handling clarified (intentionally rejected as bad data) in both the code comment and the README.

**Tests** — side-selection now asserts the verdict follows `strategy_vote` via distinct YES/NO books keyed on the token id; below-band, parse-bypass, empty-book and inverted-band-raise cases added; the new params are asserted in `test_full_init`. 100% statement + branch coverage retained.

**Docs** — `max_edge` and `POLYMARKET_SPREAD_MIN`/`POLYMARKET_SPREAD_MAX` documented in the README.

**Deferred:** hoisting `max_edge` to a first-class model param (parallel to the spread band) is left to the ChatUI-homogenization follow-up (#1000), which reworks both knobs together; the runtime-safety concern is covered here by the inverted-band guard.

dagacha:
Re-reviewed after `e0100470` + `b44cf78f` — thanks for the thorough turnaround. Verified locally on the updated branch: **209 tests passing** (+7), **100% coverage** on all three changed modules, `packages lock` clean, `analyse fsm-specs` no diff, and both new commits stay scoped to the gate files + tests + docs + hash cascade.

All three of my findings are resolved:

- **#1 (crossed book)** — resolved the other way (and better): crossed books (`spread < 0`) are now *intentionally* rejected as bad data, with the code comment + README both stating that intent, so the behaviour is no longer a surprise. Re-ran the predicate matrix to confirm the default band passes well-formed books and only rejects crossed/out-of-band ones. 👍
- **#2 (band ordering)** — fully fixed on both gates: `models.py` raises `ValueError` at init for an inverted spread band, kelly returns an `error`-level `_no_trade` for `min_edge > max_edge`, both with dedicated tests.
- **#3 (broken doc link)** — still open: the description links to `.claude/testing/polymarket-bet-selection-gates-smoke-test.md`, which isn't in the tree (`.claude/` is untracked), so it 404s for reviewers. Cosmetic, non-blocking — just drop the link or add the file.

LGTM once the dead link is dropped/added. Nice work hardening the gates.

DIvyaNautiyal07 (approved):
Re-reviewed the hardening commits (e010047, b44cf78). All P1s from the previous round are resolved with new regression tests; P2s and one nit are addressed too. Approving — one trivial follow-up nit below, non-blocking.

**Resolved**

- *Inverted spread band* — `models.py:442-447` raises `ValueError` at param init; `test_inverted_spread_band_raises` covers it.
- *Inverted edge band* — `kelly_criterion.py:296-302` short-circuits via `_no_trade()`; `test_inverted_edge_band_rejected` covers it.
- *Untested YES/NO side-selection* — `_setup_clob_with_book` now keys `_fetch_orderbook` on token id (`books = {"token_yes_123": ob, "token_no_123": ob_no}`, line 1163) and `test_spread_gate_side_selection_follows_vote` flips the vote across distinct YES/NO books and asserts the verdict flips with it. Mutation gap closed.
- *Untested spread lower-bound* — `test_spread_gate_below_band_skips` covers `spread=0.01, spread_min=0.02`.
- *Operator-facing docs* — `README.md` adds `max_edge` to the `STRATEGIES_KWARGS` table and a dedicated `POLYMARKET_SPREAD_MIN/MAX` section with the "agent refuses to start on inverted band" guarantee.
- *`max_edge` in diagnostic dump* — included in the startup `info.append` at `kelly_criterion.py:292-295`.
- *Rejection-message regression* — `kelly_criterion.py:368-379` now branches by which bound tripped; the `< min_edge` phrasing matches the FPMM path.
- *Crossed-book "no-op" mismatch* — behavior unchanged, but now explicit: README documents it, code comment calls out that crossed books are intentionally rejected as bad data, and `test_spread_gate_crossed_book_rejected_under_default` pins it.
- *Silent bypass on fetch failure* — `decision_receive.py:561-566` logs WARNING on bypass only when a non-default band is configured; default-config users stay quiet, operators who tightened the band hear about it. Good compromise.
- *`max_edge` plumbing asymmetry* — addressed via runtime check in `run()` (option b from the original review).
- *Two-ternary collapse* — `decision_receive.py:535-540` is a single tuple unpack.

**Acceptable to defer**

- `_setup_clob_with_book` still duplicates `_setup_behaviour`. Fine to leave for a separate cleanup PR.

**New nit (non-blocking)**

`decision_receive.py:548` catches `(ValueError, TypeError)` but not `KeyError`. The new `test_spread_gate_unparseable_price_bypasses` covers the null-price case (string `"None"` -> `float("None")` raises `ValueError`), but if a book level dict ever lacks the `"price"` key, `a["price"]` raises `KeyError` and the gate crashes the round instead of bypassing. Trivial one-token addition:

```python
except (ValueError, TypeError, KeyError) as exc:
```

Approving.

OjusWiZard (approved):
Approving after a multi-perspective re-review of the hardened code (HEAD b44cf78f3). Two prior hardening rounds closed all consensus issues: both gates now validate `min <= max`, the order-book `float()` parse is wrapped in try/except, side-selection is covered by a discriminative test, and crossed-book/inverted-band cases are tested. Code-review, silent-failure, type-design, and simplify passes found nothing at the reporting threshold. Remaining items are non-blocking polish (warning-branch test assertions, one stale `models.py` comment, log-context tweaks). LGTM.

PR #1000, opened 2026-06-29
feat(polymarket): expose edge + spread gate knobs via ChatUI

jmoreira-valory:
## Summary

Follow-up to #999. Makes the two Polymarket bet-selection gates tunable at **runtime via the chat UI**, in addition to env config. Both still default to no-op — nothing changes unless an operator sets a value.

## Design

Both knobs become `ChatuiConfig` fields (`None` = keep the existing default), exposed in the chat LLM prompt and validated in the handler — **gated on `is_running_on_polymarket`** (the gates are CLOB-only, so they're offered only on the Polymarket trader; `trader_pearl`/Omen is untouched).

Each knob stays in its **semantic home** (deliberately *not* crammed into a shared `strategies_kwargs`):

| Gate knob | Nature | How ChatUI reaches it |
|---|---|---|
| `min_edge` / `max_edge` | strategy input (kelly reads it) | bridged into `strategies_kwargs` in `decision_maker_abci/behaviours/base.py::set_chatui_values` — same path as `max_bet_size` |
| `min_spread` / `max_spread` | post-strategy gate (decision_receive) | read in `decision_receive.py` with fallback to the existing `polymarket_spread_*` params |

So the operator tunes both the same way (one chat surface), while internally each lives where it belongs and no strategy contract is touched.

## Changes
- **`chatui_abci`**: `models.py` (+4 `ChatuiConfig` fields), `prompts.py` (Polymarket-only gate section in the LLM template), `handlers.py` (typed config + validation: numeric, `0 <= min <= max <= 1`, clearable via `removed_config_fields`).
- **`decision_maker_abci`**: `base.py` (edge bridge), `decision_receive.py` (spread chatui-with-fallback).
- Tests added for the edge bridge, the spread fallback/override, and handler validation.

## Testing
- 4182 unit tests pass; lint green (isort/black/flake8/mypy/pylint/darglint); `autonomy packages lock` + `check-hash` clean.
- **No FSM change.** Defaults unchanged (no-op).

Stacked on `feat/polymarket-bet-selection-gates` (#999).


jmoreira-valory:
Deprioritized.

PR #1003, opened 2026-07-01
test: guard strategy-pointer↔package consistency + repoint stale fixed_bet

jmoreira-valory:
Adds a CI guard so the trader can't silently ship a **stale betting strategy** again, and fixes a second stale pointer the guard caught.

## Background — the drift

The agent downloads its strategy at runtime from the CID in `file_hash_to_strategies` (a param), **not** from the packaged `customs` code. `autonomy packages lock` bumps the package CID but never that pointer, so any strategy PR that forgets to hand-edit the pointer silently keeps deploying the **old** strategy. This has bitten twice:

- **kelly** — pointer frozen at the March-25 build across `v0.36`→`v0.40` (~7 weeks); #971's depth fix and #999's `max_edge` never ran in production. Fixed in #1001.
- **fixed_bet** — pointer frozen at the March original (`bafybeigp65do…`) while the package moved on 2026-04-30 (`bafybeidahih5c7…`). Harmless (fallback disabled) but real.

## Changes

1. **`test_strategy_pointer_consistency.py`** — parametrized guard asserting `file_hash_to_strategies[strategy] == packages.json CID` for every strategy in `polymarket_trader`, `trader_pearl`, and the agent default. Fails CI with an actionable message on any drift; runs in the existing `test` job (no workflow changes).
2. **Repoint `fixed_bet`** to the packaged CID (`bafybeidahih5c7…`) + re-lock, so both strategies are consistent and the guard is green.

## Notes

- **Stacked on #1001** (which repoints kelly). CI is green because this branch includes #1001's kelly repoint; the diff here is only `fixed_bet` + the test. Once #1001 merges, this retargets to `main` and the diff collapses to exactly those two things.
- Verified: the test passes on the fixed state and **fails with a clear `STALE` message when a pointer is reverted**; `autonomy packages lock --check` passes; black/isort/flake8/darglint/copyright clean.

### DIFF
--- a/packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
+++ b/packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
@@ -22,12 +22,14 @@
 The agent downloads its betting strategy at runtime from the CID in the
 ``file_hash_to_strategies`` param, NOT from the packaged ``customs`` code.
 ``autonomy packages lock`` bumps the package CID but never that pointer, so a
-strategy change that forgets to repoint it silently ships the OLD strategy
-(see .claude/reports/strategy-pointer-drift.md for real incidents). This test
-fails when any ``file_hash_to_strategies`` entry disagrees with the
-``packages.json`` CID for that strategy.
+strategy change that forgets to repoint it silently ships the OLD strategy —
+this actually happened: the kelly pointer sat frozen for ~2 months across
+several releases while the packaged code moved on twice. This test fails when
+any ``file_hash_to_strategies`` entry disagrees with the ``packages.json`` CID
+for that strategy.
 """
 
+import functools
 import json
 import re
 from pathlib import Path
@@ -47,20 +49,22 @@ def _repo_root() -> Path:
 ROOT = _repo_root()
 
 # Every config that pins the runtime strategy pointer.
-POINTER_FILES = [
-    p
-    for p in (
-        ROOT / "packages/valory/services/polymarket_trader/service.yaml",
-        ROOT / "packages/valory/services/trader_pearl/service.yaml",
-        ROOT / "packages/valory/agents/trader/aea-config.yaml",
-    )
-    if p.is_file()
-]
+POINTER_FILES = (
+    ROOT / "packages/valory/services/polymarket_trader/service.yaml",
+    ROOT / "packages/valory/services/trader_pearl/service.yaml",
+    ROOT / "packages/valory/agents/trader/aea-config.yaml",
+)
 
 _POINTER_RE = re.compile(r"file_hash_to_strategies:\s*\$\{[^{]*(\{.*\})\}")
 _ENTRY_RE = re.compile(r'"(bafybei[a-z0-9]+)":\s*\[\s*"([a-z_]+)"\s*\]')
 
 
+def _config_id(path: Path) -> str:
+    """Return a disambiguating label, e.g. ``polymarket_trader/service.yaml``."""
+    return f"{path.parent.name}/{path.name}"
+
+
+@functools.lru_cache()
 def _package_cids() -> Dict[str, str]:
     """Return ``{strategy_name: package_cid}`` for every ``custom`` in packages.json."""
     data = json.loads((ROOT / "packages" / "packages.json").read_text())
@@ -77,6 +81,8 @@ def _package_cids() -> Dict[str, str]:
 
 def _pointer_entries(path: Path) -> List[Tuple[str, str]]:
     """Return ``(strategy_name, runtime_cid)`` pairs from a config's pointer."""
+    if not path.is_file():
+        return []
     match = _POINTER_RE.search(path.read_text())
     if match is None:
         return []
@@ -84,16 +90,26 @@ def _pointer_entries(path: Path) -> List[Tuple[str, str]]:
 
 
 _CASES = [
-    (path.name, name, cid)
+    (_config_id(path), name, cid)
     for path in POINTER_FILES
     for name, cid in _pointer_entries(path)
 ]
 
 
-def test_pointer_cases_discovered() -> None:
-    """Guard against a silently-broken parser reporting zero cases to check."""
-    assert POINTER_FILES, "No pointer config files found — paths broke."
-    assert _CASES, "No file_hash_to_strategies entries parsed — parser broke."
+def test_all_pointer_files_present() -> None:
+    """Every expected pointer config must exist (a rename must not silently drop coverage)."""
+    missing = [_config_id(p) for p in POINTER_FILES if not p.is_file()]
+    assert not missing, f"Expected strategy-pointer config(s) missing: {missing}"
+
+
+@pytest.mark.parametrize("path", POINTER_FILES, ids=_config_id)
+def test_each_pointer_file_has_entries(path: Path) -> None:
+    """Each pointer config must parse to at least one strategy entry."""
+    assert _pointer_entries(path), (
+        f"{_config_id(path)}: no file_hash_to_strategies entries parsed. A missing "
+        f"file or a format change would otherwise silently drop this config from the "
+        f"consistency check while the others keep the suite green."
+    )
 
 
 @pytest.mark.parametrize("config, strategy, runtime_cid", _CASES)
--- a/packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
+++ b/packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
@@ -1,0 +1,114 @@
+# -*- coding: utf-8 -*-
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2026 Valory AG
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""The runtime strategy pointer must match the packaged strategy code.
+
+The agent downloads its betting strategy at runtime from the CID in the
+``file_hash_to_strategies`` param, NOT from the packaged ``customs`` code.
+``autonomy packages lock`` bumps the package CID but never that pointer, so a
+strategy change that forgets to repoint it silently ships the OLD strategy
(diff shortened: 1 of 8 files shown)


Adamantios (dismissed):

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
```diff
@@ -0,0 +1,114 @@
...
+    (path.name, name, cid)
+    for path in POINTER_FILES
+    for name, cid in _pointer_entries(path)
+]
+
+
+def test_pointer_cases_discovered() -> None:
+    """Guard against a silently-broken parser reporting zero cases to check."""
+    assert POINTER_FILES, "No pointer config files found — paths broke."
+    assert _CASES, "No file_hash_to_strategies entries parsed — parser broke."
```
bennyjo: **The brittleness:** this guard pulls the strategy pointer out of each config file with a text-pattern (regex) search. That pattern matches the *exact shape* the pointer is written in today. If a future edit to one of these three files changes that shape even cosmetically — extra whitespace, the value split across lines, a slightly different `${...}` wrapper — the pattern stops matching and finds nothing in that file. And "found nothing" isn't treated as an error: the file just silently contributes zero cases. The one backstop here checks the *combined* total across all three files, so as long as even one file still parses, CI stays green — while the other one or two have quietly dropped out of the check entirely. For a guard whose whole job is to never pass silently on a stale pointer, that's the failure mode we most want to avoid.

Two ways to close it — your call which is worth it:

**1. Cheap fix — make a failed parse loud.** Keep the regex, but require *every* file to yield at least one entry, so a format drift fails that file visibly instead of vanishing:
```python
@pytest.mark.parametrize("path", POINTER_FILES, ids=lambda p: p.name)
def test_every_pointer_file_parses(path: Path) -> None:
    assert _pointer_entries(path), (
        f"{path.name}: file_hash_to_strategies did not parse — the regex has "
        f"drifted from the config format, so this file is silently unchecked."
    )
```

**2. Thorough fix — stop pattern-matching raw text.** Load the file with a YAML parser to get the `file_hash_to_strategies` value reliably (immune to whitespace/line-formatting), then extract the dict from it. More robust because YAML understands the file's structure rather than its exact spelling; slightly more code since you still have to peel the `${dict:...}` env-var wrapper off the value before reading the mapping.

The cheap fix is enough to remove the silent-pass hole; the thorough one removes the fragile text-matching altogether. Happy either way — flagging because the guard silently narrowing is exactly the class of problem this PR set out to catch.

bennyjo (commented):
One robustness note on the new pointer-consistency guard — inline.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
OjusWiZard: **Important — per-file coverage isn't guarded.** This backstop only checks the *global* `_CASES` list is non-empty. If any single one of the 3 `POINTER_FILES` stops parsing, that file is silently dropped while the other two keep the suite green — exactly the silent drift this test exists to prevent. Confirmed triggers: no `re.DOTALL` (a multi-line reflow of the dict), `_ENTRY_RE` matching only a single-name list (`Dict[str, List[str]]` legitimately allows `["kelly_criterion","kelly"]`), and the `[a-z_]+` name charset dropping a digit-bearing name like `kelly_criterion_v2` (drift is most likely exactly when a strategy is renamed/versioned).

Suggest asserting per-file: `for p in POINTER_FILES: assert _pointer_entries(p), f"{p}: parser matched zero entries"`, and/or cross-checking the discovered strategy set against `_package_cids().keys()`.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
```diff
@@ -0,0 +1,114 @@
...
+def _pointer_entries(path: Path) -> List[Tuple[str, str]]:
+    """Return ``(strategy_name, runtime_cid)`` pairs from a config's pointer."""
+    match = _POINTER_RE.search(path.read_text())
+    if match is None:
+        return []
+    return [(name, cid) for cid, name in _ENTRY_RE.findall(match.group(1))]
+
+
+_CASES = [
+    (path.name, name, cid)
```
OjusWiZard: **Important — ambiguous failure label.** `path.name` renders both `polymarket_trader/service.yaml` and `trader_pearl/service.yaml` as `"service.yaml"` (pytest only disambiguates with a `0`/`1` suffix). If `trader_pearl`'s pointer goes stale, the assertion message can't say which service — blunting the guard's diagnostic value. Suggest `str(path.relative_to(ROOT))` (or `f"{path.parent.name}/{path.name}"`) so it threads into the failure text.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
```diff
@@ -0,0 +1,114 @@
...
+#
+# ------------------------------------------------------------------------------
+
+"""The runtime strategy pointer must match the packaged strategy code.
+
+The agent downloads its betting strategy at runtime from the CID in the
+``file_hash_to_strategies`` param, NOT from the packaged ``customs`` code.
+``autonomy packages lock`` bumps the package CID but never that pointer, so a
+strategy change that forgets to repoint it silently ships the OLD strategy
+(see .claude/reports/strategy-pointer-drift.md for real incidents). This test
```
OjusWiZard: **Advisory — dangling reference.** `.claude/reports/strategy-pointer-drift.md` is git-ignored and absent from the repo, so this pointer is dead for anyone else reading the test. The PR description already documents the incidents well — consider inlining a one-line summary here or dropping the path.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
```diff
@@ -0,0 +1,114 @@
...
+    assert POINTER_FILES, "No pointer config files found — paths broke."
+    assert _CASES, "No file_hash_to_strategies entries parsed — parser broke."
+
+
+@pytest.mark.parametrize("config, strategy, runtime_cid", _CASES)
+def test_runtime_pointer_matches_packaged_strategy(
+    config: str, strategy: str, runtime_cid: str
+) -> None:
+    """Each file_hash_to_strategies CID must equal the packages.json CID."""
+    packaged = _package_cids().get(strategy)
```
OjusWiZard, line 150: **Advisory (nit).** `_package_cids()` re-reads and re-parses `packages.json` on every parametrized case (6×) on an invariant input. Negligible today; `@functools.lru_cache()` or computing once at module scope alongside `ROOT` would avoid the repeated work. No change required.

OjusWiZard (dismissed):
Approving — the change does what it claims: the `fixed_bet` repoint and all CID re-locks verify clean (`autonomy packages lock --check` passes), pointers match `packages.json`, and the new guard passes 7/7. The inline comments below are hardening suggestions on the guard itself (non-blocking); worth a follow-up since a regression guard's whole job is to fail loudly.

- **Important:** per-file parse coverage isn't guarded, and the two `service.yaml` files collide in failure labels.
- **Advisory:** dangling docstring reference; minor repeated re-parse.

Nice work catching the drift class and adding a meta-test for parser breakage.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
jmoreira-valory: Done in 265c4f493. Replaced the global backstop with a per-file guard: `test_each_pointer_file_has_entries` is parametrized over each pointer config and fails if that single config parses to zero entries, and `test_all_pointer_files_present` fails if one goes missing — so a file that stops parsing or is renamed fails loudly instead of being silently covered by the other two.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
jmoreira-valory: Done in 265c4f493. The label is now `<parent>/<name>` (`polymarket_trader/service.yaml` vs `trader_pearl/service.yaml` vs `trader/aea-config.yaml`), so a stale pointer names the exact config.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
jmoreira-valory: Done in 265c4f493. Dropped the git-ignored report path and inlined a one-line summary of the incident (kelly pointer frozen ~2 months) in the module docstring.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
jmoreira-valory: Done in 265c4f493. Added `@functools.lru_cache()` to `_package_cids()` so `packages.json` is read and parsed once instead of per parametrized case.

jmoreira-valory: Thanks — did both. Your cheap fix #1 (per-file `assert _pointer_entries(path)`) landed in 265c4f493 right after this comment, so you'd have seen only the old global-total backstop. And I've now done your thorough fix #2 in 92af5d29d: the parser reads each config with PyYAML and locates the `file_hash_to_strategies` value, then peels only the stable `dict:{...}` env-var payload and JSON-parses it — so whitespace, line-splitting, or cosmetic `${...}` reformatting no longer breaks matching. The per-file guard stays as a backstop, and it earned its keep immediately: it failed loudly that the agent default is `${dict:...}` while the services are `${VAR:dict:...}`, a wrapper difference my first pattern missed.

packages/valory/skills/decision_maker_abci/tests/test_strategy_pointer_consistency.py
jmoreira-valory: Done in 265c4f493 — added `@functools.lru_cache()` to `_package_cids()` so `packages.json` is read/parsed once instead of per parametrized case.

bennyjo (approved):
Re-review LGTM. Prior feedback all addressed: YAML parse instead of whole-file regex, per-file backstop guards (`test_each_pointer_file_has_entries` + `test_all_pointer_files_present`), and disambiguated `<parent>/<name>` failure labels. Verified the core invariant against `packages.json` on the head SHA — both runtime pointers equal their packaged `custom/valory/*` CIDs (kelly `bafybeidme2uybzyz…`, fixed_bet `bafybeidahih5c7…`), so the guard is genuinely green rather than green-because-it-parsed-nothing. No security/attack surface (CI-only test on trusted repo files; `safe_load_all` used). One non-blocking nit for a follow-up: `POINTER_FILES` is a hardcoded triple, so a future service with its own `file_hash_to_strategies` would escape the guard — worth a comment noting new configs must be added there.

DIvyaNautiyal07 (approved):

OjusWiZard (approved):

Adamantios (approved):

PR #1006, opened 2026-07-06
chore: bump mech-interact to v0.32.4

DIvyaNautiyal07:
## Summary
- Repins `mech_interact_abci` to the [v0.32.4](https://github.com/valory-xyz/mech-interact/releases/tag/v0.32.4) release hash (`bafybeici5tsg...`), propagated to the agent, services, and dependent skills via `autonomy packages lock`.
- v0.32.4 removes the dedicated `RESPONSE_ROUND_TIMEOUT` event (reverting to the shared `ROUND_TIMEOUT`), so `trader_abci/fsm_specification.yaml` is regenerated: `(MechResponseRound, ROUND_TIMEOUT) -> HandleFailedTxRound`.
- Moves the `upstream_pins` entry in `pyproject.toml` from `v0.32.4-rc2` to the final `v0.32.4` tag.
- Drops the explicit `use_dynamic_mech_selection: false` overrides from the flag-off configs (`trader_abci/skill.yaml`, `trader_pearl` and `polymarket_trader` service.yamls). `main` never sets this key, so the overrides added by the off-chain wiring silently flipped the effective value from the mech-interact default (`true`) to `false`. Removing them restores main parity.

Timeout wiring is unchanged: `MechInteractEvent.ROUND_TIMEOUT` still maps to `mech_interact_round_timeout_seconds` (5400s), which satisfies the new off-chain poll-budget guard (needs >= 330s).

## Test plan
- [x] `check-hash`, `check-packages`, `check-dependencies`, `check-third-party-hashes`, `uv lock --check`
- [x] `check-abciapp-specs` (regenerated FSM), `check-abci-docstrings`, `check-handlers`
- [x] black / isort / flake8 / mypy / pylint / darglint / bandit
- [x] `check-doc-hashes`, gitleaks
- [x] `py3.12-darwin`: 4173 tests pass


OjusWiZard (approved):
Verified end-to-end: hash propagation is consistent across all 10 files and matches upstream v0.32.4; the regenerated FSM edge `(MechResponseRound, ROUND_TIMEOUT) → HandleFailedTxRound` keeps the same target as before (only the event name changed) and `handle_failed_tx.py` already special-cases mech timeouts into the blacklisting path; no stale `RESPONSE_ROUND_TIMEOUT`/rc references remain at PR head.

On the `use_dynamic_mech_selection` removal — confirmed the PR's claim directly against `main`: `main` never sets this key in `trader_abci/skill.yaml` or either service.yaml, and `polymarket_trader` on `main` already ships the identical `valid_mechs` allowlist, so the effective `true` default is proven production behavior, not new exposure. For `trader_pearl` the flip is a no-op (`use_mech_marketplace: false`).

Two non-blocking suggestions:
1. Nothing in the test suite pins the resolved `mech_marketplace_config` for either service — a small resolution-level test (mirroring `test_agent_config_resolution.py`) asserting `use_dynamic_mech_selection` is absent/`true` would catch a future accidental re-introduction of the override in CI rather than via production drift.
2. Description nit: `MechResponseRound`'s effective timeout moves from the rc2-era ~330s to 5400s (restoring main behavior, since the dedicated event never accepted overrides in composed apps). Worth a line since "timeout wiring is unchanged" holds vs `main` but not vs the rc2 base.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

packages/valory/services/polymarket_trader/service.yaml
```diff
@@ -137,7 +137,7 @@ models:
...
+      mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x343F2B005cF6D70bA610CD9F1F1927049414B582","priority_mech_address":"0x1DCbC9368327776F8B20Dc041D40b8076C8AC173","priority_mech_staking_instance_address":"0x0000000000000000000000000000000000000000","priority_mech_service_id":25,"requester_staking_instance_address":"0x0000000000000000000000000000000000000000","response_timeout":300,"use_offchain":false,"offchain_url":null,"offchain_deposit_target_calls":10,"auto_deposit_cap_per_cycle":5000000,"offchain_poll_interval_seconds":3.0,"offchain_poll_timeout_seconds":300.0,"offchain_failover_max_retries":2}}
```
bennyjo, line 140: Since `MECH_MARKETPLACE_CONFIG` is wholesale env-overridable, any deployment pinning the old JSON (with `"use_dynamic_mech_selection": false`) keeps dynamic selection off after this merges — the YAML default only applies where the env var is unset. Is the plan to also clean up the Propel/Pearl env values, or is env-level divergence deliberate? Happy to leave as-is if so.

bennyjo (commented):
One deployment question — pin and FSM regen otherwise verified clean against the v0.32.4 tag.

packages/valory/services/polymarket_trader/service.yaml
DIvyaNautiyal07: Checked both deployment paths that pin `MECH_MARKETPLACE_CONFIG` for trader; no env cleanup needed and no production behavior changes.

**Pearl (olas-operate-middleware)** builds `MECH_MARKETPLACE_CONFIG` from scratch every deployment at `operate/services/manage.py:740-747`. The constructed JSON only carries `mech_marketplace_address`, `priority_mech_address`, `priority_mech_staking_instance_address`, `priority_mech_service_id`, `requester_staking_instance_address`, and `response_timeout` — it does not include `use_dynamic_mech_selection` at all. So the Python class default (`True`) wins for every Pearl-deployed trader today and has done all along, regardless of what the YAML defaults to.

**Propel (mech-deployments)** only ships mech-service env configs (`mech_mm_predict.env`, `single_mech_mm_predict*.env`, `nvm_native_mech_mm_predict_*.env`, etc.), not trader services. The word "trader" that turns up in a grep is inside `TOOLS_TO_PACKAGE_HASH` — tool-name text the mech serves — not a trader deployment.

Net effect of the merge:
- Production trader (`main` + Pearl): no change; the effective value was already `True` (main's YAML has never set the key, Pearl's env doesn't include it, Python default wins).
- Pearl deploying this branch: no change; same reasoning.
- Direct YAML load (dev/CI/non-Pearl): flips from `False` → `True`, which restores parity with `main`. This was the whole point of the removal — commit `471646e7a` introduced the `false` accidentally on this branch while claiming behavior was unchanged.

So the merge is env-safe. Happy to add a one-liner to the PR body noting we verified both paths if that helps future readers.

OjusWiZard (approved):
Approving. No new commits since review; the remaining open thread is @bennyjo's `MECH_MARKETPLACE_CONFIG` env-divergence question, which Divya's deployment-path analysis answers (Pearl constructs the config from scratch; no env cleanup needed) — leaving that thread to him to close.

bennyjo (approved):
Re-review: pin + FSM regen + `use_dynamic_mech_selection` override removal all verified clean against v0.32.4; full CI green at head. My earlier `MECH_MARKETPLACE_CONFIG` env-divergence question is answered (Pearl builds the config from scratch, no env cleanup needed). Approving.

PR #1007, opened 2026-07-07
feat(roi): include off-chain BalanceTracker pre-deposits in trader ROI cost

DIvyaNautiyal07:
## Summary

Adds the pre-deposit-as-loss term to trader's in-agent ROI calculation so pre-deposited off-chain mech funds are correctly reflected as cost.

Under the decision (DG's proposal, confirmed by David M), every top-up to a Safe's `mapRequesterBalances` in the mech marketplace's `BalanceTracker` is booked as spent the moment it lands on chain — the tracker has no requester-withdraw path, so committed funds are unrecoverable regardless of whether the mech has consumed them yet.

## What changes

- **`agent_performance_summary_abci` cost formula grows one additive term:**
  ```
  total_costs_usd = totalTradedSettled + totalFeesSettled
                  + settled_mech_costs_usd     # unchanged (existing on-chain path)
                  + offchain_prepaid_usd       # new
  ```
- **New reader package `valory/mech_prepaid_reader`** — a trader-local contract that reuses the same ABI as the vendored `valory/balance_tracker` and exposes one method (`get_deposit_events_for_requester`) that filters the `Deposit` topic by the requester address and sums the amounts. Kept trader-local so the vendored `balance_tracker` (pinned to `mech-interact`) stays untouched.
- **New `OffchainDepositState` on `agent_performance.json`** — persists `total_deposited_wei` and `last_scanned_block` under the new `offchain_deposits` field. Enables incremental log scans; each cycle reads only new blocks since the last checkpoint and adds to the monotonic cumulative sum.

## What doesn't change (existing-flow safety)

- `_get_total_mech_requests` still reads the marketplace subgraph counter — unchanged.
- `_get_open_market_requests` still filters last-4-days requests via `parsedRequest.questionTitle` — unchanged.
- `settled_mech_requests = total - open` — unchanged.
- `DEFAULT_MECH_FEE = 0.01` — unchanged.
- For any Safe that has never called `depositFor` (i.e. every production on-chain trader today), the event stream is empty and the new term is zero → the ROI number matches the pre-migration calculation exactly.

## Test plan

- [x] `tox -e check-packages`, `check-hash`, `check-third-party-hashes`, `check-handlers`
- [x] `tox -e check-abci-docstrings`, `check-abciapp-specs`
- [x] `tox -e black-check`, `isort-check`, `flake8`, `mypy`, `pylint`, `darglint`
- [x] `tox -e py3.12-darwin` — 4174 passed
- [x] New test `test_roi_reduces_with_offchain_prepaid` — asserts strictly-lower ROI when the pre-deposit term is non-zero, holding all other inputs constant.
- [x] Existing `test_successful_roi_gnosis` and `test_successful_roi_polymarket` patched to return 0 from the new helper — confirms the term collapses cleanly for on-chain-only agents.

## Notes for reviewers

- Base is `offchain-epic` so it lands alongside the rest of the off-chain wiring. The formula is additive and forward-compatible with `main` if we later cherry-pick.
- The reader is a new trader-local contract rather than a method on `valory/balance_tracker` because the latter is a third-party package pinned to `mech-interact`. If mech-interact ever gains an equivalent helper, we can drop the local package.


packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +364,162 @@ def _calculate_settled_mech_requests(
...
+            # On-chain-only deployment (default for all production traders).
+            # No contract call, no persist, ROI number is bit-identical to
+            # pre-migration.
+            return 0
+
+        summary = self.shared_state.read_existing_performance_summary()
+        state = summary.offchain_deposits or OffchainDepositState()
+        cached_total = state.total_deposited_wei
+
+        balance_tracker_address = marketplace_config.balance_tracker_address
```
OjusWiZard: **[Critical] Agent crashes at startup — `MechMarketplaceConfig` has no `balance_tracker_address` field.**

This PR adds `balance_tracker_address` to the `mech_marketplace_config` dict in `aea-config.yaml`, `trader_abci/skill.yaml`, and both service.yamls, but the vendored dataclass (`mech_interact_abci/models.py:229`) doesn't define that field (the `balance_tracker_address` at models.py:148 belongs to the unrelated `NVMConfig`). Params construction does `MechMarketplaceConfig(**kwargs["mech_marketplace_config"])` (mech_interact_abci/models.py:355), so boot dies. Reproduced with the exact `trader_pearl` dict:

```
TypeError: MechMarketplaceConfig.__init__() got an unexpected keyword argument 'balance_tracker_address'
```

Unit tests mock params, which is why 4174 tests pass anyway. Even if construction survived, this line would raise `AttributeError`.

Since `mech_interact_abci` is third-party/pinned, the fix is to carry the address in this skill's own params (`AgentPerformanceSummaryParams`) rather than inside `mech_marketplace_config` — or upstream the field to mech-interact first.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +364,162 @@ def _calculate_settled_mech_requests(
...
+
+        # ``latest_block`` never regresses below ``from_block``; the contract
+        # helper falls back to ``from_block`` when no new events landed.
+        total_wei = cached_total + new_wei
+        summary.offchain_deposits = OffchainDepositState(
+            total_deposited_wei=total_wei,
+            last_scanned_block=latest_block,
+        )
+        summary.timestamp = self.shared_state.synced_timestamp
+        self.shared_state.overwrite_performance_summary(summary)
```
OjusWiZard: **[Critical] This checkpoint write is clobbered later in the same cycle — the feature permanently contributes 0 to ROI.**

This helper runs from `calculate_roi()`, which is called inside `_fetch_agent_performance_summary()` (behaviours.py:1996). After it returns, that method builds a **fresh** `AgentPerformanceSummary(...)` that never sets `offchain_deposits` (behaviours.py:2044), and `_save_agent_performance_summary()` preserves `metrics` / `agent_details` / `agent_performance` / `profit_over_time` / `prediction_history` / `agent_behavior` — but has **no branch for `offchain_deposits`** — before `overwrite_performance_summary` dumps the whole file with no merge (models.py:384).

Net effect per cycle: seed/update checkpoint here → wiped to `None` at behaviours.py:2129 → next cycle re-enters the first-run seed path → returns 0, forever. On a real off-chain deployment the pre-deposit cost never accrues.

Tests miss it because `TestFetchOffchainPrepaidWei` mocks `shared_state` directly and `TestCalculateRoi` mocks this helper out — nothing exercises the full `async_act` persist round-trip.

Fix: carry `existing_data.offchain_deposits` forward in `_save_agent_performance_summary`, and add a round-trip test that runs the fetch + save path and asserts the checkpoint survives.

packages/valory/contracts/mech_prepaid_reader/tests/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Tests for the mech_prepaid_reader contract."""
```
OjusWiZard: **[Important] CI's 100% coverage gate fails as-is (98.24% total) — this package has zero real tests (42% coverage).**

No `test_contract.py` exists and nothing in the repo imports `MechPrepaidReaderContract` from a test. The untested code is exactly the risky part: the 32-byte topic padding of the requester address, the `eth_getLogs` filter construction, event decoding, and summation (`contract.py:88-117`) plus `get_current_block_number` (140-141). A padding/case bug there would silently match zero events forever rather than erroring.

Also untested elsewhere in the PR:
- all of `OffchainDepositState.__post_init__` validation/coercion (models.py:237-253) and the dict→dataclass branch (models.py:304) — `test_models.py` never references the new type;
- the seed-path non-STATE performative branch (behaviours.py:439-443);
- the `block_number <= 0` arm (only the missing-key arm is covered);
- the wei→USD conversion is asserted only directionally (`final_high < final_low`), never numerically, and only on the Gnosis/18dp side — the Polymarket/6dp path of the "load-bearing invariant" comment has no test.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +364,162 @@ def _calculate_settled_mech_requests(
...
+            ``account`` filter on the ``Deposit`` event.
+        :return: cumulative sum (in wei-scaled units) of every
+            ``BalanceTracker.Deposit`` credit to this Safe up to and
+            including the latest block scanned. Zero for on-chain-only
+            deployments, matching pre-migration ROI exactly.
+        :yield: contract-API response(s) via the abstract-round-abci
+            base behaviour helpers.
+        """
+        marketplace_config = self.params.mech_marketplace_config
+        if not marketplace_config.use_offchain:
```
OjusWiZard, line 448: **[Important] Flipping `use_offchain` back off silently erases already-accrued sunk cost from ROI.**

This early return fires before `summary.offchain_deposits` is ever read, so it returns `0` regardless of the persisted `total_deposited_wei`. Under the PR's own pre-deposit-as-loss rule those funds are permanently unrecoverable — a config rollback (or typo) after a period of off-chain operation inflates reported ROI with no log and no error. `test_use_offchain_false_short_circuit` only covers the `initial_state=None` case.

Suggest reading the cached total before branching and returning `cached_total` (not `0`) when scanning is disabled, so past committed deposits keep counting against ROI.

packages/valory/contracts/mech_prepaid_reader/contract.py
```diff
@@ -0,0 +1,141 @@
...
+                # ``account`` is the first indexed arg on
+                # ``Deposit(address indexed account, address indexed token, uint256 amount)``.
+                # 32-byte pad the requester address to match indexed-topic encoding.
+                "0x" + requester[2:].lower().rjust(64, "0"),
+            ],
+        }
+
+        logs = ledger_api.api.eth.get_logs(filter_params)
+        total_wei = 0
+        latest_block = from_block
```
OjusWiZard: **[Important] The scan range grows unboundedly during deposit-free stretches, then stalls permanently once the provider rejects it.**

When no events match, `latest_block` falls back to `from_block` (= previous checkpoint + 1), so the checkpoint advances by ~1 block per cycle while the chain head advances ~17k blocks/day on Gnosis. After a quiet stretch, `[from_block, "latest"]` exceeds the very `eth_getLogs` block-range caps this PR's own docstrings cite as the reason for seeding from head. Once the provider errors, the framework returns an ERROR performative, the caller keeps the cached total, the checkpoint never advances — and every future deposit goes uncounted forever, with only warning logs.

Fix: resolve the head block inside this callable (e.g. `to_block = ledger_api.api.eth.block_number`) and return that as the scanned watermark so the checkpoint advances to the scanned head rather than the last matched event — or scan in bounded chunks.

packages/valory/skills/agent_performance_summary_abci/models.py
```diff
@@ -198,6 +198,64 @@ def __post_init__(self) -> None:
...
+
+    def __post_init__(self) -> None:
+        """Validate the persisted checkpoint fields."""
+        # Legacy JSON may have shipped the wei count as a string when this
+        # dataclass was first introduced; coerce transparently and keep
+        # the invariant intact.
+        if isinstance(self.total_deposited_wei, str):
+            self.total_deposited_wei = int(self.total_deposited_wei)
+        if not isinstance(self.total_deposited_wei, int):
+            raise TypeError(
```
OjusWiZard, line 241: **[Important] These raises propagate uncaught out of the summary read path — one malformed persisted field now crashes all performance reporting.**

`read_existing_performance_summary` (models.py:374) only catches `(FileNotFoundError, json.JSONDecodeError)`, but `AgentPerformanceSummary.__post_init__` constructs `OffchainDepositState(**dict)` inside that `try`, so a hand-edited / schema-drifted / partially-migrated `offchain_deposits` entry (non-numeric string, stray extra key, negative value) raises `TypeError`/`ValueError` through every read site: `_should_update`, `_fetch_offchain_prepaid_wei`, `_update_profit_over_time_storage`, `_save_agent_performance_summary`. Previously no field could break the read; now one can.

Suggest catching `(TypeError, ValueError)` in the reader and degrading the single field (drop `offchain_deposits` with a loud warning) rather than failing the whole summary.

packages/valory/skills/agent_performance_summary_abci/models.py
```diff
@@ -215,6 +273,7 @@ class AgentPerformanceSummary:
     prediction_history: Optional[PredictionHistory] = None
     profit_over_time: Optional[ProfitOverTimeData] = None
     achievements: Optional[Achievements] = None
+    offchain_deposits: Optional[OffchainDepositState] = None
```
OjusWiZard, line 278: **[Important] Irreversible financial state is now stored via a non-atomic write with a blank-on-corrupt recovery.**

`overwrite_performance_summary` writes `agent_performance.json` in place (no temp-file + `os.replace`), and a `JSONDecodeError` on read silently returns a fresh `AgentPerformanceSummary()`. Pre-existing behavior — but this PR is the first to store *irreversible* state in it: a kill mid-write (OOM, `docker stop -t 0`, disk full) truncates the file, the next cycle reads `offchain_deposits=None`, re-seeds the checkpoint at the current head, and both the accumulated `total_deposited_wei` and every deposit in the gap are permanently lost, undercounting ROI cost with only a generic warning.

Suggest: atomic write (temp file + fsync + `os.replace`), and/or persisting this checkpoint in its own small file so corruption of the large volatile summary can't destroy it.

OjusWiZard (commented):
Multi-perspective review (correctness, silent failures, type design, tests, comments, simplification). 2 Critical, 6 Important, 2 Advisory findings — posted inline. The two Criticals are independently fatal: one crashes the agent at startup with the shipped config, the other makes the new ROI term permanently zero even once boot is fixed. Both were verified by direct execution/tracing, not just inspection.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +364,162 @@ def _calculate_settled_mech_requests(
...
+            return 0
+
+        summary = self.shared_state.read_existing_performance_summary()
+        state = summary.offchain_deposits or OffchainDepositState()
+        cached_total = state.total_deposited_wei
+
+        balance_tracker_address = marketplace_config.balance_tracker_address
+        if (
+            not balance_tracker_address
+            or balance_tracker_address.lower() == _ZERO_ADDRESS
```
OjusWiZard, line 457: **[Important] The one genuine misconfiguration branch is the only silent one.**

Every other bail-out in this function logs a warning, but `use_offchain=True` with a missing/zero `balance_tracker_address` — the case where the operator *intended* off-chain accounting and the deployment is broken — returns with no log at all. The ROI cost silently freezes at `cached_total` (0 on first boot) forever, indistinguishable from a healthy zero-spend agent. Suggest a `logger.warning` with a distinct marker, e.g. `[Offchain Deposits] misconfigured: use_offchain=True but balance_tracker_address=...`.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +364,162 @@ def _calculate_settled_mech_requests(
...
+            if block_response.performative != ContractApiMessage.Performative.STATE:
+                self.context.logger.warning(
+                    "Failed to seed offchain-deposit checkpoint from current "
+                    f"block (response: {block_response}); leaving cached total."
+                )
+                return cached_total
+
+            body = block_response.state.body
+            raw_block = body.get("block_number") if body is not None else None
+            if raw_block is None or int(raw_block) <= 0:
```
OjusWiZard: **[Advisory] `<= 0` rejects a legitimate `block_number == 0`, contradicting the `OffchainDepositState` docstring.**

On a fresh devnet/Hardhat chain whose head is genesis, this treats the real head block `0` as malformed and loops the seed path (with a warning) every cycle — while models.py:215 explicitly says the `Optional` design exists to support "a legitimate genesis-block checkpoint... on devnet/Hardhat chains where block 0 is a real block." Mainnet impact is nil. Either relax to `< 0` or fix the models.py docstring.

packages/valory/contracts/mech_prepaid_reader/contract.py
```diff
@@ -0,0 +1,141 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Trader-local reader for BalanceTracker ``Deposit`` events.
+
+The vendored ``valory/balance_tracker`` contract package is third-party
+(pinned to ``mech-interact``) and therefore not modified in trader. This
+package sits alongside it with the same ABI and adds one helper
```
OjusWiZard: **[Advisory] Docstring undercounts the public API.**

"adds one helper" — the class defines two classmethods; `get_current_block_number` is load-bearing for the first-run checkpoint seeding and worth mentioning here. Relatedly, the "Legacy JSON may have shipped the wei count as a string" comment in models.py:235 describes history that never shipped: both the field and its coercion were introduced within this unmerged PR, so that coercion path currently has no real writer.

packages/valory/skills/trader_abci/skill.yaml
```diff
@@ -300,6 +300,7 @@ models:
         priority_mech_service_id: 0
         requester_staking_instance_address: '0x0000000000000000000000000000000000000000'
         response_timeout: 300
+        balance_tracker_address: '0x0000000000000000000000000000000000000000'
```
bennyjo: Blocking: the pinned `mech_interact_abci` package's `MechMarketplaceConfig` is a frozen dataclass with no `balance_tracker_address` field, so adding this key to the config dict (here and in aea-config + both service.yamls) makes `MechMarketplaceConfig(**...)` raise `TypeError` during param init — the agent crashes at startup before any behaviour runs. The PR body's "Python default in mech_interact_abci/models.py" is `NVMConfig`, a different class; either bump the mech-interact pin to a version whose config carries this field, or move the address to a trader-owned param.

bennyjo (commented):
Second-reviewer pass — two blocking findings (config incompatible with the pinned mech-interact package; persisted checkpoint wiped every cycle), one address-verification question, plus resilience/brittleness follow-ups.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
bennyjo: Blocking: this persisted state is wiped later in the same cycle — `_fetch_agent_performance_summary` builds a fresh summary without `offchain_deposits` and `_save_agent_performance_summary` overwrites the whole file with it, so every cycle re-enters the first-run seed path and the new cost term is permanently 0. Suggest preserving `offchain_deposits` from `existing_data` in `_save_agent_performance_summary` (like `agent_behavior`), plus a two-cycle test through `async_act` — the helper tests can't see this because they stop before the save.

packages/valory/services/polymarket_trader/service.yaml
```diff
@@ -137,7 +137,7 @@ models:
...
+      mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x343F2B005cF6D70bA610CD9F1F1927049414B582","priority_mech_address":"0x1DCbC9368327776F8B20Dc041D40b8076C8AC173","priority_mech_staking_instance_address":"0x0000000000000000000000000000000000000000","priority_mech_service_id":25,"requester_staking_instance_address":"0x0000000000000000000000000000000000000000","response_timeout":300,"balance_tracker_address":"0xaaFBeef195BDAb1Bb6f3Dc9cEbA875Cd72499230","use_offchain":false,"offchain_url":null,"offchain_deposit_target_calls":10,"auto_deposit_cap_per_cycle":5000000,"offchain_poll_interval_seconds":3.0,"offchain_poll_timeout_seconds":300.0,"offchain_failover_max_retries":2}}
```
bennyjo: `0xaaFBeef…` is the **Base** entry in the vendored `CHAIN_TO_NVM_CONFIG` (there's no Polygon entry at all), and the bytecode at this address on Polygon differs from Base's — so with `mech_chain_id: polygon` the reader would scan an unrelated contract. Can you confirm the intended BalanceTracker address on the chain the polymarket trader's mechs actually use?

packages/valory/contracts/mech_prepaid_reader/contract.py
bennyjo: Falling back to `from_block` when no logs match means the checkpoint advances one block per cycle during deposit-quiet periods while the head advances thousands, so the scan window grows until it hits provider `eth_getLogs` range caps — the same failure commit 2 fixed for first-run — after which the term freezes and later deposits are silently missed. Suggest resolving `to_block="latest"` to a concrete head block and returning that as `latest_block`, and covering the fallback with a contract test (this package currently has none).

packages/valory/skills/agent_performance_summary_abci/models.py
```diff
@@ -198,6 +198,64 @@ def __post_init__(self) -> None:
...
+
+    Enforced invariants at construction time:
+    - ``total_deposited_wei`` must be a non-negative integer.
+    - ``last_scanned_block`` must be ``None`` or a non-negative integer.
+    """
+
+    total_deposited_wei: int = 0
+    last_scanned_block: Optional[int] = None
+
+    def __post_init__(self) -> None:
```
bennyjo, line 232: `__post_init__` raising `TypeError`/`ValueError` isn't caught by `read_existing_performance_summary` (it only catches `FileNotFoundError`/`JSONDecodeError`), so one bad persisted value bricks every summary read across all behaviours with no self-heal. Suggest degrading to a fresh `OffchainDepositState` with a warning instead of raising.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
bennyjo, line 448: Question: if an operator disables `use_offchain` after real deposits were counted, this returns 0 instead of the cached total, so historical committed spend silently drops out of the cost side and ROI improves. Happy to leave as-is if that's deliberate — flagging since the pre-deposit-as-loss rule reads as irrevocable.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fixed as of 032d879d8. Pulled `balance_tracker_address` off `MechMarketplaceConfig` entirely and moved it onto `AgentPerformanceSummaryParams` where it's actually read. `mech_interact_abci` is back to upstream with no fork.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Real bug, fixed in a3bda100b. `_save_agent_performance_summary` now carries `offchain_deposits` forward from `existing_data` alongside `agent_behavior`. Added `test_preserves_offchain_deposits_from_existing` in `TestSaveAgentPerformanceSummary` as the round-trip regression test that would have caught this.

packages/valory/contracts/mech_prepaid_reader/tests/__init__.py
DIvyaNautiyal07: The `mech_prepaid_reader` package is gone as of 032d879d8, consolidated into the vendored `valory/balance_tracker` package (moved from third_party to dev). Added 4 contract-level tests in `balance_tracker/tests/test_contract.py` covering topic padding, filter range, empty match, and log decoding. `OffchainDepositState` validation is covered too, and the read/save round-trip test now locks in the persistence fix.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fixed in a3bda100b. Moved the `cached_total` read before the `use_offchain` gate and now return `cached_total` from every disabled/misconfigured bail. Once counted, always counted. Zero-spend agents still see 0 because `cached_total` defaults to 0. Added `test_use_offchain_false_preserves_accrued_cached_total`.

packages/valory/contracts/mech_prepaid_reader/contract.py
DIvyaNautiyal07: Fixed in a3bda100b. The behaviour resolves the current head block up front via `get_ledger_api_response("get_block_number")`, passes that as the concrete `to_block`, and advances the checkpoint to the scanned head every successful cycle regardless of matches. No more unbounded range growth during quiet periods. Added `test_incremental_empty_entries_advances_checkpoint_to_head` and `test_incremental_from_block_beyond_head_returns_cached_no_scan`.

packages/valory/skills/agent_performance_summary_abci/models.py
DIvyaNautiyal07: Fixed in a3bda100b. `read_existing_performance_summary` now catches `(TypeError, ValueError)` and degrades to a fresh summary with a loud warning, same shape as the existing FileNotFoundError/JSONDecodeError branch. Trades field-level degradation for a simpler catch, so a corrupt `offchain_deposits` field drops the whole summary rather than just that field. If we ever hit real corruption there, splitting the checkpoint into its own file is the next step.

packages/valory/skills/agent_performance_summary_abci/models.py
DIvyaNautiyal07: Fixed in a3bda100b. `overwrite_performance_summary` now writes to a same-directory tempfile, `fsync`s, and does `os.replace`, so a mid-write kill can't leave the file truncated. Best-effort tempfile cleanup on failure. Splitting the checkpoint into a separate small file is the follow-up if we hit a real corruption case in the wild.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fixed in a3bda100b. Emits `[Offchain Deposits] misconfigured: use_offchain=True but balance_tracker_address=...` at WARNING and returns cached total. Every other bail already logs; this one now matches. Covered by `test_zero_address_returns_cached_total_and_warns`.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fixed in a3bda100b. Changed to `< 0`. Genesis block 0 is now accepted as a real head, consistent with the `OffchainDepositState` docstring. Added `test_seed_accepts_genesis_block_zero`.

packages/valory/contracts/mech_prepaid_reader/contract.py
DIvyaNautiyal07: `mech_prepaid_reader` is gone as of 032d879d8, so the "one helper" docstring vanished with the file. The "Legacy JSON" comment on `OffchainDepositState.__post_init__` was rewritten in a3bda100b to describe the coercion as defensive rather than historical, since you're right that no prior JSON ever shipped that shape.

packages/valory/skills/trader_abci/skill.yaml
DIvyaNautiyal07: Fixed in 032d879d8. Moved `balance_tracker_address` off `MechMarketplaceConfig` and onto this skill's own params (`AgentPerformanceSummaryParams`). `mech_interact_abci` reverted to upstream, no fork.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Same bug as Ojus flagged, fixed in a3bda100b. `_save_agent_performance_summary` now carries `offchain_deposits` from `existing_data`. Added a round-trip regression test in `TestSaveAgentPerformanceSummary`.

packages/valory/services/polymarket_trader/service.yaml
DIvyaNautiyal07: Nice catch. That address is exactly `CHAIN_TO_NVM_CONFIG[ChainType.BASE].balance_tracker_address` in mech_interact_abci, a wrong-line copy on my part. Zeroed the Polygon default in a3bda100b so the scan is disabled until an operator supplies the real Polygon fixed-price BalanceTracker via `BALANCE_TRACKER_ADDRESS` env override. Do you have a verified BalanceTracker address for the Polygon marketplace that ships with the polymarket trader?

packages/valory/contracts/mech_prepaid_reader/contract.py
DIvyaNautiyal07: Fixed in a3bda100b, same commit as Ojus's version of this. Head block resolved concretely up front, checkpoint advances to head every cycle. Contract-level tests for the reader now live in `balance_tracker/tests/test_contract.py` (topic padding, filter range, empty match, decoding).

packages/valory/skills/agent_performance_summary_abci/models.py
DIvyaNautiyal07: Fixed in a3bda100b. `read_existing_performance_summary` catches `(TypeError, ValueError)` and degrades to a fresh summary with a warning.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Went with the "irrevocable" reading, fixed in a3bda100b. Cached total is read before the `use_offchain` gate now and returned on any disabled or misconfigured bail, so historical committed spend keeps counting against ROI. Zero-spend agents still see 0 because `cached_total` defaults to 0.

DIvyaNautiyal07:
Landed the CI unblock in 07413286f. Also worth flagging why the fix took this shape, since the class of problem will bite anyone else who edits a `third_party` contract in place.

**Root cause: third_party ripple.**

`valory/balance_tracker` is declared as a dependency by the `third_party` `valory/mech_interact_abci` skill (in mech_interact_abci's `skill.yaml` `contracts:` list, pinned by CID). When commit 032d879d8 moved `balance_tracker` to `dev` and added the `get_deposit_events_for_requester` classmethod, `autonomy packages lock` rewrote `mech_interact_abci`'s own `skill.yaml` to point at the trader-side balance_tracker CID. That re-hashed `mech_interact_abci` itself to a locally-computed CID that had never been published to IPFS. CI's `autonomy packages sync` tried to fetch that CID from the registry and hung until the 2-minute read timeout on every run.

Symptom in the logs was `Error: The read operation timed out` on the line right after `Adding (skill, valory/mech_interact_abci:0.1.0)`. Easy to misread as a flaky IPFS gateway, which is what my earlier commit messages assumed. Two consecutive re-runs failing the same way was the tell.

**Why the rename over the alternatives:**

- Reverting the balance_tracker refactor entirely would drop the ROI reader work from this PR.
- Publishing the local-drift `mech_interact_abci` CID to the registry would ship non-upstream content under `valory/mech_interact_abci`, which is not ours to publish.
- Upstreaming into `valory-xyz/mech` first is the right long-term move but blocks this PR on an unrelated review cycle.

The rename keeps the work landed, isolates the trader-side edit into a distinct package that doesn't collide with any third_party skill's dependency graph, and reduces to a proper upstream port later without additional refactor.

**What actually changed:**

- New `packages/valory/contracts/mech_prepaid_reader/` (dev) — carries only the `get_deposit_events_for_requester` classmethod plus a minimal Deposit-event ABI fragment. No bytecode, no other BalanceTracker methods.
- `packages/valory/contracts/balance_tracker/` moved back to `third_party` at `bafybeiayyxl643qfpw57vopytxgyd6flxxmwassswuxen7myekqaxdxxvi` — the CID `mech_interact_abci@v0.32.4` declares as its `balance_tracker` dependency. Untracked from git and added to `.gitignore` to match the third_party convention.
- Consumer swap in `agent_performance_summary_abci/behaviours.py`:
  ```python
  # before
  from packages.valory.contracts.balance_tracker.contract import BalanceTrackerContract
  # after
  from packages.valory.contracts.mech_prepaid_reader.contract import MechPrepaidReaderContract
  ```
  Same swap for the `contract_id=` call site. `skill.yaml` `contracts:` list and `aea-config.yaml` follow the rename.
- `mech_interact_abci` locks back at its upstream `bafybeici5tsg...` hash so CI can fetch it fresh from IPFS.

Zero on-chain behaviour change — the reader still filters `BalanceTracker.Deposit(account, token, amount)` events by the requester Safe address, same eth_getLogs shape, same aggregation in the caller.

**Follow-up:** port `get_deposit_events_for_requester` upstream into `valory-xyz/mech`'s `balance_tracker`. When a mech release with it lands, bump `upstream_pins`, drop `mech_prepaid_reader`, and switch the caller back to the `third_party` `balance_tracker`.

packages/valory/contracts/mech_prepaid_reader/tests/__init__.py
OjusWiZard: Re-checked at 324fd78: `mech_prepaid_reader/contract.py` is now at **100%** (nice test suite for the topic padding + decoding) — but the gate still fails: total **98.61%**, with `models.py` at 92%. Still uncovered new code:

- `OffchainDepositState.__post_init__` raises (models.py 239, 241, 245, 250, 255) and the dict→dataclass branch (306) — still no direct tests;
- the new `(TypeError, ValueError)` degradation path in `read_existing_performance_summary` (395–409);
- the atomic-write exception cleanup (434–440);
- the head-block non-STATE branch in `behaviours.py` (458–462);
- `test_roi_reduces_with_offchain_prepaid` still asserts direction only, and only on the Gnosis/18dp side — no numeric assertion and no Polymarket/6dp variant.

Keeping this thread open until the coverage job goes green.

OjusWiZard: CI coverage job is green at `5dee223f8` (full-suite run under `source = packages/valory` — my earlier package-scoped local runs undercounted relative to CI's aggregate). Gate condition met — resolving. Two behavioral suggestions from this thread remain optional follow-ups rather than blockers: a numeric (not just directional) ROI assertion for the wei→USD conversion, and a Polymarket/6dp variant of the offchain-prepaid ROI test.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -1866,6 +2091,14 @@ def _save_agent_performance_summary(
         # Always preserve agent_behavior from existing data
         agent_performance_summary.agent_behavior = existing_data.agent_behavior
 
+        # Always preserve offchain_deposits. ``_fetch_offchain_prepaid_wei``
+        # writes this state directly via ``overwrite_performance_summary``
+        # earlier in the cycle, but ``_fetch_agent_performance_summary``
+        # rebuilds this summary without the field (defaults to ``None``).
+        # Without this line the cycle-end write clobbers the just-persisted
+        # checkpoint every cycle and the ROI cost term stays permanently 0.
+        agent_performance_summary.offchain_deposits = existing_data.offchain_deposits
```
OjusWiZard: **[High] Corruption in any *unrelated* field can wipe the just-written checkpoint through this line.**

`read_existing_performance_summary` degrades to a **fresh** `AgentPerformanceSummary()` on a `TypeError`/`ValueError` from *any* nested dataclass's `__post_init__` (models.py:395-409) — not just `offchain_deposits`. If, at this cycle-end save, e.g. a persisted `Achievements` or `PredictionHistory` entry no longer round-trips, `existing_data.offchain_deposits` comes back `None` and this line unconditionally assigns it — overwriting the correct checkpoint `_fetch_offchain_prepaid_wei` persisted moments earlier in the same cycle. Since `total_deposited_wei` is documented as irreversible and non-re-derivable, corruption in a field that has nothing to do with off-chain deposits permanently destroys the ROI cost state, with only a generic "starting from a fresh summary" warning.

`test_preserves_offchain_deposits_from_existing` doesn't catch this — it mocks the read to always return good data.

Fix: only preserve when `existing_data.offchain_deposits is not None`, or (better, and already floated by the comment in `overwrite_performance_summary`) split the checkpoint into its own small file so nothing else's corruption can cascade into it.

packages/valory/services/polymarket_trader/service.yaml
```diff
@@ -160,6 +160,7 @@ models:
       use_x402: ${USE_X402:bool:false}
       is_agent_performance_summary_enabled: ${IS_AGENT_PERFORMANCE_SUMMARY_ENABLED:bool:true}
       is_achievement_checker_enabled: ${IS_ACHIEVEMENT_CHECKER_ENABLED:bool:true}
+      balance_tracker_address: ${BALANCE_TRACKER_ADDRESS:str:0x0000000000000000000000000000000000000000}
```
OjusWiZard: **[Medium] The config split regressed Polystrat's BalanceTracker default to the zero address.**

Before the split, this service's `mech_marketplace_config.balance_tracker_address` carried `0xaaFBeef195BDAb1Bb6f3Dc9cEbA875Cd72499230`; `trader_pearl/service.yaml:161` kept its real value (`0x7D68…D67c`) through the migration, but this one dropped to the generic `0x000…0` placeholder. Latent while `use_offchain: false`, but the moment Polystrat flips it on — the point of this PR — the scan hits the misconfiguration guard and permanently freezes the off-chain cost term at `cached_total` (a warning log is the only signal), unless the operator separately remembers to set `BALANCE_TRACKER_ADDRESS`.

Note this intersects the open question on this file about whether `0xaaFBeef…` (the **Base** entry in the vendored `CHAIN_TO_NVM_CONFIG`) was ever the right address for a Polygon deployment — so the fix here is deciding the correct Polygon BalanceTracker address and defaulting to it, not just restoring the old value. Longer term, deriving this param from the same chain-keyed table `mech_interact_abci` uses would make drift impossible.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +365,192 @@ def _calculate_settled_mech_requests(
...
+            self.context.logger.warning(
+                "Offchain-deposit scan returned malformed response "
+                f"(entries={entries!r}); leaving checkpoint and cached total unchanged."
+            )
+            return cached_total
+
+        # Aggregation lives here (not in the classmethod) by convention:
+        # contract classmethods return processed log entries; sum happens
+        # in the caller.
+        new_wei = sum(int(entry.get("args", {}).get("amount", 0)) for entry in entries)
```
OjusWiZard: **[Medium] Silent zero-fill on malformed entries is inconsistent with this function's own refuse-to-advance policy.**

The `entries is None` gate above correctly refuses to advance the checkpoint on a malformed response — but past that gate, an entry missing `args`/`amount` silently contributes `0` while the checkpoint still advances to `head_block`, so its block range is never rescanned: a permanent, irreversible under-count with no log. The pre-rework contract-side code used hard indexing (`entry["args"]["amount"]`), which would raise loudly and land in the existing ERROR-performative path.

Likelihood is admittedly low (ABI-decoded, topic-filtered `Deposit` events can't really lack these keys once `process_log` succeeds) — but since the failure direction is silently flattering ROI under a policy this PR added extensive tests to protect, the cheap fix is reverting to hard indexing so a shape drift fails loudly instead of zero-filling. Either way, there's currently no test pinning whichever behavior is intended.

packages/valory/contracts/mech_prepaid_reader/contract.py
```diff
@@ -0,0 +1,100 @@
...
+            "fromBlock": from_block,
+            "toBlock": to_block,
+            "address": contract_instance.address,
+            "topics": [
+                event.topic,
+                "0x" + requester[2:].lower().rjust(64, "0"),
+            ],
+        }
+        logs = ledger_api.api.eth.get_logs(filter_params)
+        return {"entries": [event.process_log(log) for log in logs]}
```
OjusWiZard: **[Advisory] Raw `process_log` output violates the `JSONLike` contract and the repo's own convention — latent, not live.**

Each entry is a web3 `AttributeDict` with nested `HexBytes` (`transactionHash`, `blockHash`). Building the real `ContractApiMessage(STATE, ...)` and calling `.encode()` reproduces `NotImplementedError: DictProtobufStructSerializer doesn't support dict value type <class 'web3.datastructures.AttributeDict'>`. It doesn't crash today only because the ledger connection is in-process and never serializes the message (verified through `connections/ledger/base.py` + `contract_dispatcher.py`), and the caller only pulls `amount`. But `mech_marketplace/contract.py:194-203` flattens to plain dicts (`"0x" + entry["transactionHash"].hex()`, unpacked `args`, plain ints) for exactly this reason — anyone who later routes this response across a real wire boundary or into a consensus payload gets a runtime error. Cheap to fix now: return plain-dict entries (or just `{"amounts": [...], "block_numbers": [...]}` since that's all the caller uses).

packages/valory/contracts/mech_prepaid_reader/contract.py
```diff
@@ -0,0 +1,100 @@
...
+#
+# ------------------------------------------------------------------------------
+
+"""Trader-local reader for BalanceTracker ``Deposit`` events.
+
+Exists as a distinct package so trader's pre-deposit-as-loss ROI
+accounting can attach an ``eth_getLogs`` classmethod without editing
+the third_party ``valory/balance_tracker`` contract. Editing that
+package in place would ripple into ``mech_interact_abci``'s dependency
+list and force a locally-computed hash for a third_party skill (the
```
OjusWiZard: **[Advisory] Four small inaccuracies in the rewritten docs.**

1. Here: `valory/balance_tracker` is a **contract** package, not a "third_party skill" — it's pinned via `mech_interact_abci/skill.yaml`'s `contracts:` list and lives under `contract/valory/balance_tracker` in packages.json.
2. Lines 30-32: the port-upstream path names `valory-xyz/mech`, but the repo's actual sync source for this vendored package is `valory-xyz/mech-interact` (`Makefile`'s `autonomy packages sync --source valory-xyz/mech-interact:$(MECH_INTERACT_VERSION)`); there's no fetch path wired to `valory-xyz/mech`.
3. Lines 71-74 (`get_deposit_events_for_requester` docstring): "Aggregation ... lives in the caller by convention (see `mech_marketplace.contract`)" — the cited file does the opposite: `get_marketplace_undelivered_reqs` aggregates/reconciles *inside* the contract classmethod. The citation should go (the split is still fine, just not a convention that file establishes).
4. `tests/test_contract.py:114`: "right-padded to 32 bytes" — `rjust` left-pads, and both `contract.py:79` and the neighboring `_make_deposit_log` docstring say "left-padded" correctly.

packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py
```diff
@@ -2187,11 +2194,530 @@ def test_successful_roi_polymarket(self) -> None:
...
+
+
+def _error_response() -> MagicMock:
+    """Build a non-STATE contract-API response (framework wrapping an error)."""
+    response = MagicMock()
+    response.performative = ContractApiMessage.Performative.ERROR
+    return response
+
+
+def _ledger_state_response(body: Any) -> MagicMock:
```
OjusWiZard: **[Advisory] Duplicate mock-builder; and `test_contract.py` hand-rolls what siblings mock.**

`_ledger_state_response` is a byte-for-byte copy of `_state_response` above it, differing only in which `Performative` gets stamped — a `performative` keyword on `_state_response` (defaulting to the contract-API one) would fold them into one canonical builder, consistent with the file's other shared helpers.

Relatedly, `mech_prepaid_reader/tests/test_contract.py` builds real keccak-hashed, ABI-encoded `Deposit` logs by hand (~60 lines incl. a defensive `startswith("0x")` branch that exists only because of the manual encoding), where sibling contract tests (`mech/tests/test_contract.py`, `mech_marketplace_legacy/tests/test_contract.py`) mock the event object directly. Counterpoint in fairness: the real-decode approach does buy genuine ABI-decode coverage that the mock style wouldn't. Judgment call — flagging for consistency, not correctness.

OjusWiZard (commented):
Round-2 review of the rework (head 5dee223f8). The round-1 fixes check out and CI is green — these are new findings introduced by the rework itself: 1 High, 2 Medium, 3 Advisory, posted inline. The framework-boundary risks (ledger `get_block_number` key/`chain_id` handling, `AttributeDict` crossing the in-process connection) were traced through the installed AEA/web3 stack and are confirmed safe as-is; the inline comments cover what isn't.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: You're right, fixed in 628d8a2a3. Guarded the preserve with `if existing_data.offchain_deposits is not None` so a validation-degraded fresh read (from any unrelated `__post_init__` raise on Achievements, PredictionHistory, etc.) can't wipe the checkpoint `_fetch_offchain_prepaid_wei` wrote earlier the same cycle. Added `test_does_not_wipe_offchain_deposits_when_existing_read_degraded` — feeds a fresh existing summary and a new_summary carrying a real checkpoint, asserts the checkpoint survives the save. Splitting the checkpoint into its own file is still the better long-term shape, keeping that as follow-up.

packages/valory/contracts/mech_prepaid_reader/contract.py
DIvyaNautiyal07: Confirmed the latent risk, fixed in 628d8a2a3. The classmethod now folds each log to `{"amount": int, "block_number": int}` inside the classmethod, so the return is `{"entries": [{"amount": int, "block_number": int}, ...]}` — plain dicts, plain ints, no `AttributeDict` or `HexBytes` in the response body. Behaviour aggregation switches to hard-index (`int(entry["amount"])`) and the test_contract.py decoded-entries assertion updates to the flat shape. Also see the paired reply on 3538953279 for the aggregation-side fix.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fair point on the inconsistency, fixed in 628d8a2a3. Behaviour now uses `int(entry["amount"])` hard indexing against the flat entry shape (see paired classmethod change in 3538953287). A shape drift raises loudly and lands in the existing ERROR-performative arm above rather than silently zero-filling while advancing the checkpoint to head. Test coverage for the new shape lives in `test_incremental_scan_accumulates_and_advances_checkpoint_to_head` (updated for the new mock entries).

packages/valory/services/polymarket_trader/service.yaml
DIvyaNautiyal07: Investigated in the marketplace repo. From `scripts/deployment/globals_polygon_mainnet.json` on Polygon we have `balanceTrackerFixedPriceTokenUSDCAddress = 0x5C50ebc17d002A4484585C8fbf62f51953493c0B`, and Polymarket trader pays for mech requests in USDC (via pUSD wrapping), so that's the right default. Set in 628d8a2a3. While I was there I also noticed `trader_pearl` was defaulting to `0x7D68…D67c` on Gnosis, which per the same globals file is the NVM subscription tracker (`balanceTrackerNvmSubscriptionNativeAddress`) rather than the fixed-price native one — same wrong-line shape as Polygon had. Corrected trader_pearl to `0x21cE6799A22A3Da84B7c44a814a9c79ab1d2A50D` (`balanceTrackerFixedPriceNativeAddress` — trader pays in wxDAI on Gnosis) in the same commit. Deriving from the chain-keyed table longer-term is a good follow-up, agreed.

packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py
DIvyaNautiyal07: Consolidated in 628d8a2a3. `_ledger_state_response` folded into `_state_response(body, performative=ContractApiMessage.Performative.STATE)` — the ledger-API sites pass `performative=LedgerApiMessage.Performative.STATE` at the call site. On `mech_prepaid_reader/tests/test_contract.py` I'm leaving the real-decode style as-is since (as you noted) it does buy genuine ABI-decode coverage that the mocked style wouldn't; the extra ~60 lines vs the sibling style is the price for that coverage.

packages/valory/contracts/mech_prepaid_reader/contract.py
DIvyaNautiyal07: Four fair callouts, all fixed in 628d8a2a3.

1. The module docstring on `mech_prepaid_reader/contract.py:20-33` now names `mech_interact_abci` explicitly and calls it correctly — `mech_interact_abci` is the third_party skill whose `contracts:` list would need to change; `valory/balance_tracker` is the contract package it depends on. The two are distinguished in the new phrasing.
2. Upstream path now points at `valory-xyz/mech-interact`, matching `Makefile:135`'s `autonomy packages sync --source valory-xyz/mech-interact:$(MECH_INTERACT_VERSION)` — that's the actual sync source for this vendored contract package.
3. Dropped the `mech_marketplace.contract` convention citation — you're right, `get_marketplace_undelivered_reqs` aggregates inside the classmethod, so it doesn't establish the split-the-aggregation convention. New docstring just states the caller-side aggregation as an implementation choice for the pre-deposit-as-loss reader specifically.
4. `tests/test_contract.py:114` corrected: `right-padded` → `left-padded`, matching line 63 in the same file and `contract.py`'s topic-encoding line.

packages/valory/contracts/mech_prepaid_reader/tests/__init__.py
DIvyaNautiyal07: Superseded by your follow-up on this thread — CI coverage went green at `5dee223f8` under the aggregate `source = packages/valory` run, so the specific line-level gaps here (OffchainDepositState raises, dict->dataclass branch, degradation path, atomic-write cleanup, head-block non-STATE arm) are now covered by the full-suite roll-up rather than needing targeted per-branch tests. Marking as acknowledged.

DIvyaNautiyal07: Thanks for confirming the gate condition, marking acknowledged. On the two behavioral follow-ups you flagged as optional: I'll open them as a small separate PR after this one merges — one numeric wei->USD ROI assertion on the Gnosis/18dp side, and a Polymarket/6dp variant covering both the token-decimal boundary and the load-bearing invariant comment on decimals in the ROI formula.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
OjusWiZard: Thanks — the conditional preserve + `test_does_not_wipe_offchain_deposits_when_existing_read_degraded` implement exactly what this comment suggested, but on re-checking the real flow I have to own that my one-liner was insufficient, so keeping this open:

The new test pre-populates `new_summary.offchain_deposits` before calling `_save_agent_performance_summary` — but in production that never happens. `_fetch_offchain_prepaid_wei` writes the checkpoint **to disk**, not into the summary object `_fetch_agent_performance_summary` is building, which still constructs with `offchain_deposits=None` (behaviours.py ~2057). So in an actual degraded-read cycle: the conditional skips the assign, the summary's own value is still `None`, and the cycle-end `overwrite_performance_summary` persists `None` anyway — the on-disk checkpoint is lost exactly as before. The guard only protects an in-memory value that the real flow never carries.

Reachability stays narrow (it needs the file to become validation-invalid between the mid-cycle write and the cycle-end read, or a cycle where the scan didn't write and the file was already corrupt), which is why this isn't a blocker — but the robust fixes remain: (a) when the cycle-end read degrades, skip the save entirely (or re-read just `offchain_deposits` leniently before writing), or (b) the split-file checkpoint the code comment already floats, which removes the cascade class altogether.

OjusWiZard (approved):
Approving after three review rounds at head `42ebc13ba`. All Critical/High findings from rounds 1–2 are fixed and verified (param wiring, checkpoint persistence, bounded head-block scanning, atomic writes, hard-index aggregation, JSONLike-safe contract returns), CI is fully green, and both new BalanceTracker defaults were corroborated as live deployments on their respective chains (`eth_getCode` shows the `Deposit(address,address,uint256)` topic in the bytecode on Gnosis and Polygon).

Non-blocking leftovers, fine to handle in a follow-up: the open corruption-cascade thread (degraded-read residual), a few comment/docstring accuracy nits around the new hard-index and flatten blocks (the "ERROR performative arm" wording, and the `encode()`-at-any-boundary claim vs. protobuf Struct's 2^53 double precision), and the `tempfile.mkstemp` 0600 file mode on `agent_performance.json`, which is worth checking against Pearl middleware's cross-process read before release.

packages/packages.json
```diff
@@ -72,7 +72,7 @@
         "skill/valory/registration_abci/0.1.0": "bafybeif4n2esdjvquomdkwktv7eolepskpo5jlqrfda2abxjqzy2ljsxzi",
         "skill/valory/reset_pause_abci/0.1.0": "bafybeifrnmlscfiki7wojs6zasppefupd2wkjyq6orhdllfppjud7isrxi",
         "skill/valory/termination_abci/0.1.0": "bafybeih6sa2ttcvvjk7rcfn7ure2ws23kcfufssb4qmluzcfvednkfpvoi",
-        "skill/valory/mech_interact_abci/0.1.0": "bafybeici5tsgtlypnyrnp5colj7katnsyjfdzt6js4xdbnpdms4232muje",
+        "skill/valory/mech_interact_abci/0.1.0": "bafybeigjdl23b5lzezzjvdetc2pacuv4sf33mgpa6qnt3amwlb5bxpcv3e",
```
bennyjo: **[Blocking]** This `mech_interact_abci` hash was recomputed locally (the balance_tracker edit cascaded into its dependency pin) but never published — both failing CI jobs time out fetching exactly this package, and the gateway can't resolve it (old hash resolves in ~40ms). Even if pushed to IPFS, `third_party` would then pin a mech_interact no upstream release corresponds to — suggest either reverting to a separate trader-local reader contract with its own public id (the original `mech_prepaid_reader` shape) so balance_tracker + mech_interact stay at pinned hashes, or upstreaming the classmethod to valory-xyz/mech and bumping pins properly.

packages/valory/contracts/balance_tracker/tests/test_contract.py
```diff
@@ -0,0 +1,198 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+# ------------------------------------------------------------------------------
+
+"""Encode-assertion tests for the balance_tracker contract package."""
```
bennyjo: Moving this package from third_party to dev pulls it into the coverage scope for the first time, and these tests cover only 2 of its 9 classmethods — the 100% gate that flagged this in round 1 will likely fail again once the packages-sync failure is fixed. Reverting to a small dedicated reader package (per the packages.json comment) also makes this go away, since it'd only carry the two methods already tested here.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +363,194 @@ def _calculate_settled_mech_requests(
...
+                f"(response: {block_response}); leaving cached total."
+            )
+            return cached_total
+
+        body = block_response.state.body
+        raw_head = body.get("get_block_number_result") if body is not None else None
+        # ``< 0`` (not ``<= 0``) so a legitimate genesis-block head on a
+        # fresh devnet/Hardhat chain is accepted, per the
+        # ``OffchainDepositState`` docstring.
+        if raw_head is None or int(raw_head) < 0:
```
bennyjo: `int(raw_head)` raises uncaught if the ledger returns a non-numeric string, crashing the behaviour, while every other malformed input in this helper degrades with a warning — suggest wrapping in `try/except (ValueError, TypeError)` to match (mech-interact's `_get_native_balance` does exactly this).

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +363,194 @@ def _calculate_settled_mech_requests(
...
+        # even with zero matches, so subsequent cycles start from
+        # ``head_block + 1`` (small range) rather than re-scanning the
+        # same growing range every cycle. Without this, a long
+        # deposit-quiet stretch eventually crosses provider ``eth_getLogs``
+        # caps and the ERROR performative freezes future accrual forever
+        # with only warning logs.
+        total_wei = cached_total + new_wei
+        summary.offchain_deposits = OffchainDepositState(
+            total_deposited_wei=total_wei,
+            last_scanned_block=head_block,
```
bennyjo: Advancing the checkpoint to the exact head means a reorg can double-count a near-head deposit that gets re-included, or permanently miss one reorged into an already-scanned block — for irreversible financial state, scanning to `head - N` confirmations (N≈12) is cheap insurance. Happy to leave as-is if you consider reorg exposure acceptable for a reporting metric.

bennyjo (commented):
Re-review — round-1 fixes all verified in code; one new blocking packaging issue from the consolidation refactor, plus follow-ups.

packages/valory/services/polymarket_trader/service.yaml
bennyjo: Checked mech-deployments: the Polygon marketplace is `0x343F2B005cF6D70bA610CD9F1F1927049414B582` but no BalanceTracker address is recorded anywhere in that repo, so zero-default is right for now. The authoritative address can be read off the marketplace's `mapPaymentTypeBalanceTrackers` mapping — worth confirming with the mech team rather than hardcoding a guess.

packages/packages.json
bennyjo: Retracting — this reviewed a stale revision. The unpublished-hash cascade was real at a3bda100/092d5c16 but 07413286 restored the upstream `balance_tracker` package, and head pins the original upstream `mech_interact_abci` hash (`bafybeici5tsg…`, resolves on the gateway). CI is green on head 42ebc13b. Not blocking. Resolving.

packages/valory/contracts/balance_tracker/tests/test_contract.py
bennyjo: Retracting — stale against head: 07413286 already moved `balance_tracker` back to `third_party` and re-extracted the reader into the dedicated `mech_prepaid_reader` package, which is exactly the resolution suggested here. Coverage gate is green on head. Resolving.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: You're right on the trace — my one-liner protected an in-memory value the real flow never carries. Fixed properly in ea87e0261. New `SharedState.read_offchain_deposits_from_disk` reads only the `offchain_deposits` sub-dict from raw JSON with lenient parsing, bypassing `AgentPerformanceSummary.__post_init__` on siblings. `_save_agent_performance_summary` calls it before the atomic overwrite, so a corrupt Achievements or PredictionHistory entry can't cascade into wiping the ROI checkpoint. Replaced the misleading test with `test_offchain_deposits_survives_sibling_corruption_via_raw_reread` that reproduces the real flow (`new_summary.offchain_deposits=None` per the actual builder, mocked disk read returns a real state) plus a reverse-direction test for the first-boot / field-absent case. Split-file checkpoint stays as a follow-up if the raw-JSON approach ever gets tricky.

packages/packages.json
DIvyaNautiyal07: Already resolved on the current tip — Divya reverted to a distinct trader-local reader package in commit 07413286f before this comment landed. `packages.json` now lists `contract/valory/mech_prepaid_reader/0.1.0` under `dev` with `valory/balance_tracker` staying at its upstream third_party hash. `mech_interact_abci` is back at its published hash (no local cascade), so `autonomy packages sync` fetches cleanly. Effectively the shape you suggested — flagging as acknowledged.

packages/valory/contracts/balance_tracker/tests/test_contract.py
DIvyaNautiyal07: Superseded by 07413286f — the balance_tracker revert means this file is back to its upstream shape (only the pre-existing `build_deposit_for_data` tests) and the reader classmethod + its 4 tests live in the trader-owned `mech_prepaid_reader` package (`tests/test_contract.py` there). Coverage scope moved with it, no gate regression, matches the shape you suggested.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fair, that was the one branch in this helper that raised uncaught while every other malformed path degrades. Fixed in ea87e0261 — wrapped `int(raw_head)` in `try/except (TypeError, ValueError)`, degrades with a warning identifying the invalid value, matches `mech_interact_abci._get_native_balance`. Regression test `test_head_block_non_numeric_string_degrades_with_warning` covers the non-numeric string case.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Deliberately leaving as-is for this round — Gnosis has ~5s block time with rare shallow reorgs (typical 1-2 blocks, rarely 3+), Polygon similar profile, and the ROI cost term is a reporting metric not consensus-critical. The scan-to-head + advance-checkpoint-to-head shape trades N-confirmations delay for immediate accounting response and a slightly simpler checkpoint invariant. If we ever move the accounting to a chain with deeper reorgs, or the metric graduates to something safety-critical, the fix is one line: `head_block = max(0, current_head - REORG_CONFIRMATIONS)`. Named the tradeoff in this thread as documentation for the future reader.

packages/valory/services/polymarket_trader/service.yaml
DIvyaNautiyal07: Fair operational point on going to `mapPaymentTypeBalanceTrackers` for the authoritative address. For this initial rollout I sourced the two defaults from the marketplace repo's `scripts/deployment/globals_polygon_mainnet.json` (`balanceTrackerFixedPriceTokenUSDCAddress = 0x5C50ebc17d002A4484585C8fbf62f51953493c0B`, matches the Polygon marketplace `0x343…4B582` and Polymarket's USDC payment path) and `scripts/deployment/globals_gnosis_mainnet.json` (`balanceTrackerFixedPriceNativeAddress = 0x21cE6799…` for wxDAI on Gnosis). The globals JSONs are what the marketplace deployment scripts commit to, so treating them as authoritative for a starting default felt safe. If you'd rather ship zero-default until an on-chain `mapPaymentTypeBalanceTrackers` read confirms per-chain, happy to revert both.

OjusWiZard (approved):

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +365,207 @@ def _calculate_settled_mech_requests(
...
+            summary.timestamp = self.shared_state.synced_timestamp
+            self.shared_state.overwrite_performance_summary(summary)
+            self.context.logger.info(
+                f"[Offchain Deposits] First run — seeded checkpoint to block "
+                f"{head_block}, cumulative total starts at 0."
+            )
+            return 0
+
+        # Subsequent runs: bound the scan range up front.
+        from_block = state.last_scanned_block + 1
```
bennyjo, line 529: If the agent is offline longer than the provider's `eth_getLogs` block-range cap (~14h of Gnosis blocks — a Pearl laptop closed for a weekend), every scan of `[from_block, head]` fails and the checkpoint can never advance, freezing deposit accrual permanently. Suggest `to_block = min(head_block, from_block + MAX_SCAN_RANGE)` so successive cycles catch up in chunks.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +365,207 @@ def _calculate_settled_mech_requests(
...
+            ``BalanceTracker.Deposit`` credit to this Safe up to and
+            including the latest block scanned.
+        :yield: contract-API response(s) via the abstract-round-abci
+            base behaviour helpers.
+        """
+        # Read cached_total BEFORE any gate so historical accrued spend
+        # under pre-deposit-as-loss stays counted even if the operator
+        # flips use_offchain off or clears the tracker address. Once
+        # counted, always counted.
+        summary = self.shared_state.read_existing_performance_summary()
```
bennyjo: This strict read still feeds the seed decision: if it degrades on an unrelated field (e.g. rollback after a newer release added a summary field), the first-run branch re-seeds and overwrites a still-valid on-disk `total_deposited_wei` with 0. Suggest seeding `state` from `read_offchain_deposits_from_disk()` here too, so only a genuinely absent checkpoint takes the seed path.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -357,6 +365,207 @@ def _calculate_settled_mech_requests(
...
+        # contract classmethods return processed log entries; sum happens
+        # in the caller.
+        # Hard-index the flat entry shape returned by
+        # ``get_deposit_events_for_requester`` (``{"amount": int,
+        # "block_number": int}`` per entry). ``.get(..., 0)`` would
+        # silently zero-fill a shape drift while still advancing the
+        # checkpoint below to ``head_block`` — an irreversible under-count
+        # that would flatter ROI. Raise loudly instead so the ERROR
+        # performative arm above catches it and the checkpoint stays put.
+        new_wei = sum(int(entry["amount"]) for entry in entries)
```
bennyjo: This comment isn't accurate — a raise here isn't caught by the performative check above, and with `skill_exception_policy: stop_and_exit` a shape drift terminates the whole agent instead of freezing the checkpoint as described. Suggest catching `(KeyError, TypeError, ValueError)` around the sum and routing to the existing malformed-response branch.

bennyjo (commented):
Round-4 re-review: `ea87e0261` verified. Three remaining items inline: scan-range clamp for downtime, mid-cycle seed path still on the strict reader, and an inaccurate fail-safe comment.

packages/valory/contracts/balance_tracker/tests/test_contract.py
DIvyaNautiyal07: Ack, thanks — confirmed on our end too. Coverage green on `5dee223f8`, `mech_prepaid_reader` is the trader-owned reader package, `balance_tracker` back at its upstream third_party hash. Resolving.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: You're right, my earlier reasoning was wrong. Verified in `aea-config.yaml`: `skill_exception_policy: stop_and_exit`, so an unhandled skill exception terminates the agent, and the `performative != STATE` gate above the sum only catches API performatives — not `KeyError` from `entry["amount"]`. Fixed in 3ba6af855 by wrapping the sum in `try/except (KeyError, TypeError, ValueError)` and routing into the malformed-response branch. Regression test `test_malformed_entry_shape_returns_cached_total_no_raise` feeds entries missing the `amount` key and asserts no raise, cached_total returned, warning logged.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Real reliability gap for Pearl. Fixed in 3ba6af855. Added `OFFCHAIN_MAX_SCAN_RANGE = 5000` (safely below public-provider caps like Alchemy/Infura ~10k) and clamped `to_block = min(head_block, from_block + OFFCHAIN_MAX_SCAN_RANGE - 1)`. An agent that comes back from a weekend offline now marches the checkpoint forward one chunk at a time until it catches up, instead of every scan hitting the provider cap and permanently freezing. Regression test `test_scan_range_capped_at_offchain_max_scan_range` uses a head 20× the max range and asserts (a) the contract call sees the clamped `to_block`, (b) the checkpoint advances only to the scanned upper bound.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: You're right that the narrow fix leaves the write side of the same cascade class exposed. Went with the complete split-write in 3ba6af855 to close both sides. New `SharedState.write_offchain_deposits_to_disk` mirrors `read_offchain_deposits_from_disk`: reads the raw JSON dict, updates only the `offchain_deposits` sub-dict, writes back atomically. `_fetch_offchain_prepaid_wei` now sources `state` from the raw-read helper and persists via the raw-write helper — it never touches `read_existing_performance_summary` or `overwrite_performance_summary`, so a nested-field `__post_init__` raise elsewhere on disk can neither trigger a spurious re-seed nor wipe sibling fields at write time. Three model-level tests around the split-write (sibling preservation with a real file round-trip, minimal file when nothing on disk, recovery from a truncated JSON prefix). The `TestFetchOffchainPrepaidWei` helper now asserts (via `AssertionError` side-effects on the deprecated methods) that the helper does NOT touch the whole-summary paths — a future refactor that regresses that gets caught immediately.

bennyjo (approved):
Round-5 re-review at `3ba6af855`: all findings from rounds 1–5 resolved and regression-tested — scan-range cap with lossless chunked catch-up, split read/write insulating the irreversible checkpoint from sibling corruption, and the shape-drift crash routed to the malformed-response branch. BalanceTracker defaults independently verified against the marketplace deployment globals and on-chain code. Approving; merge once the test matrix is green.

PR #1010, opened 2026-07-16
chore(polymarket-trader): restore max_edge 0.15 -> 0.30 (re-apply #1004)

DIvyaNautiyal07:
## Summary
- PR #1004 widened polymarket_trader `max_edge` from 0.15 to 0.30. That change was silently reverted during the `main` -> `offchain-epic` sync merge (3bf0e2718) — the conflict resolution kept the branch's stale 0.15 side. PR #1008 then carried the reverted value back to main, so current `main` shows `max_edge=0.15` even though #1004 is in the history.
- Re-applies the exact same two-line change as #1004: service.yaml + regenerated `packages.json` hash. No other config from #1004 was affected (verified).

## Test plan
- [ ] `autonomy packages lock --check` passes (already run locally)
- [ ] CI green (packages hash check, docstrings, deps)
- [ ] Confirm live Polymarket agent picks up widened band once merged

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -19,7 +19,7 @@
         "skill/valory/chatui_abci/0.1.0": "bafybeig4qckyl7exilbd4egapsttvfrecfu4buywx632gixuvkwcjqqssu",
         "agent/valory/trader/0.1.0": "bafybeif3ttffq4o6xzrzh7kdgjrx5tmy4sjki7gaw7smbhfnfhqhcny4ym",
         "service/valory/trader_pearl/0.1.0": "bafybeigfo273qta5ofgmidchoy6zy5ai464xi4prydsp4gen2mhgiy7rg4",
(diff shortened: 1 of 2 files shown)


OjusWiZard (approved):

jmoreira-valory (approved):
Verified the revert chain rather than relaying it: #1004's merge commit (`7ffeb815e`) is the second parent of sync-merge `3bf0e2718`, whose polymarket `service.yaml` resolution kept only the agent-hash bump — the `max_edge 0.30` side was dropped there, and #1008 carried the stale `0.15` back to `main`. Since #1004 was exactly this same two-line change, nothing else from it was lost.

Override chain checked: `aea-config.yaml` wraps `strategies_kwargs` in `${STRATEGIES_KWARGS:dict:…}`, so the service-level default flows through (and `service.yaml` is the only layer defining `max_edge` — the agent-level default dict has no `max_edge` key). `lock_check` green on all three OSes confirms the regenerated CID reproduces.

One non-blocking note: the restored default only takes effect where `STRATEGIES_KWARGS` isn't pinned in the deployment env — your test-plan item to confirm the live agent picks up the widened band is the right guard for that (no pin found in agent-deployments or quickstart).

PR #1013, opened 2026-07-23
feat(agent-performance-summary): mech-analytics as an alternative source for mech data, flag-gated

DIvyaNautiyal07:
## Summary

Adds a second source for the four pieces of mech data the trader today reads from the marketplace subgraph. When the marketplace switches off-chain, that subgraph stops seeing the content of new requests (titles, tool names, response payloads). Without this branch: day-bucketed fee attribution, per-position tool attribution, prediction fallbacks, and the aggregate total-count all silently go wrong. The mech-analytics HTTP endpoints keep working post-switch because they read from a private database.

Behind `USE_MECH_ANALYTICS` (default `false`, merging changes nothing):

- **Daily fee attribution** (`_fetch_all_mech_requests` in `graph_tooling/requests.py`) — pages `/v1/data/scored-rows` with a chain-scoped keyset cursor, reshaped to the same list-of-dict shape the subgraph returned so the day-bucketing / title-matching in `behaviours.py` is untouched.
- **Per-position tool attribution** (`fetch_mech_tool_for_question`, both Omen and Polymarket fetchers) — filters scored-rows client-side by `question_title`, reads `tool` off the matched row.
- **Prediction fallback** (`_fetch_prediction_response_from_mech`, both fetchers) — same client-side filter, returns `{p_yes, p_no, confidence, info_utility}` off the row.
- **Aggregate total-count** (`_get_total_mech_requests`) — hits `/v1/metrics/requester/{chain}/{safe}` and reads `windows.all.n_mech_requests`. The subgraph counter it used to read (`sender.totalMarketplaceRequests`) only increments on the on-chain `MarketplaceRequest` event and freezes for off-chain volume, which would silently under-count `all_mech_costs` post-switch (found by @jmoreira-valory on mech-analytics PR #11).

## Depends on

- valory-xyz/mech-analytics#14 — `info_utility` column, `chain_id` on `agent_aggregates`, `/v1/metrics/requester/{chain_id}/{address}` endpoint. This PR must land after that one.

## Chain-safety

Every mech-analytics call passes `chain_id` explicitly. Gnosis Safe proxies CREATE2 to the same address on every chain by design, so an unfiltered query would sum a Safe's activity across chains without saying so.

## Blast radius

- Flag off: no code path changes. Every existing test continues to pass because the flag check uses `is True` (strict, not truthy), so `MagicMock` contexts in existing tests don't accidentally flip themselves onto the new path.
- Flag on with `MECH_ANALYTICS_URL` unset: throws rather than silently returning zero (which would zero the cost leg and inflate ROI).
- Rollback: flip flag off.

## Test plan

Local (mirrors CI `checks` job):

- [x] `uv run tomte tox -e black-check`
- [x] `uv run tomte tox -e isort-check`
- [x] `uv run tomte tox -e flake8`
- [x] `uv run tomte tox -e mypy`
- [x] `uv run pytest packages/valory/skills/agent_performance_summary_abci/tests/` — 1133 passed
- [x] `uv run autonomy packages lock`

Not run locally (needs full CI env):

- [ ] `tomte tox -e pylint -e darglint -e safety -e bandit`
- [ ] `tomte tox -e check-abci-docstrings -e check-abciapp-specs -e check-handlers`
- [ ] `tomte tox -e check-doc-hashes -e check-third-party-hashes -e check-packages -e check-hash`

New tests:

- 43 unit tests for the mech-analytics HTTP client (pagination edge cases, `next_cursor` `undefined`→`None` normalisation, cursor-cycle safety cap, chain-scoped queries, JSON / shape drift guards, the `/requester` endpoint)
- 10 unit tests for flag branching on both fetchers, including an explicit "flag off does not call mech-analytics client" assertion that would catch a future refactor turning the flag into a no-op

## Verification before flipping the flag on a Pearl deployment

Per `docs/consumer_migration.md` §7 in mech-analytics:

1. mech-analytics catch-up on the target chain complete
2. Set `MECH_ANALYTICS_URL` + `USE_MECH_ANALYTICS=true` on one staging trader
3. Compare its daily-profit report against the current subgraph-sourced number for a Pearl-staked Safe over a 7d window
4. Ship to Pearl on the natural release cycle (per `docs/consumer_migration.md` §7 "Release timing" — trader does not need to be gated on the mech-analytics Day N cut-over because mech-analytics indexes both on-chain and off-chain data)


packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -278,6 +283,35 @@ def _get_total_mech_requests(
...
+                address=agent_safe_address,
+                logger=self.context.logger,
+            )
+            # None means transport / schema failure. An unknown Safe
+            # returns a zero-shaped body (not None), so a real 0 count
+            # still caches correctly; only failures skip caching so the
+            # next tick can retry rather than baking a wrong 0 in and
+            # inflating ROI for the rest of the round.
+            if usage is None:
+                return 0
```
OjusWiZard: **[Critical]** On `fetch_requester_usage` failure (any non-200, connection error, malformed JSON, missing `windows` key) this returns a literal `0` — which flows into `all_mech_costs = total_mech_requests * DEFAULT_MECH_FEE` and `settled_mech_costs`. A single transient mech-analytics outage zeroes the cost leg and inflates reported ROI for the round.

The comment above says failures only "skip caching so the next tick can retry rather than baking a wrong 0 in" — but this `return 0` *is* the wrong 0; skipping the cache doesn't help the current round's math. The caller can't distinguish "confirmed zero requests" from "fetch failed" because the return type collapses to bare `int`.

Suggest returning `Optional[int]` (or raising) on `usage is None` and short-circuiting the cost computation, mirroring the `None`-vs-`[]` discipline `_fetch_all_mech_requests` already follows.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -278,6 +283,35 @@ def _get_total_mech_requests(
...
+        # per-address usage endpoint. The subgraph counter this used to
+        # read (``sender.totalMarketplaceRequests``) is only incremented
+        # on the on-chain MarketplaceRequest event and freezes for
+        # off-chain volume post-switch, so the pre-migration read
+        # undercounts ``all_mech_costs`` as off-chain adoption grows.
+        # ``/v1/metrics/requester/{chain_id}/{address}`` returns the
+        # per-window rollups for this exact Safe on this exact chain
+        # (Gnosis Safe proxies CREATE2 to the same address on every
+        # chain, so ``chain_id`` scoping is required).
+        if is_flag_enabled(self.params):
```
OjusWiZard, line 311: **[Critical]** This migrates the *total* to mech-analytics, but `_get_open_market_requests` (just below) still reads `sender.requests` from the marketplace subgraph — the exact field this PR says stops seeing off-chain volume. `_calculate_settled_mech_requests` then computes `settled = total (complete, mech-analytics) − open (frozen, subgraph)`, so once volume moves off-chain the subtraction **overstates** settled requests and corrupts `settled_mech_costs` — silently, and precisely when the flag is doing its job.

Either `_get_open_market_requests` needs a flag-on path too (scored-rows filtered by open-market titles within the window), or the mismatch should be documented as a known limitation with the settled/open split disabled under the flag.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
```diff
@@ -641,6 +647,23 @@ def _fetch_all_mech_requests(
         self, agent_safe_address: str
     ) -> Generator[None, None, Optional[List]]:
         """Fetch all mech requests for the agent with pagination support."""
+        # Flag-on path: read the per-Safe request list from mech-analytics'
+        # /v1/data/scored-rows instead of the marketplace subgraph. Same
+        # downstream shape via ``rows_as_subgraph_mech_requests`` so the
+        # day-bucketing / title-matching in behaviours.py is untouched.
+        # chain_id is passed explicitly — a Safe address can exist on
+        # multiple chains and an unfiltered call would sum them silently.
+        if is_flag_enabled(self.params):
```
OjusWiZard, line 727: **[Critical]** Two issues on this branch:

1. **Partial migration.** Only this one-time backfill is flag-gated. `_fetch_mech_requests_by_titles` (line ~712) — used on *every tick* by `_perform_incremental_update` for per-day fee attribution — still always queries the subgraph. Post off-chain-switch, days whose titles stop matching silently land on `fees_by_day.get(ts, 0)`, understating mech costs and overstating daily profit with no log line (and if *no* titles match, the H6 guard freezes `profit_over_time` forever with the round still green). Enabling the flag doesn't actually fix the steady-state accounting the PR description says goes wrong.

2. **Blocking I/O in a behaviour.** `fetch_scored_rows` uses synchronous `requests.get` with no `yield` — see comment in `mech_analytics_client.py`. Every other fetch in this file goes through `yield from` / `get_http_response`.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,372 @@
...
+        }
+        if since is not None:
+            params["since"] = since.isoformat()
+        if until is not None:
+            params["until"] = until.isoformat()
+        if cursor is not None:
+            params["cursor"] = cursor
+
+        try:
+            response = requests.get(url, params=params, timeout=REQUEST_TIMEOUT_SECONDS)
```
OjusWiZard: **[Critical]** Synchronous `requests.get` (30s timeout) inside a loop capped at `MAX_PAGES=200`, called from `FetchPerformanceSummaryBehaviour` code paths (`_fetch_all_mech_requests`, `_get_total_mech_requests`) with no `yield`. The agent is single-threaded: while this runs, ABCI message handling, Tendermint heartbeats, and the chatui server all stall — worst case minutes, on a live trading agent.

The framework's generator-based `get_http_response` machinery exists exactly for this; the prediction-helper call sites at least match a pre-existing blocking pattern reached only via chatui, but the `requests.py`/`behaviours.py` consumers are new blocking calls inside the core FSM round.

packages/valory/skills/agent_performance_summary_abci/models.py
```diff
@@ -334,6 +334,15 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.balance_tracker_address: str = self._ensure(
             "balance_tracker_address", kwargs, str
         )
+        # Mech-analytics migration: base URL of the read-only mech-analytics
+        # API and a feature flag gating whether the trader reads request
+        # data from that API instead of the marketplace subgraph. Empty URL
+        # disables the flag-on path defensively (throws in the client rather
```
OjusWiZard, line 340: **[Important]** "throws in the client" — nothing throws, anywhere in the new code. `is_flag_enabled` silently returns `False` when `mech_analytics_url` is empty (no log), so every call site quietly falls back to the legacy subgraph path, and the client's own empty-URL warning branches are unreachable in production. The PR description's blast-radius claim ("flag on with URL unset: throws rather than silently returning zero") is likewise not what the code does.

An operator who sets `USE_MECH_ANALYTICS=true` but forgets the URL gets a permanent, invisible no-op — the trader keeps reading the counters the migration exists to replace. Suggest enforcing the pairing here in `__init__` (raise at startup when flag is true and URL is empty), which also makes the comment true again.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -720,6 +749,36 @@ def _fetch_prediction_response_from_mech(
...
+                until=until,
+                logger=self.logger,
+            )
+            if rows is None:
+                return None
+            matched = find_latest_row_for_title(rows, question_title)
+            if matched is None:
+                return None
+            return {
+                "p_yes": matched.get("p_yes"),
```
OjusWiZard: **[Important]** These are nullable columns, so `matched.get("p_yes")` can return a present-but-`None` value — and the comment above is wrong about the safety net: `dict.get(key, default)` only substitutes when the *key is absent*, not when its value is `None`. Downstream, `_format_bet_for_position` does `prediction_response.get("p_yes", 0) * 100`, which raises `TypeError` on `None`; the broad `except Exception` in `fetch_position_details` then swallows it and silently drops the entire position from the report.

The flag-branching test only asserts the intermediate dict contains `None`s — it never pushes one through `_format_bet_for_position`. Suggest coalescing here (`matched.get("p_yes") or 0` per field, or skipping null rows), plus a test that formats a null-column row end-to-end. Same fix needed in the Polymarket mirror.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
```diff
@@ -634,6 +663,33 @@ def _fetch_prediction_response_from_mech(
...
+                until=until,
+                logger=self.logger,
+            )
+            if rows is None:
+                return None
+            matched = find_latest_row_for_title(rows, question_title)
+            if matched is None:
+                return None
+            return {
+                "p_yes": matched.get("p_yes"),
```
OjusWiZard: **[Important]** Same nullable-column issue as the Omen mirror (`predictions_helper.py`): a present-but-`None` `p_yes`/`p_no`/`confidence` survives to `_format_bet_for_position`'s `.get(field, 0) * 100` and raises `TypeError`, which `fetch_position_details` swallows — silently dropping the position.

OjusWiZard (commented):
Multi-perspective review (correctness, silent failures, tests, types, comments, simplification). Inline comments below carry the file-specific findings; these ones don't anchor to a single diff line:

**[Critical] Coverage gate fails.** Running the coverage-gated suite over the changed modules yields **96.20% vs the required 100%** — the missing lines are exactly the new flag-on branches in `behaviours.py:296-313` and `requests.py:655-663` (`test_behaviours.py` / `test_requests.py` were not updated). Beyond the gate, these branches carry the ROI-affecting key-chain parsing (`windows.all.n_mech_requests`), so a typo'd key would silently produce 0 with no test signal.

**[Important] Asymmetric failure-path tests.** The Omen `fetch_mech_tool_for_question` tests both the `rows is None` and `matched is None` branches, but the Polymarket mirror and `_fetch_prediction_response_from_mech` (both variants) only test the happy path — exactly where copy-paste divergence slips through. `fetch_requester_usage` also has no malformed-JSON test (its `except ValueError` is uncovered), and it feeds the untested `behaviours.py` branch.

**[Advisory] `_build_mech_request_lookup` (`behaviours.py:1511`)** treats the new `None` (fetch failure) return of `_fetch_all_mech_requests` identically to `[]` (genuinely empty), discarding the distinction this PR introduced. Smaller blast radius than the `_get_total_mech_requests` issue (not cached, self-corrects next round), but same pattern.

**[Advisory] Rot-prone references.** Three comments cite "PR #14" and one cites "consumer migration §7" in the external mech-analytics repo — unpinned cross-repo references that will silently go stale. Consider linking full URLs or describing the schema fact directly.

What's solid: the client's 43 unit tests are genuinely correctness-oriented (pagination order, cursor-cycle cap, mid-pagination failure returning None instead of partial data, zero-vs-None for unknown Safes), chain-scoping is consistently explicit, and license headers are correct.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -720,6 +749,36 @@ def _fetch_prediction_response_from_mech(
...
         ts = bet_timestamp or int(datetime.now(timezone.utc).timestamp())
+
+        # Flag-on path: read prediction fields directly off the scored row.
+        # ``p_yes``/``p_no``/``confidence``/``info_utility`` are all columns
+        # on ``per_request_scores`` (info_utility added in mech-analytics
+        # PR #14). Downstream consumers (e.g. position-details display) use
+        # ``.get(field, 0)`` so a NULL column lands as a graceful 0 rather
+        # than a KeyError, matching the old JSON-parse-tolerant behaviour.
+        params = getattr(self.context, "params", None)
+        if is_flag_enabled(params):
```
OjusWiZard, line 778: **[Advisory]** This flag-on block is byte-for-byte identical to the one in `polymarket_predictions_helper.py` (and likewise for the `fetch_mech_tool_for_question` pair) — 4 blocks, ~17 lines each, with zero platform-specific content beyond `chain_id_for_platform(...)`, which is already shared. Consider hoisting into two helpers in `mech_analytics_client.py` (e.g. `fetch_tool_via_analytics(...)`, `fetch_prediction_response_via_analytics(...)`); the bodies are pure functions of `(params, sender_address, question_title, ts, logger)`. The flag-branching tests would just need their patch targets moved to `mech_analytics_client.fetch_scored_rows`.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,372 @@
...
+    :param base_url: mech-analytics base URL (e.g.
+        ``https://mech-analytics.autonolas.tech``). Trailing slash tolerated.
+    :param requester: the Safe address. Endpoint lowercases before matching,
+        so mixed-case checksummed addresses work.
+    :param chain_id: chain id to scope the query to. Pass explicitly rather
+        than defaulting — a Safe address can exist on multiple chains, and
+        an unfiltered call sums their activity silently.
+    :param since: optional lower bound on ``requested_at`` (inclusive).
+        Omit for the all-time count use case.
+    :param until: optional upper bound on ``requested_at`` (exclusive).
```
OjusWiZard, line 312: **[Advisory]** Boundary-semantics conflict between two new docstrings: `until` is documented here as **exclusive**, while `find_latest_row_for_title` claims to mirror the old subgraph's `blockTimestamp_lte` (**inclusive**) semantics. One of the two is wrong — and if it's the code, a mech request timestamped exactly at the bet time is included on the flag-off path but silently dropped on the flag-on path. Worth pinning down against the actual endpoint behaviour and making the two docstrings agree.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,372 @@
...
+            return None
+
+        all_rows.extend(rows)
+
+        # Endpoint returns ``next_cursor: null`` (JSON null → Python None)
+        # when the page is the last one. Normalise ``undefined`` (absent
+        # key) to None as well, so a schema drift that dropped the key
+        # doesn't produce a literal ``cursor=None`` string on the next
+        # request.
+        cursor = payload.get("next_cursor") or None
```
OjusWiZard: **[Advisory]** The comment's rationale doesn't match what this protects against: `.get("next_cursor")` already returns `None` for an absent key, and the loop below only sends `cursor` when it's not `None`, so a literal `cursor=None` query param was never possible. The only behaviour `or None` actually changes is a present-but-falsy cursor (e.g. `""`) — which is the real edge case worth stating.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,372 @@
...
+
+    :param iso_string: an ISO-8601 UTC timestamp, ``Z`` or ``+00:00``
+        suffix accepted, or ``None``.
+    :return: Unix seconds since epoch, or 0 if the input is
+        None / empty / unparseable.
+    """
+    if not iso_string:
+        return 0
+    try:
+        # datetime.fromisoformat accepts the Z suffix on 3.11+, which is
```
OjusWiZard: **[Advisory]** Two inaccuracies: the trader's minimum Python is 3.10 (`pyproject.toml`: `>=3.10`), not 3.11 — and the claim is moot anyway, since the next line `replace("Z", "+00:00")`s before calling `fromisoformat`, so native Z-suffix support is never relied on.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,372 @@
...
+    ``None`` (behaviour-side callers that pre-check ``self.params``).
+
+    :param params: params-like object (or ``None``) exposing
+        ``use_mech_analytics`` and ``mech_analytics_url`` attributes.
+    :return: True when the migration flag is truly enabled and a URL
+        is configured; False otherwise.
+    """
+    if params is None:
+        return False
+    return getattr(params, "use_mech_analytics", False) is True and bool(
```
OjusWiZard, line 117: **[Advisory]** `params: Any` + `getattr` means mypy gives zero protection here: a future rename of `use_mech_analytics` on the params class makes this silently return `False` forever (the migration path goes dark with no static or test signal). The `is True` workaround exists only to tolerate bare-`MagicMock` contexts in the fetcher tests. A narrow `Protocol` (or `spec=AgentPerformanceSummaryParams` mocks in the tests) would make the workaround non-load-bearing.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -667,6 +673,29 @@ def fetch_mech_tool_for_question(
...
+        # attribution keeps working.
+        params = getattr(self.context, "params", None)
+        if is_flag_enabled(params):
+            assert params is not None  # narrowed by is_flag_enabled  # nosec B101
+            until = datetime.fromtimestamp(ts, tz=timezone.utc)
+            rows = fetch_scored_rows(
+                base_url=params.mech_analytics_url,
+                requester=sender_address,
+                chain_id=chain_id_for_platform(params.is_running_on_polymarket),
+                until=until,
```
bennyjo, line 692: No `since` bound means this pages the Safe's full mech history for every position (again at :767), which can block the behaviour past the round timeout — could `since` be narrowed to just before the bet, or the rows fetched once per round and reused?

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,372 @@
...
+    """
+    if not iso_string:
+        return 0
+    try:
+        # datetime.fromisoformat accepts the Z suffix on 3.11+, which is
+        # the trader's minimum Python. Also accepts ``+00:00`` form.
+        normalised = iso_string.replace("Z", "+00:00")
+        return int(datetime.fromisoformat(normalised).timestamp())
+    except (ValueError, TypeError):
+        return 0
```
bennyjo, line 496: Every other failure here logs, but an unparseable `requested_at` silently becomes `0` and drops the row into the unmatched bucket — could a format drift log once per batch so fee attribution doesn't quietly go to zero?

bennyjo (commented):
Two questions on the flag-on path — both about how it behaves when the data or the endpoint misbehaves.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fixed in `a227d1732`. Changed `_get_total_mech_requests` return type to `Optional[int]` and return `None` on fetch failure (no cache write). Propagated through `_calculate_settled_mech_requests` (also `Optional[int]` now) and `_calculate_performance_metrics` (returns `Optional[PerformanceMetricsData]`); on `None` the caller renders an empty `PerformanceMetricsData()` — same fallback shape as the trader-agent fetch-failure path — instead of a partial with a wrong-0 in `all_mech_costs`. Also updated `_fetch_agent_performance_summary` to preserve the cached `_settled_mech_requests_count` on failure rather than overwrite with 0.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Fixed in `a227d1732`. `_get_open_market_requests` now has its own flag-on branch that pulls scored rows for the Safe from mech-analytics (windowed on `market_open_timestamp`), reshapes them through `rows_as_subgraph_mech_requests`, and runs the same open-market-title matching. Uses the async pagination helper, so no blocking I/O either.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
DIvyaNautiyal07: Both fixed in `a227d1732`. `_fetch_mech_requests_by_titles` now has a flag-on branch: fetch scored rows via mech-analytics (windowed on `block_timestamp_gt`), filter to the requested title set client-side, reshape. Under the flag, the per-tick incremental path no longer silently drops off-chain days. Blocking I/O addressed by the async refactor (see the reply on `mech_analytics_client.py`).

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Fixed in `a227d1732`. Added `_page_mech_analytics_scored_rows` on `APTQueryingBehaviour` that pages `/v1/data/scored-rows` via the framework's `yield from self.get_http_response(...)` — no more sync `requests.get` on the FSM path. All three behaviour-side callers (`_fetch_all_mech_requests`, `_fetch_mech_requests_by_titles`, `_get_open_market_requests`) route through it, plus `_get_total_mech_requests` uses `get_http_response` directly for its single `/requester` call. Kept the sync `fetch_scored_rows` / `fetch_requester_usage` in the client for the helper-side callers (`predictions_helper.py`, `polymarket_predictions_helper.py`) which are only reached via chatui and match your "pre-existing blocking pattern reached only via chatui" carve-out. URL builder (`build_scored_rows_url`, `build_requester_url`) and payload parsers (`parse_scored_rows_page`, `parse_requester_payload`) are factored so the sync and async paths share the same URL construction and shape validation.

packages/valory/skills/agent_performance_summary_abci/models.py
DIvyaNautiyal07: Fixed in `a227d1732`. `AgentPerformanceSummaryParams.__init__` now raises `ValueError` when `use_mech_analytics=True` and `mech_analytics_url` is empty. Fail-loud at startup, matches the "throws rather than silently returning zero" contract the PR description claimed but the code didn't do.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
DIvyaNautiyal07: Fixed in `a227d1732`. Coalesced each field with `matched.get(k) or 0` so a present-but-None value (which slips through `.get(k, 0)` downstream) doesn't reach the `* 100` multiplication and TypeError-then-silent-drop. Updated the existing null-fields test to assert the new zero-coalescing behaviour and added a sanity assertion that multiplying each returned field by 100 doesn't raise. Same fix on the Polymarket mirror.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/polymarket_predictions_helper.py
DIvyaNautiyal07: Fixed in `a227d1732`. Same `or 0` coalesce applied here as on the Omen mirror.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
DIvyaNautiyal07: Skipped in `a227d1732`, deliberately. The DRY win is real (~34 lines saved) but hoisting into two shared helpers in the client would introduce an abstraction the reviewers on this PR would then reasonably want independently tested. If the four blocks ever diverge (platform-specific behaviour creeps in), the refactor is cleaner as a follow-up.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Fixed in `a227d1732`. Rewrote `find_latest_row_for_title`'s docstring so it no longer claims to mirror `blockTimestamp_lte` — it just picks the latest from what was fetched, and the caller owns the boundary. The four per-position call sites now add `+1 second` to `until` when converting `bet_timestamp`, mapping the endpoint's exclusive `requested_at < until` onto the old subgraph's inclusive `blockTimestamp_lte` so a request timestamped exactly at `bet_timestamp` still counts.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Fixed in `a227d1732`. Comment rewritten to name the actual case: `.get()` already returns `None` for an absent key, so the `or None` is there to coerce a present-but-falsy value (empty string, 0) that would otherwise leak into the next request's params.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Fixed in `a227d1732`. Dropped the Python-version-specific claim from the comment. The `Z` suffix is normalised to `+00:00` before `fromisoformat` anyway, so native Z-suffix support is never relied on.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Skipped in `a227d1732`. Correct that `params: Any` + `getattr` gives mypy zero help — the `is True` guard is the workaround for MagicMock-shaped contexts in existing fetcher tests. Migrating to `Protocol` (or `spec=AgentPerformanceSummaryParams` in the mocks) would require touching every existing fetcher-side test fixture. Filed as a follow-up if we ever drop bare-MagicMock usage in those tests.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
DIvyaNautiyal07: Good catch, both are real. Added `PER_POSITION_LOOKUP_WINDOW_DAYS = 30` in `mech_analytics_client.py` and now pass `since = until - 30d` at all four per-position sites (Omen path 2 + 3, Polymarket path 2 + 3). Trader's prediction-to-bet lag is seconds to minutes so 30d is a very safe margin, and it trims a long-history Safe from paging the full history down to at most one page per position on cold-start. Kept it as a `since` narrowing rather than a per-round cache because it's simpler and adds no state on the fetcher. Pinned by 4 new tests so a future refactor that drops the `since=` kwarg fails loudly.

Fixed in aa04a1466.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Agreed, silent zero on parse failure was the wrong default here. Added a module-level `_LOGGER` and `_iso_to_unix_seconds` now emits a WARNING with the offending value (truncated to 100 chars) whenever the parse actually fails. Kept `None` and `""` on the silent path since those are expected for rows with genuinely no timestamp and logging every one would be noise. New test pins the log for parse failures and asserts it stays quiet for missing values.

Fixed in aa04a1466.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -2066,9 +2195,13 @@ def _fetch_agent_performance_summary(self) -> Generator[None, None, None]:
...
         agent_safe_address = self.synchronized_data.safe_contract_address
-        self._settled_mech_requests_count = (
-            yield from self._calculate_settled_mech_requests(agent_safe_address)
-        )
+        # ``None`` from ``_calculate_settled_mech_requests`` means the
+        # total-count fetch failed this round. Leave the cached value
+        # in place (previous round's known-good count) rather than
+        # overwriting with 0, which would inflate settled costs.
+        settled = yield from self._calculate_settled_mech_requests(agent_safe_address)
+        if settled is not None:
```
OjusWiZard, line 2264: **[Critical]** This guard doesn't do what the comment says: `FetchPerformanceSummaryBehaviour` is re-instantiated on every entry into its round, so `self._settled_mech_requests_count` always starts from its `__init__` default of `0` (line 162) — there is no "previous round's known-good count" to preserve. On a mech-analytics fetch failure, `_calculate_settled_mech_requests` correctly returns `None`, this guard skips the assignment, and `calculate_roi()` (lines 741/782) then multiplies the stale `0` into `settled_mech_costs_usd`. `calculate_roi` has no None-check on the count, so `final_roi` comes back as a real float — not `NA` — and the NA-preservation fallback in `_save_agent_performance_summary` never triggers: the cost-undercounted "Total ROI" (the `is_primary=True` metric users see) **overwrites the previously good persisted value**.

This reproduces the round-1 `return 0` bug one hop further down the chain. Suggest propagating the failure into `calculate_roi` (e.g. track `settled` as `Optional` on the instance and return `(None, None)` when unknown), or persisting the count somewhere that genuinely survives round re-instantiation.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
```diff
@@ -703,6 +797,26 @@ def _fetch_mech_requests_by_titles(
...
+        # Flag-on path: pull scored rows since the watermark from
+        # mech-analytics and filter to the requested titles client-side.
+        # This is the per-tick incremental fetch path used by
+        # ``_perform_incremental_update`` — without a flag branch here,
+        # off-chain days would silently attribute zero fees even when
+        # the flag is on for the initial backfill.
+        if is_flag_enabled(self.params):
+            since = None
+            if block_timestamp_gt > 0:
+                since = datetime.fromtimestamp(int(block_timestamp_gt), tz=timezone.utc)
```
OjusWiZard: **[Critical]** Boundary mismatch on the watermark: the endpoint documents `since` as **inclusive** (`mech_analytics_client.py:125,239`), but the subgraph query this branch replaces used `blockTimestamp_gt` (strictly greater), and `block_timestamp_gt` here is `existing_data.last_mech_timestamp` — the timestamp of a row already consumed in a prior round. Passing it straight through re-fetches that row on every incremental tick; if its title recurs among the round's `new_question_titles`, `_match_mech_requests_to_days` re-counts it into `fees_by_day`/`unplaced_count`. And because `settled_mech_requests_count = max(prev, new)` (line ~2103) is a monotonic floor, the overcount **never self-corrects**.

The `until` side of exactly this mismatch was fixed with `ts + 1` in the per-position helpers — the same adjustment is needed here: `datetime.fromtimestamp(int(block_timestamp_gt) + 1, tz=timezone.utc)`.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,521 @@
...
+        return None
+    # ``.get("next_cursor")`` already returns None for an absent key;
+    # ``or None`` additionally coerces a present-but-falsy value so a
+    # schema-drift edge case can't leak an empty cursor into the next
+    # request's params.
+    next_cursor = payload.get("next_cursor") or None
+    return {"rows": rows, "next_cursor": next_cursor}
+
+
+def parse_requester_payload(
```
OjusWiZard, line 224: **[Critical]** This validates one level too shallow: it checks `windows` is a dict but not that `windows["all"]` exists or that `n_mech_requests` exists inside it. A schema drift in a healthy 200 response (e.g. mech-analytics renames `n_mech_requests`) passes validation, and the consumer chain in `_get_total_mech_requests` — `(usage.get("windows") or {}).get("all") or {}` then `int(all_window.get("n_mech_requests") or 0)` — silently produces `0`, which is **cached** (`self._total_mech_requests = 0`) and short-circuits `_calculate_settled_mech_requests` to `0`. Net effect: `all_mech_costs = 0` and silently inflated ROI, with zero log output — indistinguishable from the genuine unknown-Safe zero payload this function's docstring is careful to distinguish.

Suggest validating `windows["all"]` is a dict containing `n_mech_requests`, returning `None` (logged) otherwise, so schema drift lands in the existing failure lane instead of the success lane.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
```diff
@@ -303,17 +369,40 @@ def _get_open_market_requests(
...
+            rows = yield from self._page_mech_analytics_scored_rows(
+                requester=agent_safe_address,
+                since=since,
+            )
+            if rows is None:
+                self.context.logger.warning(
+                    "mech-analytics scored-rows fetch failed; skipping "
+                    "open_market_requests for this round"
+                )
+                return 0
```
OjusWiZard: **[Important]** This is the one function that didn't get the `Optional` treatment, and it undermines its siblings: on a scored-rows fetch failure this returns a bare `0`, so when `_get_total_mech_requests` succeeds (say 500) in the same round, `_calculate_settled_mech_requests` computes `500 − 0 = 500` — every open request counted as settled. That value is a concrete `int`, so it passes the `if settled is not None:` guard at line ~2203 and gets baked into the round's `settled_mech_costs` (overstated, understating ROI this time). Three independent review passes converged on this one.

Suggest returning `Optional[int]` with `None` here, exactly mirroring `_get_total_mech_requests`, and letting `_calculate_settled_mech_requests` propagate it.

packages/valory/skills/agent_performance_summary_abci/models.py
```diff
@@ -334,6 +334,23 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        # than silently returning zero rows and inflating ROI). Both default
+        # to the safe pre-migration behaviour: flag off, subgraph read
+        # unchanged. See docs (consumer migration §7 in mech-analytics repo).
+        self.mech_analytics_url: str = self._ensure("mech_analytics_url", kwargs, str)
+        self.use_mech_analytics: bool = self._ensure("use_mech_analytics", kwargs, bool)
+        # Enforce the pairing at startup — fail loudly rather than
+        # silently no-op'ing back to the subgraph path when an operator
+        # enables the flag but forgets the URL.
+        if self.use_mech_analytics and not self.mech_analytics_url:
+            raise ValueError(
```
OjusWiZard, line 350: **[Important]** This `raise` is never executed by any test — the coverage run shows line 350 as Missing, and the flag-on fixtures in `test_requests.py` construct a bare `SimpleNamespace`, bypassing `__init__` validation entirely. If a future refactor inverts the guard or moves it after an early return, a misconfigured deployment (flag on, URL unset) silently falls through again — the exact regression this line exists to prevent. Two tests needed: flag-on + empty URL raises through the real `AgentPerformanceSummaryParams.__init__`; flag-off + empty URL doesn't.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,521 @@
...
+        adapted.append(
+            {
+                "blockTimestamp": block_timestamp,
+                "parsedRequest": parsed_request,
+            }
+        )
+    return adapted
+
+
+def fetch_requester_usage(
```
OjusWiZard: **[Advisory]** Dead code: this sync `fetch_requester_usage` has no production caller — `_get_total_mech_requests` uses the async inline path (`build_requester_url` + `get_http_response` + `parse_requester_payload`) instead, and grep confirms the only references are its own 8 tests in `TestFetchRequesterUsage`. Either delete it (with its test class), or if it's kept deliberately as API surface, its untested `except ValueError` block (lines ~437-442) needs the malformed-JSON test its sync sibling `fetch_scored_rows` already has.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
```diff
@@ -0,0 +1,521 @@
...
+        if logger is not None:
+            logger.error(
+                "mech-analytics /requester: unexpected response shape "
+                "(windows is not a dict)"
+            )
+        return None
+    return payload
+
+
+def fetch_scored_rows(
```
OjusWiZard, line 293: **[Advisory]** This sync loop hand-rolls its URL/params construction and the `rows`/`next_cursor` extraction instead of calling `build_scored_rows_url` and `parse_scored_rows_page` — the shared helpers that sit in this same file and that the async path already uses. `build_scored_rows_url`'s docstring even claims the sync path uses it, which is currently false. The two implementations agree today, but nothing enforces that: a future edit to a query-param name or the cursor normalisation made in one place silently desyncs the per-position sync lookups from the bulk async ones. Routing this loop through the shared helpers is behavior-preserving (identical param order and checks) — the only fallout is mechanical: the `TestFetchScoredRows` assertions on `call_args.kwargs["params"]` move to asserting on the built URL.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
```diff
@@ -199,6 +208,74 @@ def _fetch_from_subgraph(
 
         return result
 
+    def _page_mech_analytics_scored_rows(
+        self,
+        requester: str,
+        since: Optional[Any] = None,
```
OjusWiZard: **[Advisory]** `since`/`until: Optional[Any]` — but these forward into `build_scored_rows_url`, which calls `.isoformat()` unconditionally on non-None values. `Any` means mypy can't catch a caller passing a raw epoch int (the natural pattern here — the sibling `_fetch_mech_sender` takes `timestamp_gt: int`), which would crash with `AttributeError` mid-generator instead of the graceful `None` every other failure path returns. All current call sites pass `datetime`, so it's latent — but `Optional[datetime]` makes the type system do the work.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
```diff
@@ -720,6 +758,53 @@ def _fetch_prediction_response_from_mech(
...
+                base_url=params.mech_analytics_url,
+                requester=sender_address,
+                chain_id=chain_id_for_platform(params.is_running_on_polymarket),
+                since=since,
+                until=until,
+                logger=self.logger,
+            )
+            if rows is None:
+                return None
+            matched = find_latest_row_for_title(rows, question_title)
```
OjusWiZard, line 791: **[Advisory]** Two small things on this lookup: (1) when the fetch *succeeds* but nothing matches inside the window, the return is `None` with no log at all — indistinguishable from "this mech request never existed." A mech request older than the 30-day `PER_POSITION_LOOKUP_WINDOW_DAYS` (e.g. after extended downtime) silently blanks tool/confidence for a whole batch of positions with zero ops visibility; a DEBUG/INFO line here would make a fleet-wide window-too-narrow pattern diagnosable. (2) The `until = fromtimestamp(ts + 1)` + `since = until − 30d` construction is duplicated verbatim at all four per-position sites — a tiny `per_position_lookup_window(ts)` helper next to the constant would keep the boundary semantics in one place (narrower than the deferred flag-block hoist, so it doesn't conflict with that decision).

packages/valory/skills/agent_performance_summary_abci/tests/graph_tooling/test_mech_analytics_client.py
```diff
@@ -0,0 +1,648 @@
...
+        """Returns none when title empty."""
+        rows = [{"question_title": "Q", "requested_at": "2026-07-10T00:00:00Z"}]
+        assert find_latest_row_for_title(rows, "") is None
+
+    def test_returns_none_when_rows_empty(self) -> None:
+        """Returns none when rows empty."""
+        assert find_latest_row_for_title([], "Q") is None
+
+    def test_picks_latest_matching_row_by_requested_at(self) -> None:
+        """Picks lamatching row by requested at."""
```
OjusWiZard: **[Advisory]** Corrupted docstring — `"""Picks lamatching row by requested at."""` (typo) followed by a stray second string literal `"""Picks la."""` that's just an unreachable expression statement. Looks like a botched edit; should read `"""Picks latest matching row by requested_at."""` on one line.

OjusWiZard (commented):
Round-2 review of the fix commits (same six-perspective sweep, briefed on round 1 so only new findings are reported). The fixes are real improvements — but one of them doesn't reach the headline metric, and the coverage gate regressed.

**[Critical] Coverage gate now fails at 95.30% (was 96.20% in round 1; CI requires 100%).** The suite passes (1162 tests) but the uncovered lines are precisely the fix code: the entire flag-on bodies of `_get_total_mech_requests` (behaviours.py 310-345) and `_get_open_market_requests` (379-393); every `None` hop of the new Optional chain (`behaviours.py` 460, 1091, 1136, 2203→2206); the `models.py:350` `ValueError` (never raised in any test — `test_requests.py` builds a bare `SimpleNamespace` that bypasses `__init__` validation entirely); and the async MAX_PAGES exhaustion path (`requests.py` 273-277 — only the sync twin has that test). Round-1 gaps still open: Polymarket `fetch_mech_tool_for_question` failure paths (615/618), `_fetch_prediction_response_from_mech` failure paths in both helpers (Omen 788/791, Polymarket 699/702), sync `fetch_requester_usage` malformed-JSON (437-442).

An end-to-end test of the Optional chain (fetch failure → `calculate_roi` output) would have caught the inline `calculate_roi` finding below — the test gap and the bug are the same hole.

What's solid this round: the async pagination tests (non-200 mid-page, JSON parse failure, shape drift, multi-page ordering) are genuinely behavioral; the `None`→empty-metrics path is correctly protected by the pre-existing all-NA preservation in `_save_agent_performance_summary`; and the startup `ValueError` surfaces loudly through the AEA instantiation wrapper rather than being swallowed.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
DIvyaNautiyal07: You're right, this was a silent overcount waiting to fire. The subgraph query used `blockTimestamp_gt` (strictly greater) but mech-analytics' `since` is inclusive, and since `settled_mech_requests_count = max(prev, new)` is monotonic the overcount would never self-correct once triggered.

Fixed in bab541ad6 by offsetting the watermark by +1s on the flag-on branch, mirroring the `ts + 1` treatment already used on `until` in the per-position helpers. Pinned by two tests: one asserts the +1 offset made it onto the `since=` param when `block_timestamp_gt > 0`, the other asserts the backfill entry point (`block_timestamp_gt=0`) still omits `since` entirely, so we don't regress in the opposite direction.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Confirmed the whole chain, and confirmed my previous "preserve cached count" comment was wrong for exactly the reason you laid out: the behaviour is re-instantiated per round, so there is no prior-round `self._settled_mech_requests_count` to fall back on. On failure the count stayed at the init `0`, `calculate_roi` produced a real undercounted float, the NA-preservation fallback never fired, and the persisted "Total ROI" got overwritten.

Fixed in 6d64bfe82 by taking your first suggestion (track as `Optional[int]`, propagate). Concretely:

- `self._settled_mech_requests_count: Optional[int] = None` in `__init__`
- `calculate_roi` returns `(None, None)` up front when the count is None
- `_calculate_performance_metrics` mirrors the same guard (returns None), matches the existing `total_mech_requests is None → return None` pattern one branch down
- `_fetch_agent_performance_summary` still uses `if settled is not None: self._settled_mech_requests_count = settled` to skip overwriting, but the comment now names what actually happens (leave None so downstream propagates → NA)

Pinned by three tests: `calculate_roi` → `(None, None)` on unknown count, `_calculate_performance_metrics` → `None` on unknown count, and a sanity test that a real `0` settled-count still produces a real float ROI (so the guard only trips on `None`, not zero). With the count propagated as None, `final_roi=None` → metric value = `NA` → NA-preservation in `_save_agent_performance_summary` keeps the last persisted good value.

bennyjo (dismissed):
Approving. Checked the flag-off invariant line by line rather than taking it on trust: the subgraph branch of `_get_total_mech_requests` is unchanged and still returns 0 on failure, so the `Optional[int]` widening is unreachable with the flag off; splitting `if not total_mech_requests` into an `is None` and an `== 0` branch preserves the old falsy semantics for a genuine zero; and although `_settled_mech_requests_count` now initialises to `None`, every read site sits downstream of an assignment that always fires when the flag is off. Both of my earlier points are fixed in the code too — the per-position lookups bound the window with `since`, and `_iso_to_unix_seconds` warns instead of silently zeroing a row. Good catch on `until = ts + 1` for the exclusive boundary; that one was yours.

One thing to note rather than block on: `behaviours.py:1638` downgrades "No mech requests found for agent" from WARNING to INFO, and that line is on the shared path, so it quiets an existing signal with the flag off too.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Confirmed the silent-zero path end-to-end: `parse_requester_payload` accepting the shallow `windows` dict, then the consumer's permissive `.get("all") or {}` + `int(all_window.get("n_mech_requests") or 0)` chain, then the `self._total_mech_requests = 0` cache, then the settled-count short-circuit — exactly your trace, no log emitted anywhere on the drift path.

Fixed in 86cff8f1e by deepening the validation in `parse_requester_payload`: also checks `windows["all"]` is a dict and `n_mech_requests` is present and numeric (with an explicit bool reject so a drift flipping to `True/False` doesn't cache as `1/0`). Any of the checks failing routes to the existing None-with-log lane, so schema drift now lands in the failure lane consistently.

Tests pin each new drift shape (missing `all`, non-dict `all`, missing `n_mech_requests`, wrong-type including str / bool / dict) plus the sanity partner that a genuine `0` count still passes.

packages/valory/skills/agent_performance_summary_abci/behaviours.py
DIvyaNautiyal07: Right, `_get_open_market_requests` was the last sibling that wasn't propagating None — mirror of the total-fetch silent-zero one branch up, just in the opposite direction (returning `0` here makes `settled = total - 0 = total`, so every open request counts as settled and over-inflates `settled_mech_costs`, understating ROI instead of inflating it).

Fixed in 35fbf1d11: `_get_open_market_requests` now returns `Optional[int]`, returns `None` on flag-on fetch failure. `_calculate_settled_mech_requests` short-circuits to `None` if either count is unknown. Also updated the incremental-scan `max_settled` bound to `None` when either count is unknown, rather than treating an unknown open as `0` and loosening the ceiling to `total`.

Tests pin the `None` return on flag-on failure and that `_calculate_settled_mech_requests` returns `None` when open is unknown.

packages/valory/skills/agent_performance_summary_abci/models.py
DIvyaNautiyal07: Fair, coverage confirmed the raise was Missing. Fixed in 7bad1cbc5 with three tests through the real `AgentPerformanceSummaryParams.__init__`:

- `test_flag_on_with_empty_url_raises_at_init` — pins the raise.
- `test_flag_off_with_empty_url_does_not_raise` — sanity partner proving the guard only trips on the specific misconfiguration, not on any Safe that hasn't opted in.
- `test_flag_on_with_populated_url_does_not_raise` — the intended production configuration.

So a future refactor that inverts the guard, moves it after an early return, or drops the raise entirely, fails a test instead of shipping a misconfigured deployment.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Deleted in a6bc75089 (along with its 8 tests). Grep confirmed no production caller — `_get_total_mech_requests` uses the async inline path (`build_requester_url` + `get_http_response` + `parse_requester_payload`) directly and there were no other importers.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/mech_analytics_client.py
DIvyaNautiyal07: Right, the two paths agreed today but nothing enforced it. Refactored in a6bc75089: `fetch_scored_rows` now calls `build_scored_rows_url` and `parse_scored_rows_page` (same helpers the async path already uses). Behaviour-preserving — same param order, same limit, same cursor normalisation, same `None`-on-shape-drift.

Mechanical test fallout was as you called out — the `TestFetchScoredRows` assertions moved from `kwargs["params"]` to URL substring checks, since the URL is now built up front and passed to `requests.get` without a `params=` kwarg.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
DIvyaNautiyal07: Fixed in a6bc75089 — `_page_mech_analytics_scored_rows` now types `since`/`until` as `Optional[datetime]`. All current call sites already pass `datetime`, but mypy now catches the raw-epoch-int mistake at the call site instead of letting it crash mid-generator on `.isoformat()`.

packages/valory/skills/agent_performance_summary_abci/graph_tooling/predictions_helper.py
DIvyaNautiyal07: Both done in a6bc75089:

(1) Empty match on the flag-on branches now logs at INFO with the sender + title + window size, so a fleet-wide "window too narrow" pattern is diagnosable in ops output. Added on all four per-position sites (Omen path 2/3, Polymarket path 2/3).

(2) Extracted `per_position_lookup_window(bet_timestamp) -> (since, until)` colocated with `PER_POSITION_LOOKUP_WINDOW_DAYS` in `mech_analytics_client.py`. Every call site now reads `since, until = per_position_lookup_window(ts)`. The `+1s` inclusive-boundary adjustment and the `-30d` capping decision are now in one place, so a future tweak lands everywhere at once and stops duplicating the boundary comments across four sites. Kept the helper narrow enough to sit alongside the constant without conflicting with the flag-block hoist you called out earlier as deliberately deferred.

packages/valory/skills/agent_performance_summary_abci/tests/graph_tooling/test_mech_analytics_client.py
DIvyaNautiyal07: Fixed in a6bc75089 — botched edit as you suspected. Now reads `\"\"\"Picks latest matching row by requested_at.\"\"\"` on a single line, stray second string literal removed.

OjusWiZard (approved):
Approving after three review rounds (36 findings raised, all critical/important code issues fixed and verified at head a6bc75089: Optional propagation through the full ROI chain incl. round re-instantiation, both watermark boundary fixes, deep /requester payload validation, async pagination on the FSM path, startup flag/URL validation, dead code removal).

Remaining known follow-ups, none blocking approval from my side:
- Coverage gate (95.79% vs the 100% CI requirement) — the flag-on body of `_get_total_mech_requests`, the async MAX_PAGES path, the Polymarket/Omen failure-path tests, and an end-to-end fetch-failure→NA-preservation test. CI will enforce this regardless.
- Two small validator hardenings in `parse_requester_payload`: reject NaN/Infinity (crashes the unguarded `int()` at behaviours.py:352) and negative `n_mech_requests` (silently inflates Total ROI).
- `_fetch_agent_performance_data`: compute `stats` unconditionally so a mech-analytics blip doesn't null/freeze `predictions_made`/`prediction_accuracy`, which don't depend on it.

bennyjo (approved):

PR #1014, opened 2026-07-24
chore: drop unused olas_mech_subgraph override from polymarket_trader

Tanya-atatakai:
## What

Removes the `olas_mech_subgraph` model override from the Polymarket (Polystrat) service, which was the **only remaining reference** in the repo to the deprecated `https://api.subgraph.autonolas.tech/api/proxy/mech` subgraph. Also bumps the `polymarket_trader` service hash in `packages.json` accordingly.

## Why it's safe (dead config)

`olas_mech_subgraph` is only read on the **non-Polymarket** code path:

```python
# packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py
if self.params.is_running_on_polymarket:
    subgraph = self.context.polygon_mech_subgraph   # marketplace-polygon
else:
    subgraph = self.context.olas_mech_subgraph      # <- never reached on Polystrat
```

Polystrat runs with `IS_RUNNING_ON_POLYMARKET=true` (default in `polymarket_trader/service.yaml`), so it **always** uses `polygon_mech_subgraph` and never touches `olas_mech_subgraph`. The override was stale leftover config pointing at the old `/api/proxy/mech` endpoint.

## Why drop rather than re-point

- Matches the existing partial-override pattern in this service — `polygon_mech_subgraph` (the one Polystrat actually uses) has no override block either; it inherits the skill default.
- The model still exists at the skill level, so if it were ever read (e.g. `IS_RUNNING_ON_POLYMARKET=false`), it falls back to the skill default `marketplace-gnosis`, **not** the deleted `/mech` endpoint.
- Re-pointing would leave a dead block that misleadingly implies Polystrat queries a Gnosis mech subgraph.

This unblocks deleting the `/api/proxy/mech` subgraph with zero runtime impact.

## Not affected

- **Omenstrat** (`trader_pearl`): its `olas_mech_subgraph` already points at `marketplace-gnosis` (a different subgraph) and is untouched.
- Skill-level tests mock `context.olas_mech_subgraph` directly and don't read `service.yaml`.

## Verification

- `grep` confirms `/api/proxy/mech` is now referenced nowhere in the repo.
- Service package hash recomputed with the framework's `IPFSHashOnly.hash_directory` (method validated against an unchanged package first) and updated in `packages.json`; on-disk dir now matches the recorded hash.
- `service.yaml` still parses (multi-doc).

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -19,7 +19,7 @@
         "skill/valory/chatui_abci/0.1.0": "bafybeig4qckyl7exilbd4egapsttvfrecfu4buywx632gixuvkwcjqqssu",
         "agent/valory/trader/0.1.0": "bafybeihrtti72gii23aufe6zeezzydyqusvz4bt3wn6rcvyk6lu5it6g2i",
         "service/valory/trader_pearl/0.1.0": "bafybeico5agjhkwljmiro4563n3rclmai2v7i7oeluxi52hncnpx4r7tve",
-        "service/valory/polymarket_trader/0.1.0": "bafybeiehvfewnlmmt5cnythj3dzhmfu6aj3swjqys36cci34clyajd3edi"
+        "service/valory/polymarket_trader/0.1.0": "bafybeicu443g4b36shsznd4my7yfmfgybp4j3ujnak3gha4mkqrl6rbiym"
     },
     "third_party": {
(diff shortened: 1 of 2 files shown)


OjusWiZard (approved):
Reviewed via multi-perspective review: all four safety claims verified against the code (dead config on Polystrat, safe fallback to marketplace-gnosis at agent/skill level, no remaining /api/proxy/mech references, package hash verifies cleanly). No blocking findings.

jmoreira-valory (approved):

Issue #970, opened 2026-05-15
Polymarket redeem: data-API failure crashes the whole agent (unhandled error response)


LOCKhart07:
## Summary

A transient Polymarket data-API failure during a redeem round crashes the **entire agent** with an unhandled `AttributeError`, instead of being retried or no-op'd. This is a long-standing latent bug (predates all recent Polymarket redeem work), only triggered when the `data-api.polymarket.com/positions` call fails.

## Observed crash

From a Pearl `agent_runner.log` (Polystrat, agent `sc-c427c8fc...`):

```
[ERROR] Error fetching positions: HTTPSConnectionPool(host='data-api.polymarket.com', port=443):
        Max retries exceeded with url: /positions?...&redeemable=True
        (Caused by ConnectTimeoutError(... connect timeout=10))
[INFO]  Received Srr message: ... error=True, payload={"error": "Error fetching positions: ... timed out ..."}
[INFO]  Fetched 1 redeemable positions
...
File ".../decision_maker_abci/behaviours/polymarket_reedem.py", line 77, in _update_policy_for_redeemable_positions
    condition_id = position.get("conditionId")
AttributeError: 'str' object has no attribute 'get'

Error: AEA was terminated cause exception `... TraderConsensusBehaviour ...`
```

The agent process is terminated by the framework (uncaught `AEAActException`).

## Root cause

Two layered defects:

1. **`send_polymarket_connection_request`** (`packages/valory/skills/market_manager_abci/behaviours/base.py:100`) returns `json.loads(response.payload)` and **never checks `response.error`**. When the connection fails it returns the dict `{"error": "...timed out..."}` instead of a list of positions. This affects *every* caller, not just redeem.

2. **`PolymarketRedeemBehaviour`** (`packages/valory/skills/decision_maker_abci/behaviours/polymarket_reedem.py`) never validates the response shape:
   - `len({"error": ...})` is `1`, so it logs the misleading `Fetched 1 redeemable positions`.
   - `_update_policy_for_redeemable_positions` (line 76–77) does `for position in redeemable_positions:` — iterating a dict yields its **keys**, so `position == "error"` (a `str`), and `position.get("conditionId")` raises `AttributeError`.
   - The same crash exists in the `_prepare_redeem_tx` and `_redeem_via_builder` loops (they also call `position.get(...)`).

A recoverable network blip becomes a fatal agent crash.

## Was this always a bug?

Yes — not a regression:

[...]

PR #1016, opened 2026-07-29
fix(polymarket_client): defensive type guard + MAX_PAGES cap on keyset cursor loop (closes #936, #970)

### LINKED ISSUE #936: Two robustness-vs-upstream concerns worth a follow-up PR — same shape (less defensive against the server misbehaving than the old offset-based loop):
Two robustness-vs-upstream concerns worth a follow-up PR — same shape (less defensive against the server misbehaving than the old offset-based loop):

1. **`response.get("events")` here can `AttributeError` if `_request_with_retries` returns a non-dict.** That helper is typed `Tuple[Any, str]` and just returns `response.json()` — a JSON `null`, a stray list, or a string from a misbehaving edge proxy would crash here and bubble up to `fetch_markets`'s blanket `except Exception`, dropping the whole category. The old code's `if not events_data: break` degraded gracefully on the same inputs. Cheap fix: `if not isinstance(response, dict): break` (or `(response or {}).get("events")`).

2. **No client-side termination bound on the cursor loop** (visible at L609–611). The old code had `len(events_data) < EVENTS_LIMIT: break` as a natural ceiling; the new code relies entirely on the server eventually returning a falsy `next_cursor`. If the upstream ever returns the same cursor in a row (server bug, cache poisoning, partial outage), `all_markets` grows unbounded until OOM. Worth a `for _ in range(MAX_PAGES)` outer with a logged warning on cap — the agent runs unattended.

Both are follow-up-eligible — flagging because the migration changes the failure mode here from "stop paginating gracefully" to "raise / loop forever."

_Originally posted by @bennyjo in https://github.com/valory-xyz/trader/pull/931#discussion_r3172789033_

lzhao8956-glitch:
## Summary

Closes #936

Two robustness-vs-upstream concerns in `connection.py:fetch_markets` (both flagged by @bennyjo on PR #931 review):

1. `response.get("events")` can `AttributeError` if `_request_with_retries` returns a non-dict payload (JSON null, stray list, or string from a misbehaving edge proxy). Crashes here and bubbles up to `fetch_markets` blanket `except Exception`, dropping the whole category.

2. No client-side termination bound on the cursor loop. The pre-#931 `len(events_data) < EVENTS_LIMIT: break` was a natural ceiling; the post-#931 code relies entirely on the server eventually returning a falsy `next_cursor`. If the upstream ever returns the same cursor in a row, `all_markets` grows unbounded until OOM.

## Changes

- Add `MAX_PAGES = 100` constant next to `EVENTS_LIMIT` (>=10x current single-call ceiling of ~10 events on Polystrat).
- Replace `while True:` with `for _ in range(MAX_PAGES):` outer loop.
- Add `isinstance(response, dict)` guard before `.get()` calls.
- Add `for-else` warning on cap-hit so the next refresh can re-attempt.

## Backwards compatibility

- All existing call sites unchanged.
- The cap only triggers on the pathological server-bug path. On Polystrat today the loop terminates after 1-2 pages.
- New warning log line is informational only.

## Out of scope (deferred)

- Per-call circuit breaker for sustained subgraph outage (issue #938 P3) - separate PR.
- `_request_with_retries` typing-hygiene fix - separate PR.

### DIFF
--- a/packages/valory/connections/polymarket_client/connection.py
+++ b/packages/valory/connections/polymarket_client/connection.py
@@ -89,7 +89,7 @@ POLYMARKET_CATEGORY_TAGS = [
     "international",
 ]
 MARKETS_LIMIT = 300
-EVENTS_LIMIT = 200
+EVENTS_LIMIT = 200`nMAX_PAGES = 100  # Safety cap on the keyset cursor loop (issue #936)
 MARKETS_TIME_WINDOW_DAYS = 4
 API_REQUEST_TIMEOUT = 10
 MAX_API_RETRIES = 3
@@ -858,7 +858,7 @@ class PolymarketClientConnection(BaseSyncConnection):
         after_cursor: Optional[str] = None
         all_markets: list = []
 
-        while True:
+        for _ in range(MAX_PAGES):
             params: Dict[str, Any] = {
                 "tag_slug": tag_slug,
                 "end_date_max": end_date_max,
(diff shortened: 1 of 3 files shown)


lzhao8956-glitch (commented):
#936,改成 #970

PR #1018, opened 2026-08-04
fix(funding): raise omenstrat Agent Safe topup to 10 xDAI (OPE-1853)

atepem:
## Problem

Omenstrat operators see **Low Agent Wallet Balance** alerts re-fire within minutes of funding (ZD#1181). Root cause is not a routing bug: the omenstrat staking contracts now require **16 mech req/day instead of 8**, so the agent trades — and spends — at double the previous rate, while redemptions still land days later. The `funds_manager` topup/threshold were still calibrated for 8 req/day.

Log evidence: Agent Safes at 0–0.09 xDAI (depleted) while Agent Signers sat at 0.73–1.26 xDAI against a 0.21 threshold (healthy).

## Change

`packages/valory/services/trader_pearl/service.yaml` — Gnosis `safe` native `topup` **5 → 10 xDAI**.

- Threshold stays at **1 xDAI**, so runway between top-up and next alert goes **4 → 9 xDAI**
- Agent Signer values unchanged — signer balances were healthy
- `polymarket_trader` (Polystrat) untouched — out of scope

## Verification

Validated against the shipped `funds_manager` models and open-aea's env-var machinery:

- `FundRequirements.from_dict` accepts the payload (both `agent` and `safe` keys required by the validator)
- `apply_env_variables` resolves the `${FUND_REQUIREMENTS:dict:...}` default to `topup=10000000000000000000`
- Deficit math (`behaviours.py`): empty safe → requests 10 xDAI; balance ≥ 1.0 xDAI → 0

## Still required before merge

- [ ] `make hashes` in this repo's pinned `open-autonomy==0.21.19` env — `check-hash` / `check-packages` will fail until `packages/packages.json` is regenerated
- [ ] IPFS publish + release, so the new hash can be picked up by the Pearl template

Related: valory-xyz/olas-operate-app (template bump), valory-xyz/olas-operate-middleware#471 (tests)

Relates to: OPE-1853


atepem:
### Note on the `make hashes` step

I attempted the hash regeneration and **backed it out** — it cannot be done correctly on Windows.

`uv sync --all-groups` reproduces the pinned toolchain exactly (`autonomy 0.21.19`, `aea 2.2.1`), and after syncing third-party packages at their committed hashes, `autonomy packages lock` runs and produces a single expected change:

```
service/valory/trader_pearl/0.1.0: bafybeifkx4fk4ls7dagl6l3mzmsx3bp6auheapsjfuk6cbhja2lah7qdlm
                                -> bafybeif3sanp4u3tpbmqcf26ytx522pamj5eoptid2qyvpjcbrzxkhsueq
```

**That hash is wrong.** `packages lock` rewrites the component yamls through Python text-mode writes, which emit CRLF on Windows, so it hashes CRLF content. This repo's own `.gitattributes` documents the hazard:

> `# Force LF line endings for all text files to ensure consistent IPFS hashes across platforms`

After the lock run, 21 yamls showed as modified with **CRLF-only** diffs (no content change), confirming the rewrite. A hash computed over CRLF will not match what Linux CI computes over LF.

**Action for whoever picks this up:** run `make hashes` on Linux or macOS and commit the resulting `packages.json` here. Two useful notes:

1. Don't run `make sync-packages` first — it resolves `MECH_INTERACT_VERSION` to the *latest* tag (`git ls-remote ... | tail -n1`), which no longer matches the committed `packages.json` and rewrites 14 unrelated third-party contract hashes. Plain `autonomy packages sync` fetches at the pinned hashes and leaves `packages.json` untouched.
2. Only `packages.json` should change. Everything else in the lock output was line-ending noise.

The `service.yaml` change in this PR is unaffected and verified independently — see the PR description.

OjusWiZard (dismissed):

Tanya-atatakai (dismissed):

packages/valory/services/trader_pearl/service.yaml
```diff
@@ -388,7 +388,7 @@ type: skill
 models:
   params:
     args:
-      fund_requirements: ${FUND_REQUIREMENTS:dict:{"gnosis":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":2000000000000000000,"threshold":210000000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":5000000000000000000,"threshold":1000000000000000000},"0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d":{"topup":0,"threshold":0}}}}}
+      fund_requirements: ${FUND_REQUIREMENTS:dict:{"gnosis":{"agent":{"0x0000000000000000000000000000000000000000":{"topup":2000000000000000000,"threshold":210000000000000000}},"safe":{"0x0000000000000000000000000000000000000000":{"topup":10000000000000000000,"threshold":1000000000000000000},"0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d":{"topup":0,"threshold":0}}}}}
```
Tanya-atatakai, line 391: are we sure this should be the same as initial requirements? those are for low funding alert, and e.g. for connect we say "Thresholds are 1/5 of the initial requirement" cc @OjusWiZard

atepem: for topup, it's doubled for 2x the activitydone on the new staking contracts. Setup should be correct :)

OjusWiZard: it's the topup, not threshold. So basically the config is - if balance gets below 1 xDAI, then ask to refill it back to 10 xDAI.

atepem:
### Update: hashes regenerated ✅

Superseding my earlier comment — the hash **is** now done. I regenerated it in a real Linux environment (WSL Ubuntu) rather than on Windows, using the branch's pinned toolchain (`open-autonomy 0.21.26`, `open-aea 2.2.9` — the pin moved with the `main` merge).

```
service/valory/trader_pearl/0.1.0
  bafybeiatfdabhpbzpnwrpbkufa4mhto5t6pqdwfuvzyypez7t47vedvcdq
→ bafybeiaxztwn6nzpqituedp7rrogau5egvlxzf4na4zu4fmqjdugag6s6m
```

Under Linux the lock touched **only** `packages/packages.json` — none of the CRLF noise Windows produced, confirming line endings were the whole problem.

### Remaining CI failure is not from this PR

`linter_checks` still fails, but at its **first** step (`tomte tox -p -e safety -e bandit`), so it never reaches `check-hash`. The error:

```
ImportError: Blocked import of regex from current working directory for security reasons.
Use '-P' or set PYTHONSAFEPATH to prevent Python from searching the current working directory.
```

Cause is in tomte's canonical tox template, not this repo:

```ini
[testenv]
setenv =
    PYTHONPATH={env:PWD:%CD%}
```

That puts the repo root on `sys.path` for every env including `safety`, which trips nltk's `inisec.py` guard, and `safety` dies on import before scanning anything. Two notes:

- `passenv` does not include `PYTHONSAFEPATH`, so the fix the error suggests cannot reach the env from the workflow.
- **Bumping tomte does not help** — I diffed `v0.7.0` vs `v0.7.1`; both `[testenv:safety]` and the `PYTHONPATH` line are byte-identical.

This blocks every trader PR, not just this one, and the check currently provides no security value (it crashes before scanning). Fixing it needs either an upstream tomte change or `continue-on-error: true` on the Security checks step — a repo-wide call I've left to the maintainers rather than making inside a funding PR.

Tanya-atatakai (approved):

OjusWiZard (approved):

PR #1020, opened 2026-08-10
chore(trader): bump mech_interact_abci to v0.32.7; hoist use_offchain and offchain_deposit_target_calls to top-level params

DIvyaNautiyal07:
## Summary

Companion to [valory-xyz/mech-interact#113](https://github.com/valory-xyz/mech-interact/pull/113) ([release v0.32.7](https://github.com/valory-xyz/mech-interact/releases/tag/v0.32.7)).

Two knobs — `use_offchain` and `offchain_deposit_target_calls` — moved out of the nested `MechMarketplaceConfig` dataclass into top-level `MechParams` attributes upstream. This PR wires the change through every trader-side overlay (yamls) and the one Python consumer that read the old nested shape.

## Changes

- **`packages/packages.json`** — `mech_interact_abci` hash bumped to `bafybeic5xfxdod5l56o7aojdamojnvi562nql2mlah73bf47d4guz4xnwi` (v0.32.7). Cascade re-lock through trader_abci, decision_maker_abci, check_stop_trading_abci, chatui_abci, tx_settlement_multiplexer_abci, agent_performance_summary_abci, trader agent, both services.
- **Yaml overlays** (5 files) — hoist to top-level Params args:
  - `agents/trader/aea-config.yaml` — add `use_offchain: ${USE_OFFCHAIN:bool:false}` + `offchain_deposit_target_calls: ${OFFCHAIN_DEPOSIT_TARGET_CALLS:int:10}` next to the `mech_marketplace_config` line. Inline JSON default didn't carry either key already, so no removal there.
  - `services/trader_pearl/service.yaml` and `services/polymarket_trader/service.yaml` — strip both keys from the `${MECH_MARKETPLACE_CONFIG:dict:{…}}` JSON default and add top-level env-var forms next to it.
  - `skills/trader_abci/skill.yaml` — move `use_offchain: false` out of the nested `mech_marketplace_config` block to top-level; `offchain_deposit_target_calls: 10` also top-level.
  - `skills/decision_maker_abci/skill.yaml` — add both as top-level args (didn't have either before).
- **Python consumer swap** — `skills/agent_performance_summary_abci/behaviours.py` (`_fetch_offchain_prepaid_wei` early guard): `self.params.mech_marketplace_config.use_offchain` → `self.params.use_offchain`. Only 1 consumer in this repo (verified via grep).
- **Tests** — `agent_performance_summary_abci/tests/test_behaviours.py`: 14 sites doing `params.mech_marketplace_config = _make_marketplace_config(use_offchain=X)` also get `params.use_offchain = X` on the next line to mirror the runtime shape.

## Operator migration note

**Every deploy-time `MECH_MARKETPLACE_CONFIG` value** (env vars, Propel envs, Pearl middleware, `.env` files, Docker Compose files) **MUST drop the two nested keys and set the replacements at top level:**

```diff
-MECH_MARKETPLACE_CONFIG='{"...","use_offchain":false,"offchain_deposit_target_calls":10,...}'
+MECH_MARKETPLACE_CONFIG='{"..."}'  # drop both keys from the JSON
+USE_OFFCHAIN=false
+OFFCHAIN_DEPOSIT_TARGET_CALLS=10
```

Otherwise `MechParams.__init__` raises a `ValueError` at agent boot with a message naming the replacement env vars (fail-loud migration guard added in v0.32.7).

Pearl-specific: `olas-operate-middleware/operate/services/manage.py:541` already composes `MECH_MARKETPLACE_CONFIG` without either key, so the middleware path is unaffected — but Pearl frontend still needs to set `USE_OFFCHAIN` / `OFFCHAIN_DEPOSIT_TARGET_CALLS` as top-level env vars (proposed FIXED overrides in a follow-up Pearl PR).

## Test plan

- [x] `uv run autonomy packages lock --check` green
- [x] `uv run pytest packages/valory/skills/{agent_performance_summary,mech_interact,trader,decision_maker}_abci/tests/` → **3339 passed**
- [x] `mypy` / `black-check` / `darglint` / `flake8` / `isort-check` all green
- [x] Local live-boot verified: `make run-agent` with `USE_OFFCHAIN=false` + `OFFCHAIN_DEPOSIT_TARGET_CALLS=1` at top level + JSON blob stripped. Agent progressed past `MechParams.__init__` (both `_ensure` calls succeeded, `validate_offchain_params()` ran cleanly), HTTP server bound to `:8716`, `/healthcheck` responding, FSM waiting on Tendermint. No hoist-related error in the boot log.
- [ ] CI green on this PR
- [ ] Migration verified on Pearl preview / staging

## Companion PRs (other consumers)

- `optimus` — needs the same hash bump + yaml hoist (nested `mech_marketplace_config` in `skills/optimus_abci/skill.yaml:350,352`)
- `agents-fun` — needs the same
- `market-resolver` — needs the same PLUS strip inline-JSON in `agents/market_resolver/aea-config.yaml:235`

Each surface's crash mode is now explicit — the fail-loud `ValueError` in v0.32.7 names what to change.

### DIFF
--- a/packages/valory/agents/trader/aea-config.yaml
+++ b/packages/valory/agents/trader/aea-config.yaml
@@ -70,15 +70,15 @@ skills:
 - valory/reset_pause_abci:0.1.0:bafybeifrnmlscfiki7wojs6zasppefupd2wkjyq6orhdllfppjud7isrxi
 - valory/termination_abci:0.1.0:bafybeih6sa2ttcvvjk7rcfn7ure2ws23kcfufssb4qmluzcfvednkfpvoi
 - valory/transaction_settlement_abci:0.1.0:bafybeihizaapype5sz7vlosms3nfo6jzfj4ujycuf5uthviq2otvmw6f6m
-- valory/tx_settlement_multiplexer_abci:0.1.0:bafybeiffci4jqgnmjud3j6m5cam7vvjttt6uugfh57az6a4quiwzmk6llm
+- valory/tx_settlement_multiplexer_abci:0.1.0:bafybeifjxdbup557ynvbjfo6q2zoeupz3rerzf3uwyntiqvhmoqupy7a7a
 - valory/market_manager_abci:0.1.0:bafybeig3ak2eoimxwuupybde4napyi2jlf7pxk35se2zclikete2ni5hti
-- valory/decision_maker_abci:0.1.0:bafybeiefxkbb3qh35gnsoqq65bi7wvp7yfbb7uiwef747oahgtyicisumu
-- valory/trader_abci:0.1.0:bafybeiduf4gkl6f6dvhiwe5shf7vtc6fyb5ypyyamlkq4sqdw7xpjf6t4m
+- valory/decision_maker_abci:0.1.0:bafybeihgq4jjqyjrxyzzzi7hd7r27vf2lj45h5ljnp57kzg6enz3vtyqvy
+- valory/trader_abci:0.1.0:bafybeibf5d2gaacgrxxxwv4himk3jmuhnh7j4ahezjrxklqiewzgglgiq4
 - valory/staking_abci:0.1.0:bafybeia6csjgex5zxdt6vymiaengu75lyzjdgfrnp65ewlghthiniquhje
-- valory/check_stop_trading_abci:0.1.0:bafybeicn32bur353d3msdypfcwq257ucmmg55rzyklxftff2l2uoez62wu
-- valory/mech_interact_abci:0.1.0:bafybeifbkpgivxjexlhrptn2iaju7vyzeusevvucsin6tbjwveniykad2q
-- valory/chatui_abci:0.1.0:bafybeiga534fdtg7544qf3paaxcunauy46hu4xphmo2k2izjmdncw72pmy
-- valory/agent_performance_summary_abci:0.1.0:bafybeibw4cjxuswzkhilfxwqodxuwpnqxq7e5egfmrmufoigjtu4iyma6u
+- valory/check_stop_trading_abci:0.1.0:bafybeierqjfa6yq4vmhwsweght4w5vao2f7d732jczuexa7ovy4grhqmui
+- valory/mech_interact_abci:0.1.0:bafybeic5xfxdod5l56o7aojdamojnvi562nql2mlah73bf47d4guz4xnwi
+- valory/chatui_abci:0.1.0:bafybeicsqursfh4svzecimnfbv77jn4vq3cc42e7bislzxfbqigos4ulgm
+- valory/agent_performance_summary_abci:0.1.0:bafybeiaguxjexjxth4e3zajfxvqjprjhn4r6njgq6ddmhkql3uhqdsttp4
 - valory/funds_manager:0.1.0:bafybeicfw7qjtnlcbw6awsvxdahqu42fxsfsd3wy7mbyhcbls4l5pee3by
 customs:
 - valory/kelly_criterion:0.1.0:bafybeidme2uybzyzagky3jtosheqmpgc6b6w4rum7cdxw2fc35nnpwblfe
@@ -259,6 +259,8 @@ models:
       use_mech_marketplace: ${USE_MECH_MARKETPLACE:bool:false}
       policy_store_update_offset: ${int:259200}
       mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x0000000000000000000000000000000000000000","priority_mech_address":"0x0000000000000000000000000000000000000000","priority_mech_staking_instance_address":"0x0000000000000000000000000000000000000000","priority_mech_service_id":0,"requester_staking_instance_address":"0x0000000000000000000000000000000000000000","response_timeout":300}}
+      use_offchain: ${USE_OFFCHAIN:bool:false}
+      offchain_deposit_target_calls: ${OFFCHAIN_DEPOSIT_TARGET_CALLS:int:10}
       expected_mech_response_time: ${int:300}
       mech_invalid_response: ${str:Invalid Response}
       mech_consecutive_failures_threshold: ${int:2}
--- a/packages/valory/services/polymarket_trader/service.yaml
+++ b/packages/valory/services/polymarket_trader/service.yaml
@@ -8,7 +8,7 @@ license: Apache-2.0
 fingerprint:
   README.md: bafybeichoh5tnk4qctaazcjcrxvpwc5yasodleqoeapnfapwrrdh3tm5gu
 fingerprint_ignore_patterns: []
-agent: valory/trader:0.1.0:bafybeihhfz6we66nkc3xtnozidwjtke2ahzoj6ixetzzj5ztp2gekt6zte
+agent: valory/trader:0.1.0:bafybeib42tl6yicz6k5eaezf2kuulpot2uune3ycm7nm6r573crl4tg2ui
 number_of_agents: 1
 deployment:
   agent:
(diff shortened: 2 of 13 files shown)


OjusWiZard (approved):

packages/valory/services/polymarket_trader/service.yaml
```diff
@@ -137,7 +137,9 @@ models:
...
+      mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x343F2B005cF6D70bA610CD9F1F1927049414B582","priority_mech_address":"0x1DCbC9368327776F8B20Dc041D40b8076C8AC173","priority_mech_staking_instance_address":"0x0000000000000000000000000000000000000000","priority_mech_service_id":25,"requester_staking_instance_address":"0x0000000000000000000000000000000000000000","response_timeout":300,"offchain_url":null,"auto_deposit_cap_per_cycle":5000000,"offchain_poll_interval_seconds":3.0,"offchain_poll_timeout_seconds":300.0,"offchain_failover_max_retries":2}}
+      use_offchain: ${USE_OFFCHAIN:bool:false}
```
jmoreira-valory, line 141: 🔵 **F1 · LOW — Fail-loud migration guard is unreachable when use_mech_marketplace is true**

The PR body's migration story is that a stale `MECH_MARKETPLACE_CONFIG` produces "a `ValueError` at agent boot with a message naming the replacement env vars". On this service that message will not appear, because `use_mech_marketplace` defaults to **true** here (line 139).

The trader has a *second*, unguarded construction site for the same dict. `TraderParams` lists `CheckStopTradingParams` first in its MRO (`trader_abci/models.py:162-169`), and `CheckStopTradingParams.__init__` calls `_configure_marketplace(kwargs)` **before** `super().__init__(...)`, which does:

```python
# packages/valory/skills/check_stop_trading_abci/models.py:106
self.mech_marketplace_config = MechMarketplaceConfig(**kwargs["mech_marketplace_config"])
```

`MechMarketplaceConfig` is a frozen dataclass that no longer declares either key, so an operator who forgets the migration gets a bare `TypeError: __init__() got an unexpected keyword argument 'use_offchain'` out of a dataclass constructor deep inside model instantiation — never `MechParams`' self-describing guard, which sits further down the same `super()` chain.

Not blocking: both paths still fail loudly at boot and the offending key name is in the message, so nothing runs with a half-applied config. But it costs the fleet the one migration affordance this PR advertises, on exactly the service most likely to be marketplace-enabled. It is also a durable trap — every future upstream `MechMarketplaceConfig` schema change will surface here first, opaquely.

Fix: mirror the upstream guard in `_configure_marketplace`, e.g. reuse the exported `_LEGACY_MARKETPLACE_KEYS` (or a small local frozenset) and raise the same operator-facing `ValueError` before constructing the dataclass. `check_stop_trading_abci/tests/test_models.py` already has a `_configure_marketplace` test class, so a regression test is a few lines.

packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py
```diff
@@ -2551,6 +2551,7 @@ def test_use_offchain_false_returns_cached_total_no_rpc(self) -> None:
         b = _make_fetch_behaviour()
         ctx, params, synced_data, _ = _mock_context()
         params.mech_marketplace_config = _make_marketplace_config(use_offchain=False)
+        params.use_offchain = False
```
jmoreira-valory, line 2554: 🔵 **F2 · LOW — Tests cannot catch a regression back to the nested use_offchain read**

All 14 updated sites now set the nested stub *and* the top-level param to the same value:

```python
params.mech_marketplace_config = _make_marketplace_config(use_offchain=False)
params.use_offchain = False
```

and the helper at line 2436 still pins the removed attribute onto the mock:

```python
cfg = MagicMock()
cfg.use_offchain = use_offchain
```

Because both sources always agree, reverting `behaviours.py:578` to `self.params.mech_marketplace_config.use_offchain` leaves the entire suite green — the one line this PR changes in production code is untested against the shape change it exists to make.

Fix: drop the `use_offchain` kwarg and the `cfg.use_offchain` assignment from `_make_marketplace_config` (the helper then just returns a bare `MagicMock`). A `MagicMock` auto-attribute is truthy, so any regression to the nested read would send the `use_offchain=False` tests down the RPC branch and fail them — turning the 14 call sites into a real guard instead of a mirror. The `use_offchain=X` argument at each call site can go with it.

packages/valory/agents/trader/aea-config.yaml
```diff
@@ -259,6 +259,8 @@ models:
       use_mech_marketplace: ${USE_MECH_MARKETPLACE:bool:false}
       policy_store_update_offset: ${int:259200}
       mech_marketplace_config: ${MECH_MARKETPLACE_CONFIG:dict:{"mech_marketplace_address":"0x0000000000000000000000000000000000000000","priority_mech_address":"0x0000000000000000000000000000000000000000","priority_mech_staking_instance_address":"0x0000000000000000000000000000000000000000","priority_mech_service_id":0,"requester_staking_instance_address":"0x0000000000000000000000000000000000000000","response_timeout":300}}
+      use_offchain: ${USE_OFFCHAIN:bool:false}
```
jmoreira-valory, line 262: ⚪ **F3 · NIT — USE_OFFCHAIN is exposed at agent level but can never be set to true there**

The agent-level `MECH_MARKETPLACE_CONFIG` default (line 261) has no `auto_deposit_cap_per_cycle` and no `offchain_url`, so `validate_offchain_params` in v0.32.7 rejects `use_offchain=True` unconditionally on this layer:

```
use_offchain requires auto_deposit_cap_per_cycle to be set ...
```

That means `USE_OFFCHAIN=true` against a bare `make run-agent` (the dev path used in the test plan, with the flag set to `false`) is guaranteed to die at boot unless the whole dict is overridden too. Both `service.yaml`s do carry the cap, so no deployment is affected — this is purely about the dev-loop surface being self-consistent. Either add `"auto_deposit_cap_per_cycle"` to the agent-level JSON default alongside the other placeholder values, or leave it and accept that the off-chain branch is service-layer-only.

jmoreira-valory (approved):

**What this does.** Upstream mech-interact v0.32.7 moved two knobs — `use_offchain` and `offchain_deposit_target_calls` — out of the nested `mech_marketplace_config` dict and up to top-level `MechParams` attributes. This PR bumps the `mech_interact_abci` pin to that release, re-locks the whole CID cascade, and rewires every trader-side overlay (agent aea-config, both service.yamls, trader_abci + decision_maker_abci skill.yamls) plus the single Python consumer in `agent_performance_summary_abci`. It is a pure config-shape migration: no FSM, round, or behaviour-logic change, and the default values are byte-identical to before.

Verified independently against HEAD rather than taking the PR body's word for it:

- **Upstream contract** — at tag `v0.32.7`, `MechParams.__init__` does `_ensure("use_offchain", kwargs, bool)` and `_ensure("offchain_deposit_target_calls", kwargs, int)` at top level, plus a `_LEGACY_MARKETPLACE_KEYS` rejection on the nested dict. So the hoist really is mandatory, not cosmetic.
- **CID** — `bafybeic5xfxdod5l56o7aojdamojnvi562nql2mlah73bf47d4guz4xnwi` matches `mech-interact@v0.32.7`'s own `packages.json` exactly. No phantom-CID drift, and `upstream_pins` in `pyproject.toml` was bumped in step.
- **3-layer parity** — `trader_abci/skill.yaml` (the only `is_abstract: false` skill, so the operative one), `agents/trader/aea-config.yaml`, and both `service.yaml`s all carry both keys at top level, and the legacy keys are gone from every nested JSON default. Checked; no layer is missing either key.
- **Consumer count** — a repo-wide code search confirms `use_offchain` has exactly one Python reader (`agent_performance_summary_abci/behaviours.py`), and the swap at line 578 preserves the semantics exactly (no orphaned local, no other use of the removed `marketplace_config` binding in that function).
- **Cross-repo migration blast radius** — `olas-operate-middleware` composes `MECH_MARKETPLACE_CONFIG` at `operate/services/manage.py:541-548` with neither legacy key, so the Pearl path boots clean; `quickstart` and `agent-deployments` carry no trader-side `MECH_MARKETPLACE_CONFIG` at all. The remaining exposure is genuinely just hand-set Propel / `.env` values, which the PR body calls out.
- CI is 28/28 green at this HEAD.

No FSM, composition, or round-transition surface is touched, so the usual 4-place sync and unmapped-final-state checks don't apply here.

**Findings** — 3 · 🔵 2 Low · ⚪ 1 Nit

| F1 | 🔵 LOW | [packages/valory/services/polymarket_trader/service.yaml:141](https://github.com/valory-xyz/trader/blob/b23ce9f67a6056373914832ac8941f1db8e0fd6d/packages/valory/services/polymarket_trader/service.yaml#L141) | **Fail-loud migration guard is unreachable when use_mech_marketplace is true** |
| F2 | 🔵 LOW | [packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py:2554](https://github.com/valory-xyz/trader/blob/b23ce9f67a6056373914832ac8941f1db8e0fd6d/packages/valory/skills/agent_performance_summary_abci/tests/test_behaviours.py#L2554) | **Tests cannot catch a regression back to the nested use_offchain read** |
| F3 | ⚪ NIT | [packages/valory/agents/trader/aea-config.yaml:262](https://github.com/valory-xyz/trader/blob/b23ce9f67a6056373914832ac8941f1db8e0fd6d/packages/valory/agents/trader/aea-config.yaml#L262) | **USE_OFFCHAIN is exposed at agent level but can never be set to true there** |

PR #1023, opened 2026-08-13
fix(polymarket): stop the bet-placement loop on an under-minimum order

OjusWiZard:
Closes [OPE-1883](https://linear.app/valory-xyz/issue/OPE-1883). ZD#1204.

## The bug

A Polystrat agent sat in `polymarket_bet_placement_round` for over an hour, logging "Opening a trade on Polymarket" 8+ times without advancing. No self-recovery — it needed a restart, and a stuck agent risks eviction after two missed staking epochs.

The chain:

1. `polymarket_top_up` sweeps the DepositWallet, then transfers **exactly** `bet_amount` pUSD Safe→DW.
2. `_place_bet` passes that live DW balance to the CLOB SDK as `user_usdc_balance`. In `py_clob_client_v2/fees.py`:
   ```python
   total_cost = amount + platform_fee + builder_fee
   if user_usdc_balance <= total_cost:
       return max(user_usdc_balance - platform_fee - builder_fee, 0)
   ```
   Because `balance == amount` by construction, that branch is taken on **every** buy on a fee-charging market. The shrink is not an edge case.
3. `absolute_min_bet_size` is `1_000_000` (= $1.00) and Polymarket's marketable minimum is *also* $1.00, so a floor-sized bet lands at ~$0.98 and is refused:
   `invalid amount for a marketable BUY order ($0.98), min size: 1`
4. The rejected signed order was cached, and `BET_PLACEMENT_FAILED` transitions **back into the same round** (`rounds.py:479`). The cache key contains `period_count`, which doesn't advance inside the loop — so the identical doomed order was replayed forever.

The $0.98 in the ticket is not a price coincidence; any $1.00 bet on a fee-charging market lands there.

## The fix

Three independent cuts, so the loop cannot re-form if one is bypassed.

**1. Fund the fee rather than letting the SDK carve it out of the bet.** A new `QUOTE_BUY` request prices the fee a round ahead, and the top-up transfers `bet + ceil(1.5 × fee)`. The SDK then leaves the order at the full bet. This also fixes a quieter pre-existing discrepancy: `kelly_criterion` charges `fee_per_trade` *on top of* `spend` (it's in the growth objective, not deducted from the budget), so the agent had been buying fewer shares than the strategy sized on every trade.

**2. Refuse an order the venue cannot take, before signing it.** `_buy_sizing` measures price, fee and spendable amount through the SDK's own `calculate_market_price` / `_adjust_buy_amount_for_balance` rather than reimplementing the fee formula, so the preflight and the order it guards cannot disagree. `_below_minimum_reason` separates the two causes — *bet too small* (raise it) vs *fee shrank it* (fund it) — because they need opposite fixes. Ported from the equivalent fix in `valory-xyz/connect` (PR #51).

**3. Leave the round instead of retrying.** A below-minimum refusal emits `BET_PLACEMENT_IMPOSSIBLE` → `BlacklistingRound` (a queue demotion via `next_status()`, **not** `blacklist_forever`) and drops the cache entry. The venue's own `"invalid amount"` rejection is treated the same way and withholds `signed_order_json`, so a terminal refusal can never be cached.

### Failing safe

The preflight is advisory in both directions:

- an unreadable **book** leaves the order for the CLOB to rule on;
- an unreadable **DepositWallet balance** skips the check entirely. This one matters: the nominal-amount fallback is numerically identical to "the DW holds exactly the bet", which forces the fee shrink — so judging it would demote a correctly funded market on any transient RPC blip. `_read_dw_collateral_balance` exists purely to make that distinction visible to the caller.

## Verification

- **2320 tests pass** across `decision_maker_abci`, `polymarket_client` and `market_manager_abci`.
- **100% line and branch coverage** on all four modified production modules.
- `check-hash`, `check-packages`, `check-abci-docstrings` clean; black/isort/flake8 clean; mypy and pylint at their pre-existing baselines for these files.
- No FSM change — `BET_PLACEMENT_IMPOSSIBLE` was already wired to `BlacklistingRound`.
- `third_party` hashes untouched.

Notable tests:

| Test | Pins |
|---|---|
| `test_reserve_clears_the_venue_minimum_end_to_end` | the real `_quote_buy` → real `_top_up_amount` → real `_place_bet` gate, asserting the order reaches the book at the full bet |
| `test_async_act_below_minimum_leaves_the_round` | event is `BET_PLACEMENT_IMPOSSIBLE` **and** the cache is emptied |
| `test_place_bet_skips_the_preflight_when_the_balance_is_unreadable` | an RPC blip cannot demote a funded market |
| `test_below_minimum_reason_allows_exactly_the_minimum` | the `>=` boundary, by name |
| `test_place_bet_does_not_cache_a_venue_minimum_rejection` | a post-preflight price move still ends terminally |

This was reviewed by a multi-agent pass before submission; the findings it raised (the RPC-blip path, two comments the change had staled, a `blocked`/`blocked_reason` drift hazard, and the missing seam test) are all addressed here.


packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_top_up.py
```diff
@@ -206,6 +216,251 @@ def test_unresolvable_dw_insufficient(self, tmp_path) -> None:  # type: ignore[n
...
+                PolymarketTopUpBehaviour,
+                "tx_hex",
+                new_callable=PropertyMock,
+                return_value="0xsafehash",
+            ),
+        ):
+            payload = _drive(behaviour)
+        assert payload.event == Event.PREPARE_TX.value
+        # 1.0 pUSD + ceil(0.02 * 1.5 * 1e6) = 1_030_000 base units.
+        transferred = int(behaviour.multisend_batches[0].data.hex()[-64:], 16)
```
DIvyaNautiyal07, line 255: **P1 — `math.ceil` vs `math.floor` not distinguished.**

`0.02 * 1.5 * 1e6 = 30_000.0` is an exact integer, so `ceil` and `floor` return the same value. A regression that swapped `math.ceil` for `math.floor` in `_top_up_amount` — the whole reason the reserve is rounded — would still pass this test.

Use a fee that is not an integer multiple: `fee_usd: 0.021` gives `ceil = 31_500` vs `floor = 31_499`.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_top_up.py
```diff
@@ -206,6 +216,251 @@ def test_unresolvable_dw_insufficient(self, tmp_path) -> None:  # type: ignore[n
...
+        )
+        behaviour._sampled_outcome_token_id = lambda: "token123"  # type: ignore[method-assign]
+        _set_balance(behaviour, 1_000_000)
+        with patch.object(
+            PolymarketTopUpBehaviour,
+            "synchronized_data",
+            new_callable=PropertyMock,
+            return_value=_synced(DW, 1_000_000),
+        ):
+            payload = _drive(behaviour)
```
DIvyaNautiyal07, line 272: **P1 — Missing "no transfer enqueued" assertion.**

The event check is right, but nothing verifies the multisend was skipped. A refactor that enqueued the transfer *before* checking the balance would pass this test while doing the wrong thing.

Add `assert not behaviour.multisend_batches` after the event assertion.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_bet_placement.py
```diff
@@ -593,6 +593,80 @@ def mock_wait(condition) -> None:  # type: ignore[no-untyped-def, misc]
         assert len(payloads_sent) == 1
         assert payloads_sent[0].event == Event.BET_PLACEMENT_IMPOSSIBLE.value
 
+    def test_async_act_below_minimum_leaves_the_round(self) -> None:
```
DIvyaNautiyal07, line 596: **P1 — Flag-symmetry gap: no explicit test for the `below_minimum`-absent path.**

The new `if response.get("below_minimum")` branch is tested for `True` here. The absent case is covered *implicitly* by `test_async_act_failure_with_signed_order_caches`, but that test predates this PR and is not named for the gate.

A future refactor that widens the gate (e.g. defaults `below_minimum` to `True`, or evaluates against a truthy error string) would silently convert every placement failure into a blacklisting and still pass every named test.

Add an explicit `test_async_act_non_terminal_failure_caches_the_order`: feed `{"error": "…", "signed_order_json": "{…}"}` with no `below_minimum` key, assert `BET_PLACEMENT_FAILED` and non-empty cache.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -574,6 +605,134 @@ def _test_connection(self) -> bool:
...
+                self.client._adjust_buy_amount_for_balance(  # pylint: disable=protected-access
+                    token_id, amount, price, amount, builder_code
+                )
+            )
+            affordable = float(
+                self.client._adjust_buy_amount_for_balance(  # pylint: disable=protected-access
+                    token_id, amount, price, balance, builder_code
+                )
+            )
+        except Exception as e:  # noqa: BLE001 - advisory preflight, never fatal
```
DIvyaNautiyal07: **P2 — Bare `except Exception` swallows coding bugs, not just RPC blips.**

An `AttributeError` from an SDK rename of `_adjust_buy_amount_for_balance`, a `TypeError` from a signature change, or any other bug here degrades silently to "no preflight" — reintroducing the exact fee-shrink bug this PR was written to catch. The `warning` log is the only trace.

Either narrow to `(requests.exceptions.RequestException, ValueError, KeyError)` or whatever the SDK actually raises, or bump the log level to `error` for unexpected exception types so a coding bug leaves an operator-visible signal.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -574,6 +605,134 @@ def _test_connection(self) -> bool:
...
+        they will always say blocked. They answer "could this order go out
+        now?", not "how much should I send?".
+
+        :param token_id: CTF token id of the outcome to buy.
+        :param amount: the nominal pUSD spend.
+        :param funder: DepositWallet address, for the balance read.
+        :return: ``(quote, None)``; ``fee_usd`` is ``None`` when unmeasurable.
+        """
+        self._ensure_dw_funder(funder or self.dw_address)
+        balance = self._dw_collateral_balance(funder or self.dw_address, amount)
```
DIvyaNautiyal07: **P2 — `_quote_buy` uses the fallback wrapper, losing the None signal.**

`_dw_collateral_balance` folds an RPC failure into `fallback = amount`, which makes the balance numerically equal to the bet and forces `spendable = amount - fee` in `_buy_sizing`. On a floor-sized bet during a transient RPC blip, the quote reports `blocked = True` with no way for the caller to tell that apart from a real block.

Today `_top_up_amount` only reads `fee_usd`, so this doesn't misbehave, but the field's own docstring says `blocked` answers "could this order go out now?" — and it lies under a blip. Switch to `self._read_dw_collateral_balance(...)` here and short-circuit `sizing` when it returns `None`.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -1469,29 +1664,47 @@ def _encode_erc1155_safe_transfer(
...
             return self._erc20_balance_of(self.collateral_address, dw) / 10**6
         except Exception as e:  # noqa: BLE001 - best-effort on-chain read
             self.logger.warning(
                 f"Could not read DepositWallet pUSD balance ({e}); "
                 "reserving the fee from the nominal order amount."
             )
-            return fallback
+            return None
+
+    def _dw_collateral_balance(self, dw: Optional[str], fallback: float) -> float:
```
DIvyaNautiyal07: **P2 — Wrapper is now a one-liner with only two call sites.**

`_dw_collateral_balance` is `_read_dw_collateral_balance(dw) or fallback`-shaped, one call in `_place_bet` and one in `_quote_buy` (the latter arguably shouldn't be using the fallback form — see the `_quote_buy` comment).

Fold into `_read_dw_collateral_balance(self, dw, fallback=None)` and drop the wrapper.

packages/valory/connections/polymarket_client/tests/test_connection.py
```diff
@@ -440,6 +441,13 @@ def test_place_bet_success_no_cache(self) -> None:
         so the caller can retry with the same order if the submission fails.
         """
         conn = _make_connection()
+        # Arm the sizing preflight with a funded DW and a real fee model so
+        # this test clears the CLOB-minimum gate by design. Left unstubbed,
+        # the MagicMock client makes every sizing figure 1.0, landing the
+        # order exactly on the minimum and passing only because the
+        # comparison is ``>=`` — a boundary this test never meant to assert.
+        conn._read_dw_collateral_balance = MagicMock(return_value=100.0)
+        _stub_sizing_client(conn, price=0.5, fee_rate=0.02)
```
DIvyaNautiyal07: **P2 — Six near-identical "arm the sizing preflight" blocks.**

The same 7 lines (`_read_dw_collateral_balance` mock + `_stub_sizing_client` + 5-line comment, word-for-word) repeat across six consecutive test methods (this one and lines 496–497, 533–534, 562–563, 583–584, 601–602). That's ~42 lines that mean one thing.

Extract into `_arm_preflight(conn)` with the comment moved there once; call it at the top of each test.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_top_up.py
```diff
@@ -144,6 +163,83 @@ def _prepare_top_up(self) -> Generator[None, None, None]:
...
+            # the pre-fix behaviour — the bet gets funded bare, the SDK shrinks
+            # the order by the fee, and a floor-sized bet is then refused. Say so
+            # here or the refusal looks unexplained a round later.
+            self.context.logger.warning(
+                "No CLOB fee quote available; topping up the bare bet. The order "
+                "may be shrunk by the taker fee and refused if it is near the "
+                "venue minimum."
+            )
+            return buy_amount
+        if fee <= 0:
```
DIvyaNautiyal07: **P2 — `fee <= 0` collapses two different signals silently.**

A `fee == 0` is a fee-free market — correctly silent. A `fee < 0` is nonsense crossing a process boundary (the comment even calls the negative arm "unreachable with a well-behaved SDK"), which is exactly when it deserves a `warning`.

Split: `if fee == 0: return buy_amount` (silent); `if fee < 0: self.context.logger.warning(...); return buy_amount`. The `test_negative_fee_is_refused` test would then also gain a `warning.assert_called_once()`.

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -98,6 +98,24 @@
...
+# the shrunk figure and the minimum. Re-signing cannot change the amount, so a
+# rejection carrying this is terminal: the order must not be cached for retry.
+#
+# Deliberately matched broadly rather than pinned to the full "min size" phrasing.
+# The two ways to be wrong are not symmetric: treating an unrelated "invalid
+# amount ..." rejection as terminal costs one queue demotion in
+# ``BlacklistingRound``, while failing to recognise a real one re-caches the
+# order and puts the FSM back in the unbounded BET_PLACEMENT_FAILED self-loop
+# this whole path exists to break. Prefer the cheap error.
+BELOW_MINIMUM_ERROR = "invalid amount"
```
DIvyaNautiyal07: **P2 — Broad substring by design, but the false-positive class is unbounded.**

The comment above weighs the tradeoff — false positive = one blacklist (cheap), false negative = infinite loop (expensive) — and the reasoning is sound. But an upstream reword to any `"invalid amount ..."` shape (e.g. `"invalid amount for a limit order"`, `"invalid amount of collateral"`) blacklists whole classes of markets, not one queue demotion.

A safer form still prefers the cheap error but keeps the pattern shape-specific: require both `"invalid amount"` and `"min size"` (or `"marketable"`) in the message. Not blocking — hardening.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_top_up.py
```diff
@@ -37,16 +38,30 @@
 )
 
 ERC20_TRANSFER_SELECTOR = "0xa9059cbb"  # keccak("transfer(address,uint256)")[:4]
+# pUSD base units per whole token. Spelled out rather than taken from
+# ``get_token_precision()`` so it stays the exact inverse of ``usdc_to_native``,
+# which hardcodes the same 6 decimals two lines below — deriving one of the pair
```
DIvyaNautiyal07: **Nit — "two lines below" is wrong.**

The referenced `usdc_to_native` isn't defined here — it's a method inherited from the base, and the *call* site (`self.usdc_to_native(buy_amount)` inside `_top_up_amount`) is well below this constant. "Two lines below" is comment rot waiting to happen.

Reword: "shares the 6 pUSD decimals with `usdc_to_native`" (no positional reference).

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -574,6 +605,134 @@ def _test_connection(self) -> bool:
...
+            "blocked": False,
+            "blocked_reason": None,
+        }
+        if sizing is None:
+            return quote, None
+        # ``blocked`` is derived from the reason rather than recomputed against
+        # MIN_MARKETABLE_USD: two expressions of one rule can drift apart into
+        # "blocked with no reason".
+        reason = self._below_minimum_reason(amount, sizing.spendable)
+        quote.update(
```
DIvyaNautiyal07, line 716: **Nit — `_place_bet` return type is `Tuple[Any, Any]` but the shape is now a three-way union.**

The return is one of `{error, below_minimum}`, `{error, signed_order_json}`, or `{status, ...}`. Nothing in the type system prevents a future termination cause from inventing a fourth shape, and the behaviour's dispatch stays a chain of `.get("below_minimum")` + string-matching on `str(response)`.

A `TypedDict` union — or better, a `termination_reason: Optional[TerminationReason]` enum key that both `"below_minimum"` and the existing `"No orderbook exists"` case share — would make the caller's dispatch exhaustive by construction.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_polymarket_top_up.py
OjusWiZard: Fixed, but not with `0.021` — that one has the same problem. `0.021 * 1.5 * 1e6` is exactly `31500.0`, so `ceil` and `floor` still agree.

It turns out most plausible fee values land on an integer after `* 1.5 * 1e6`:

```
0.021   -> 31500.0                ceil == floor
0.0201  -> 30150.0                ceil == floor
0.02001 -> 30015.0                ceil == floor
0.0123  -> 18450.0                ceil == floor
0.007   -> 10500.0                ceil == floor
0.013   -> 19500.0                ceil == floor
0.017   -> 25500.000000000004     ceil 25501 / floor 25500   <-
0.019   -> 28499.999999999996     ceil 28500 / floor 28499   <-
```

Using `0.017`, so the pinned transfer is `1_025_501`. Swapping `math.ceil` for `math.floor` now fails the test. Good catch — the assertion was decorative as written.

packages/valory/connections/polymarket_client/connection.py
OjusWiZard: Dropped the wrapper, but without the `fallback=None` parameter — that would put the conflation straight back.

The reason the two methods existed at all is that `fallback` makes an unreadable balance numerically identical to "the DW holds exactly the bet", which is the exact state that makes the SDK shrink the order by the fee. Any caller that receives the fallback cannot tell a blip from a real under-funding. Keeping `fallback` as a parameter would let a future caller opt back into that without noticing.

So: one method, `_read_dw_collateral_balance(dw) -> Optional[float]`, and `_place_bet` applies the fallback at the call site — where it is still wanted for `MarketOrderArgs.user_usdc_balance` (so the fee is reserved rather than risking a "not enough balance" rejection) but explicitly *not* passed to the preflight, which now takes the raw `Optional` and stands aside when it is `None`.

Updated the three `_dw_collateral_balance` tests in `test_connection_dw.py` to the new contract.

packages/valory/connections/polymarket_client/connection.py
OjusWiZard: Done — `BELOW_MINIMUM_MARKERS = ("invalid amount", "min size")`, all must match.

Flagging the cost, though, because I do not think it is free and you may want to weigh it again.

The two failure directions are asymmetric in the *other* direction too. A false positive is one queue demotion. A false negative re-caches the order and puts the FSM back in the unbounded `BET_PLACEMENT_FAILED` self-loop — the bug this PR exists to close. Narrowing the pattern makes false negatives more likely, not less.

The preflight is only a partial backstop: it stands aside when the book is unpriceable *or* when the DW balance is unreadable, so there are live paths where this substring match is the only thing between us and the loop. If the venue rewords to drop `min size` (say `"invalid amount ($0.98), minimum 1"`), we are back to spinning.

I have taken your version because an unbounded blacklist class is a real risk and the wording has been stable. But if we see this fire wrongly, the fix I would reach for is a retry counter on the round rather than loosening the match again — that bounds both failure modes instead of trading one for the other.

packages/valory/connections/polymarket_client/connection.py
OjusWiZard: Not doing this one here — flagging rather than silently skipping it.

The suggestion is right, but it is not a rename. The two terminal causes are classified in different layers today: `below_minimum` is set by the connection, while `"No orderbook exists for the requested token id"` is detected by the *behaviour* doing `str(response)` matching. A shared `termination_reason` only pays off if both are set in one place, which means moving the no-orderbook classification into `_place_bet` and rewriting the behaviour's dispatch chain plus the tests that feed it raw response dicts.

Half-doing it — a `termination_reason` key alongside the existing `below_minimum` flag and string match — leaves three markers where there were two, which is worse than what is here now.

Happy to do it as its own PR against the same round, where the diff is legible as a dispatch refactor rather than buried in a bugfix. Want me to open one?

packages/valory/connections/polymarket_client/connection.py
OjusWiZard: Switched to `_read_dw_collateral_balance` — but deliberately **not** short-circuiting `sizing` when it returns `None`, which is the other half of what you asked for. Explaining, because the difference matters.

Short-circuiting would drop `fee_usd`, and `fee_usd` is the only field the top-up actually consumes. `_top_up_amount` treats a missing fee as "no reserve" and funds the bare bet — which is precisely the state that makes the SDK shrink the order by the fee. So a transient RPC blip at quote time would silently under-fund the DepositWallet and recreate the OPE-1883 failure a round later, with only a connection-layer warning to explain it.

The fee does not need the balance: it is measured at full size, by calling the SDK with `balance == amount`. Only `spendable` does. So instead of short-circuiting, `balance` is now `Optional` all the way through `_buy_sizing`:

- `fee` / `price` / `required_usd` — always measured, balance-independent.
- `spendable_usd` — `None` when the balance is unreadable.
- `blocked` — `False`, via `_below_minimum_reason` returning `None` for an unjudgeable balance.

That gets the property you were after (a blip can no longer report `blocked: true` and be indistinguishable from a real block) without the side effect of losing the number the caller came for. Pinned by `test_buy_sizing_without_a_balance_still_measures_the_fee`.

Shout if you would rather have the short-circuit and a separate fee-only path.

packages/valory/connections/polymarket_client/connection.py
OjusWiZard: Took both halves of the "either/or" rather than one.

`EXPECTED_SIZING_ERRORS = (PolyApiException, requests.exceptions.RequestException, ValueError, KeyError)` — a thin book, a market with no fee schedule, a transient CLOB read — stays a `warning`. Anything else logs at `error` naming the exception type and says explicitly that the preflight was skipped.

Narrowing alone would have turned an SDK rename into an uncaught exception escaping `_place_bet`, which is a different failure rather than a louder one; the error-level catch-all keeps the "advisory, never fatal" contract while making a bug visible. Both arms pinned: `test_buy_sizing_returns_none_when_the_book_is_unreadable` (ValueError, warning, no error) and `test_buy_sizing_logs_an_unexpected_error_loudly` (AttributeError, error, no warning).

packages/valory/skills/decision_maker_abci/behaviours/polymarket_top_up.py
OjusWiZard: Right, the positional reference was wrong. Removed the comment rather than rewording it — the constant now carries one line saying it must stay the exact inverse of the 6 decimals `usdc_to_native` hardcodes, and the reasoning for not deriving it from `get_token_precision()` moved to the commit message.

packages/valory/connections/polymarket_client/tests/test_connection.py
```diff
@@ -565,6 +572,345 @@ def test_place_bet_post_order_none_response(self) -> None:
...
+        assert quote["required_usd"] == pytest.approx(1.02)
+        # Funded with exactly the bet, the order lands at 0.98 — under the floor.
+        assert quote["blocked"] is True
+        assert "taker fee" in quote["blocked_reason"]
+
+    def test_quote_buy_degrades_when_unpriceable(self) -> None:
+        """An unreadable book yields a fee-less quote, never a block."""
+        conn = _make_connection()
+        conn._read_dw_collateral_balance = MagicMock(return_value=1.0)
+        conn.client.calculate_market_price.side_effect = Exception("no book")
```
DIvyaNautiyal07, line 737: **Nit — raw `Exception` now hits the *unexpected* branch.**

The refactor sends bare `Exception` down the `except Exception` arm of `_buy_sizing` (logs at `error`, not `warning`). This test (and `test_quote_buy_keeps_one_key_set_when_unpriceable` on line 846) means to model "book unreadable" — an *expected* sizing failure. Behavior assertions still pass, but the tests now exercise the wrong branch.

Change `side_effect = Exception("no book")` to `ValueError("no book")` (or `PolyApiException(...)`) so they stay on the expected path.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_top_up.py
```diff
@@ -37,16 +38,20 @@
 )
 
 ERC20_TRANSFER_SELECTOR = "0xa9059cbb"  # keccak("transfer(address,uint256)")[:4]
+# Must stay the exact inverse of the 6 decimals ``usdc_to_native`` hardcodes.
+PUSD_UNITS = 10**6
```
DIvyaNautiyal07, line 42: **Nit — `FEE_HEADROOM_RATIO` lost its rationale in the trim.**

The old comment explained the fee moves ~0.5%–3.7% across the price range (why 1.5×). Now the constant sits bare and a future tweak has nothing to consult.

One line is enough: `# ≈50% cushion for fee drift between quote and settlement.`

packages/valory/connections/polymarket_client/connection.py
```diff
@@ -619,16 +768,22 @@ def _place_bet(
                     )
 
             if signed is None:
```
DIvyaNautiyal07, line 770: **Nit — CLOB fee-formula reference comment was deleted alongside the refactor.**

The pre-existing block documenting `GET /fee-rate` + market exponent (`fee = (amount/price) * rate * (price*(1-price))^e`) is gone. The remaining code no longer explains where the SDK's fee comes from.

Worth a one-liner near `_buy_sizing` pointing at the SDK helper (`_adjust_buy_amount_for_balance`) so the next reader doesn't have to reverse-engineer the fee shape.

DIvyaNautiyal07 (approved):
LGTM — all substantive findings from the previous review are addressed. Three nits inline about test-branch alignment and comment context; none blocking.

packages/valory/connections/polymarket_client/tests/test_connection.py
DIvyaNautiyal07, line 737: **Nit — raw `Exception` now hits the *unexpected* branch.**

The refactor sends bare `Exception` down the `except Exception` arm of `_buy_sizing` (logs at `error`, not `warning`). This test (and `test_quote_buy_keeps_one_key_set_when_unpriceable` on line 846) means to model "book unreadable" — an *expected* sizing failure. Behavior assertions still pass, but the tests now exercise the wrong branch.

Change `side_effect = Exception("no book")` to `ValueError("no book")` (or `PolyApiException(...)`) so they stay on the expected path.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_top_up.py
DIvyaNautiyal07, line 42: **Nit — `FEE_HEADROOM_RATIO` lost its rationale in the trim.**

The old comment explained the fee moves ~0.5%–3.7% across the price range (why 1.5×). Now the constant sits bare and a future tweak has nothing to consult.

One line is enough: `# ≈50% cushion for fee drift between quote and settlement.`

packages/valory/connections/polymarket_client/connection.py
DIvyaNautiyal07, line 770: **Nit — CLOB fee-formula reference comment was deleted alongside the refactor.**

The pre-existing block documenting `GET /fee-rate` + market exponent (`fee = (amount/price) * rate * (price*(1-price))^e`) is gone. The remaining code no longer explains where the SDK's fee comes from.

Worth a one-liner near `_buy_sizing` pointing at the SDK helper (`_adjust_buy_amount_for_balance`) so the next reader doesn't have to reverse-engineer the fee shape.

DIvyaNautiyal07 (approved):
LGTM — all substantive findings from the previous review are addressed. Three nits inline about test-branch alignment and comment context; none blocking.

packages/valory/connections/polymarket_client/tests/test_connection.py
DIvyaNautiyal07, line 737: **Nit — raw `Exception` now hits the *unexpected* branch.**

The refactor sends bare `Exception` down the `except Exception` arm of `_buy_sizing` (logs at `error`, not `warning`). This test (and `test_quote_buy_keeps_one_key_set_when_unpriceable` on line 846) means to model "book unreadable" — an *expected* sizing failure. Behavior assertions still pass, but the tests now exercise the wrong branch.

Change `side_effect = Exception("no book")` to `ValueError("no book")` (or `PolyApiException(...)`) so they stay on the expected path.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_top_up.py
DIvyaNautiyal07, line 42: **Nit — `FEE_HEADROOM_RATIO` lost its rationale in the trim.**

The old comment explained the fee moves ~0.5%–3.7% across the price range (why 1.5×). Now the constant sits bare and a future tweak has nothing to consult.

One line is enough: `# ≈50% cushion for fee drift between quote and settlement.`

packages/valory/connections/polymarket_client/connection.py
DIvyaNautiyal07, line 770: **Nit — CLOB fee-formula reference comment was deleted alongside the refactor.**

The pre-existing block documenting `GET /fee-rate` + market exponent (`fee = (amount/price) * rate * (price*(1-price))^e`) is gone. The remaining code no longer explains where the SDK's fee comes from.

Worth a one-liner near `_buy_sizing` pointing at the SDK helper (`_adjust_buy_amount_for_balance`) so the next reader doesn't have to reverse-engineer the fee shape.

DIvyaNautiyal07 (approved):
LGTM — all substantive findings from the previous review are addressed. Three nits inline about test-branch alignment and comment context; none blocking.

ROMAN-VALORY (approved):

packages/valory/connections/polymarket_client/tests/test_connection.py
OjusWiZard: Good catch, and it is broader than the two you named — it applies to three tests, all of which meant to model "book unreadable" and were landing on the *unexpected* arm after the refactor:

- `test_quote_buy_degrades_when_unpriceable` (737)
- `test_place_bet_falls_through_when_unpriceable` (786)
- `test_quote_buy_keeps_one_key_set_when_unpriceable` (846)

`ValueError("no book")` in all three, which puts them back on the `EXPECTED_SIZING_ERRORS` path. Left `get_ok` (400) and `get_order_book` (3394) alone — different call sites, unrelated to sizing.

**Not landing it in this PR.** The change is written and verified locally (full suite green, 100% coverage held, lint clean at CI scope) but held back deliberately: another commit here dismisses the approval and re-blocks the release. It goes in the next follow-up on this area.

Same applies to the identical re-posts of this comment at 14:41 and 14:47 — treating this thread as the live one.

packages/valory/connections/polymarket_client/connection.py
OjusWiZard: Agreed, and restored — as a line in `_buy_sizing`'s docstring rather than a block comment, since that is where a reader lands when they want to know where the numbers come from:

> Price comes from `calculate_market_price` and the fee from what `_adjust_buy_amount_for_balance` declines to spend at full size, so this and the order that follows it cannot disagree.

I deliberately did not restore the `fee = (amount/price) * rate * (price*(1-price))^e` formula itself. It is the SDK's, not ours, and copying it here creates a second copy to drift out of date the next time `py_clob_client_v2` changes its fee model — which is the same reason `_buy_sizing` calls the SDK to *measure* the fee instead of computing it. Pointing at the helper gets the reader there without duplicating it.

**Not landing it in this PR** — written and verified locally, but held back so no new commit dismisses the approval and re-blocks the release. Goes in the next follow-up.

Covers the identical re-posts at 14:41 and 14:47 too.

packages/valory/skills/decision_maker_abci/behaviours/polymarket_top_up.py
OjusWiZard: Holding this one, deliberately — the rationale was moved rather than lost.

The 0.5%–3.7% drift figure and the reasoning for 1.5× are in the commit that introduced the constant (`5fe03a164`) and in the PR description. Our convention here is that a comment earns its place only when `git blame` → commit → PR *cannot* answer the question; when it can, the comment is a second copy of the answer that goes stale the next time someone edits the line above it. That is exactly why the trim happened.

So a future tweak does have something to consult — `git blame` on the line lands on a commit whose body explains the fee curve and why the cushion is 50%. That is more durable than the comment was, because it cannot drift away from the change that motivated it.

Two things I would genuinely reconsider on:

- If you would rather the derivation live in-file regardless, say so and I will add your one-liner — it is your review and this is a preference, not a correctness point.
- If the real worry is discoverability rather than its absence, a better fix than a comment might be making the constant self-describing (`FEE_DRIFT_CUSHION = 1.5`), which survives edits in a way a comment does not.

Either way it would ride the same follow-up, not this PR — a new commit here dismisses the approval and re-blocks the release.

Covers the identical re-posts at 14:41 and 14:47.

PR #1024, opened 2026-08-24
feat(polymarket): migrate predict-polymarket from subgraph to squid

OjusWiZard:
## What

Migrates the `predict-polymarket` data source from a graph-node subgraph (The Graph dialect) to the SQD/Subsquid indexer:

| | URL |
|---|---|
| Old | `https://api.subgraph.autonolas.tech/api/proxy/predict-polymarket` |
| **New** | `https://subgraph.autonolas.tech/squid/predict-polymarket/graphql` |

Omen (`predict-omen`) stays on The Graph and is untouched — `olas_agents_subgraph`, its `response_key: data:traderAgent`, and every non-Polymarket query are unchanged.

## Dialect changes

`first`/`skip` → `limit`/`offset`; `orderBy: X` + `orderDirection: desc` → `orderBy: X_DESC`; `traderAgent(id:)` → `traderAgentById(id:)`; `where: {rel_: {id:}}` → `where: {rel: {id_eq:}}`; `$id: ID!` → `$id: String!`.

Two OpenReader behaviours have no Graph equivalent and shaped the diff:

- **It rejects declared-but-unused variables.** `GET_POLYMARKET_SPECIFIC_BET_QUERY` declared `$betId` without using it (filtering is client-side). The Graph tolerated this; the squid returns `GRAPHQL_VALIDATION_FAILED`. The declaration is dropped; the caller still passes the value, which is ignored.
- **A nested list with no explicit limit returns every row**, where The Graph capped at 100. Nested selections now carry an explicit limit.
- **`orderBy` needs an explicit tiebreaker.** `blockTimestamp` is not unique per agent — several markets can be entered in one block — and OpenReader adds no implicit `id` ordering, so rows straddling a `limit`/`offset` page boundary could be returned twice or skipped. Both `marketParticipants` queries order by `[blockTimestamp_DESC, id_DESC]`.

The daily-profit query keeps its nested shape under `traderAgentById` rather than moving to a top-level `dailyProfitStatistics`, because `ApiSpecs.response_key` is fixed per model and a top-level response would not parse through `polymarket_agents_subgraph`.

## The one real behaviour change

`DailyProfitStatistic.profitParticipants` is now a list of conditionId strings where it was `Question` objects. Rather than rewrite the two consumers in `behaviours.py`, `_hydrate_profit_participants` re-fetches those questions in one batched pass and splices them back in, so the mech-fee day-matching keeps the shape it expects.

**That rehydration scopes bets to the agent.** `Question.bets` spans every tracked agent on the market. Unscoped, other traders' bets entered `_match_mech_requests_to_days` and consumed this agent's mech-request timestamps in the bisect. Measured against live data for one agent:

| | bets returned | belonging to other traders |
|---|---|---|
| unscoped | 11,891 | 10,880 (**91.5%**) |
| scoped (`bettorId_eq`) | 1,011 | 0 |

The address is lowercased inside the helper, so the invariant does not depend on the caller getting it right.

## Diagnostics

A resolved participant always carries at least one of the agent's own bets — verified 120/120 on live data — so an empty `bets` list is anomalous. Both failure modes are now counted and logged:

- conditionIds that resolve to no question
- resolved questions carrying no bets (the signature of a `bettorId` that stops matching, e.g. a casing regression: a checksummed address resolves every question but returns zero bets)

Without these, both degrade mech-fee attribution from bet-level to day-level silently, because the consumer treats an empty `bets` list identically to a missing key.

## Config

`polymarket_questions_subgraph` is a new `ApiSpecs` model: `response_key` is frozen per model and this query returns `data:questions`, not `data:traderAgentById`.

`service.yaml` also gains env-var overrides for the three `polymarket_*` subgraphs, so the next endpoint change is a deployment config change rather than a release.

> [!IMPORTANT]
> **This needs a package release.** The URLs in `aea-config.yaml` are hardcoded with no `${...}` override.

## Drive-by fix

The `GET /api/v1/agent/predictions` deep-pagination fallback built the Omen `PredictionsFetcher` unconditionally, so a Polymarket deployment queried the wrong subgraph with an Omen query. It now branches on `is_running_on_polymarket`, like `behaviours.py:1337` and the other handler already did. Pre-existing, unrelated to this migration, fixed here because it sits in the same code path.

The two Omen-branch overrides in `polymarket_trader/service.yaml` (`open_markets_subgraph`, `olas_agents_subgraph`) named a predict-polymarket URL while carrying Omen queries. They are dead code on this service, but now default to `predict-omen`.

## Unrelated: CI safety scan

`common_checks.yaml` also bumps `setuptools` in the safety tox venv past SFTY-20260721-58460 (CVE-2026-59890). This is not part of the migration: the safety scan started failing on every branch, the flagged package is seeded by the toolchain rather than declared by this repo, and extending the existing `commands_pre` override is the only fix that does not need a tomte bump. It is bundled here because it was blocking this PR's CI.

## Testing

- Every shipped query executed against the **live squid** with the exact variables its call sites pass. Real-data run: 167 daily stats, 986 unique conditionIds, 986/986 hydrated.
- 1392 tests pass. `queries.py` and `handlers.py` at 100% line coverage; `requests.py` at 97% (remaining misses are pre-existing untouched lines).
- New tests pin the dialect (an old-dialect regression fails loudly), multi-page pagination combined with hydration, the bettor scoping, the lowercasing, both diagnostic warnings, and the end-to-end contract between the hydration query's `blockTimestamp` selection and the consumer that reads it.
- `mypy`, `black`, `isort` (against tomte's config), `pylint`, and `autonomy packages lock --check` all clean.

## Expected data shift at cutover

The squid has more data than the retired deployment: v2-market token derivation yields more bets, post-cutover redemptions give non-zero recent `totalPayout`, and `dailyTradedSettled` is populated. `totalBets` / `totalPayout` moving upward after the switch is expected and correct.


packages/valory/skills/agent_performance_summary_abci/graph_tooling/queries.py
```diff
@@ -317,13 +317,12 @@
...
 GET_POLYMARKET_PREDICTION_HISTORY_QUERY = """
-query GetPolymarketPredictionHistory($id: ID!, $first: Int!, $skip: Int!) {
+query GetPolymarketPredictionHistory($id: String!, $first: Int!, $skip: Int!) {
   marketParticipants(
-    orderBy: blockTimestamp
-    orderDirection: desc
-    where: {traderAgent_: {id: $id}}
-    first: $first
-    skip: $skip
+    orderBy: blockTimestamp_DESC
```
Tanya-atatakai: Non-blocking: `blockTimestamp` isn't unique per agent (several markets can be entered in one block), and OpenReader adds no implicit `id` tiebreaker — so with `limit`/`offset` paging, rows straddling a page boundary can be returned twice or skipped. Suggest `orderBy: [blockTimestamp_DESC, id_DESC]` (multi-key orderBy verified working on this squid). The same ordering in `GET_POLYMARKET_TRADER_AGENT_BETS_QUERY` (line 359) is single-page, so there it only affects which rows fall past the 1000-row truncation — lower priority, but the same one-line fix applies.

.github/workflows/common_checks.yaml
```diff
@@ -116,20 +116,24 @@ jobs:
...
-        # in ``<26.1.2``) before ``safety check`` runs, without
-        # editing tomte's upstream tox config. ``tox-uv`` creates
-        # venvs without a bootstrap ``pip``, so ``ensurepip`` runs
-        # first to install one, then the upgrade lands a patched
-        # version. The ``+=`` form makes the intent explicit even
-        # though tomte's upstream ``[testenv:safety]`` currently
-        # declares neither key, so the override is only additive if
-        # a future tomte bump adds one on that section directly.
+        # and ``setuptools`` inside the venv past SFTY-20260601-69199
+        # (pip Path Traversal in ``<26.1.2``) and SFTY-20260721-58460
```
Tanya-atatakai, line 120: Non-blocking scope note: this safety-check maintenance is unrelated to the squid migration — fine if bundling toolchain fixes is repo practice, but worth a sentence in the PR description so the change isn't a surprise in the migration's blame history.

Tanya-atatakai (dismissed):
Reviewed against the migration plan (dialect table, config surfaces, query rewrites, `profitParticipants` shape change) with every uncertain construct verified against the live squid endpoint. The implementation is faithful to the plan and improves on it in three places:

- `service.yaml` gained env-overridable `POLYMARKET_*_SUBGRAPH_URL` params, so the next endpoint change is config, not a release.
- The hydration query scopes `Question.bets` to the querying agent (`bettorId_eq`) — the old subgraph shape let other agents' bets on the same market leak into mech-fee day-matching, so this is a small correctness improvement, not just parity. Verified `bettorId_eq` executes correctly on the live squid.
- The pre-existing `handlers.py` predictions-fallback bug (unconditional Omen fetcher) is fixed.

Also verified live: nested `dailyProfitStatistics(where/orderBy/limit/offset)` under `traderAgentById` works; the squid ignores the extra undeclared `betId` variable still sent by `polymarket_predictions_helper.py`, so leaving that file untouched is safe. `response_key` changes are correctly scoped to the polymarket connections only — the Omen `olas_agents_subgraph` blocks are untouched, and `_unwrap_trader_agent` handles both key names. `behaviours.py` reads only `metadata.title` + `bets[].blockTimestamp` from `profitParticipants`, so the hydrated shape is drop-in and the dropped circular `bets[].question.metadata.title` selection was never consumed.

Two non-blocking comments inline. One QA note: expect higher bet counts and non-zero recent `totalPayout` after cutover — the squid indexes v2-era markets and post-cutover redemptions the old deployment missed; that's the correction, not a regression.

DIvyaNautiyal07 (dismissed):
LGTM — approving. Findings below are all non-blocking (P2 + nits). The bettor-scoping fix, query dialect changes, and `is_running_on_polymarket` fallback branch are correctly implemented and would each be caught by the new tests on regression.

### P2

**1. `_unwrap_trader_agent` silently returns unknown-shape responses** — `packages/valory/skills/agent_performance_summary_abci/graph_tooling/requests.py:~380-391`. If neither `traderAgent` nor `traderAgentById` appears in the response dict, the helper returns the input unchanged. Callers then do `result.get("dailyProfitStatistics") or {}` on what is really the outer wrapper — get misses, loop breaks, empty list returned. If the squid ever renames the field again, this is the exact same silent-failure this PR series exists to prevent. Add a `logger.warning(f"unexpected response shape: {list(result.keys())}")` in the fallthrough branch.

**2. `_hydrate_profit_participants` would crash on a mixed str/dict `profitParticipants` list** — `requests.py:~539-544`. The splice does `participant in questions_by_id`. Dicts are unhashable, so a dict participant raises `TypeError`. Not reachable today (squid always returns `[String!]`), but `test_non_string_participants_are_ignored` only tests the all-dict early-return path — the mixed str+dict case where the fetch runs and the splice hits a dict is untested. If squid ever pre-hydrates partially, this crashes the whole `_fetch_daily_profit_statistics` cycle. Two fixes: (a) `if isinstance(participant, str) and participant in questions_by_id`, or (b) keep dicts as-is. Then extend the test.

### Nit

**3.** `_unwrap_trader_agent(result: Any) -> Any` — `requests.py:~380`. Every caller narrows to `Optional[Dict[str, Any]]` before invoking. Tighten to `Optional[Dict[str, Any]] -> Optional[Dict[str, Any]]`.

**4.** `_hydrate_profit_participants` docstring omits pre/post shape — `requests.py:~476-485`. Doesn't say `profitParticipants` is `[str]` conditionIds before this call and `[Dict]` after, and doesn't explain the `bettorId_eq` scoping.

**5.** `_unwrap_trader_agent` docstring omits the tie-break — `requests.py:~381-391`. `test_prefers_the_graph_key_when_both_present` pins that `traderAgent` wins when both keys are present; the docstring doesn't say so.

**6.** No behavioral test pins the `bets(...)` sub-selection shape — `test_polymarket_hydrated_shape_uses_bet_timestamps` at `tests/test_behaviours.py:1093` feeds pre-constructed dicts with `blockTimestamp` already set. If a field is added or removed inside the `bets(...)` block in `GET_POLYMARKET_QUESTIONS_BY_IDS_QUERY`, no test parses a real response and catches the drift. The query test pins `bets(limit: 1000, where: { bettorId_eq: $bettorId })` exactly so `limit`/`bettorId_eq` regressions are caught, but field additions inside the block are not.

**7.** Fix-narrative in test docstring — `tests/test_handlers.py:1136`. "Without the platform branch this path built the Omen fetcher on a Polymarket deployment..." describes the old bug rather than the invariant. Reword as "The deep-pagination fallback must route to the Polymarket fetcher when `is_running_on_polymarket=True`."

**8.** Inline comment above `GET_POLYMARKET_QUESTIONS_BY_IDS_QUERY` at `queries.py:~329-330`. "letting other traders' bets consume this agent's mech requests" — the actual risk is corrupted mech-fee-per-day attribution, not that mech requests get "consumed." The test name at `test_queries.py:744` phrases it correctly.

**9.** `result.get(key)` at `requests.py:390` — the enclosing `if key in result:` guarantees the key is present. Use `result[key]`.

**10.** `profitParticipants` is polymorphic (`List[str]` before hydration, `List[Dict]` after) — a `TypedDict` pair would encode the transformation and prevent passing a raw list where a hydrated one is expected. Larger than this PR's scope.

**11.** `TRADER_AGENT_RESPONSE_KEYS: Tuple[str, ...]` — `Tuple[Literal["traderAgent", "traderAgentById"], ...]` would catch a typo added later. Cosmetic.

DIvyaNautiyal07 (dismissed):

packages/valory/skills/agent_performance_summary_abci/graph_tooling/queries.py
Tanya-atatakai: Confirmed fixed in dff6e698 — tiebreaker applied to both `marketParticipants` queries (including the single-page one), hash cascade propagated. Approval stands.

.github/workflows/common_checks.yaml
Tanya-atatakai: Resolved — isolated into its own commit (9750f700) and explained in the PR description. Thanks.

iasonrovis (approved):

DIvyaNautiyal07 (approved):

PR #1025, opened 2026-08-24
fix(trader_abci): report honest Gnosis native deficit

DIvyaNautiyal07:
## Summary

- The `/funds-status` handler was folding the Safe's wxDAI balance into the xDAI balance before computing the native deficit that Pearl reads to trigger top-ups. On Gnosis every Safe-side `msg.value` path (offchain mech BalanceTracker deposit, on-chain mech request price) needs real xDAI, so a Safe with wxDAI but zero xDAI reported deficit=0 while it was actually unable to fund mech deposits, leaving the agent in a silent-failure loop with no top-up ever arriving from the master Safe.
- The Gnosis branch of `_get_adjusted_funds_status` now computes the native deficit against the native balance only.
- The Polygon USDC→POL smoothing is preserved: Polymarket's real Safe-side spending is USDC and no `msg.value` path drains native POL from the Safe (`execTransaction` gas is paid by the submitting EOA), so that branch does not hide a real shortfall.
- Adjusts the affected Gnosis test cases and adds one that reproduces the failure mode (0 native + ~30 wxDAI must surface the full topup deficit).

## Test plan

- [x] `uv run pytest packages/valory/skills/trader_abci/tests/` — 170/170 pass locally
- [x] `uv run pytest ... --cov=packages.valory.skills.trader_abci.handlers` — 100% coverage locally
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — OK locally
- [x] `tomte tox -p -e check-hash -e check-packages -e check-abci-docstrings -e check-dependencies -e check-doc-hashes` — OK locally
- [x] `autonomy push-all` done — updated package hashes published to IPFS registry
- [ ] After merge, on an affected agent: confirm `/funds-status` now returns a non-zero native deficit when the Safe holds wxDAI but zero xDAI, and Pearl triggers a native top-up from the master Safe

### DIFF
--- a/.github/workflows/common_checks.yaml
+++ b/.github/workflows/common_checks.yaml
@@ -133,7 +133,7 @@ jobs:
             -x 'testenv:safety.passenv+=NLTK_DISABLE_IMPORT_SECURITY' \
             -x 'testenv:safety.commands_pre+=python -m ensurepip --upgrade
                 python -m pip install --upgrade pip>=26.1.2' \
-            -x 'testenv:safety.commands=safety check --policy-file {toxinidir}/.safety-policy.yml'
+            -x 'testenv:safety.commands=safety check --policy-file .safety-policy.yml'
       - name: Bandit scan
         # ``!cancelled()`` runs bandit whether the safety scan above
         # succeeded or failed, but skips it when the run itself is
--- a/.github/workflows/common_checks.yaml
+++ b/.github/workflows/common_checks.yaml
@@ -115,21 +115,25 @@ jobs:
           # ``NLTK_DISABLE_IMPORT_SECURITY=1``.
           NLTK_DISABLE_IMPORT_SECURITY: "1"
         # The ``-x`` overrides (a) forward the escape-hatch env var
-        # into the safety tox venv, and (b) bootstrap+upgrade ``pip``
+        # into the safety tox venv, (b) bootstrap+upgrade ``pip``
         # inside the venv past SFTY-20260601-69199 (pip Path Traversal
         # in ``<26.1.2``) before ``safety check`` runs, without
-        # editing tomte's upstream tox config. ``tox-uv`` creates
-        # venvs without a bootstrap ``pip``, so ``ensurepip`` runs
-        # first to install one, then the upgrade lands a patched
-        # version. The ``+=`` form makes the intent explicit even
-        # though tomte's upstream ``[testenv:safety]`` currently
-        # declares neither key, so the override is only additive if
-        # a future tomte bump adds one on that section directly.
+        # editing tomte's upstream tox config, and (c) replace the
+        # policy file with the repo-level ``.safety-policy.yml`` so
+        # per-repo ignores for transitive deps we cannot pin land in
+        # the scan. ``tox-uv`` creates venvs without a bootstrap
+        # ``pip``, so ``ensurepip`` runs first to install one, then
+        # the upgrade lands a patched version. The ``+=`` form makes
+        # the intent explicit even though tomte's upstream
+        # ``[testenv:safety]`` currently declares neither key, so the
+        # override is only additive if a future tomte bump adds one
+        # on that section directly.
         run: |
           tomte tox -e safety \
             -x 'testenv:safety.passenv+=NLTK_DISABLE_IMPORT_SECURITY' \
             -x 'testenv:safety.commands_pre+=python -m ensurepip --upgrade
-                python -m pip install --upgrade pip>=26.1.2'
+                python -m pip install --upgrade pip>=26.1.2' \
+            -x 'testenv:safety.commands=safety check --policy-file {toxinidir}/.safety-policy.yml'
       - name: Bandit scan
         # ``!cancelled()`` runs bandit whether the safety scan above
         # succeeded or failed, but skips it when the run itself is
--- a/.safety-policy.yml
+++ b/.safety-policy.yml
@@ -1,0 +1,14 @@
+# Repo-level safety policy. Layered on top of tomte's shipped baseline
+# (see tomte/configs/safety-policy.yml). Add per-repo ignores here with
+# a specific PyUp ID, a reason, and an expiry so entries don't rot.
+
+security:
+  ignore-vulnerabilities:
+    # setuptools <83.0.0 (CVE-2026-59890, PyUp SFTY-20260721-58460).
+    # Pulled in transitively by tomte[safety]; not consumed by the
+    # trader's runtime image. Not pinnable from this repo — remove
+    # once tomte bumps the extra past the fix.
+    58460:
+      reason: "setuptools transitive of tomte[safety]; awaiting upstream bump"
+      expires: "2026-12-31"
+  continue-on-vulnerability-error: false
--- a/packages/valory/skills/trader_abci/handlers.py
+++ b/packages/valory/skills/trader_abci/handlers.py
@@ -406,11 +406,21 @@ class HttpHandler(BaseHttpHandler):
 
     def _get_adjusted_funds_status(self) -> FundRequirements:
         """
-        Adjust fund status based on chain-specific token equivalence:
+        Compute the fund status exposed to Pearl.
 
-        - Gnosis (Omen): treat wxDAI as xDAI (1:1, same decimals)
-        - Polygon (Polymarket): treat USDC as POL by converting via exchange rate
-          and fold USDC.e balance into the pUSD bucket (v2 primary collateral)
+        Chain-specific behaviour:
+
+        - Gnosis (Omen): no wxDAI-as-xDAI smoothing. Mech deposits (offchain
+          BalanceTracker and on-chain mech request price) are msg.value paths
+          that need real native xDAI from the Safe, so counting wxDAI as xDAI
+          would hide the shortfall and block Pearl's native top-up.
+        - Polygon (Polymarket): USDC is folded into POL via CoinGecko rate.
+          The Safe's real spending on Polymarket is USDC; no significant
+          msg.value path consumes native POL from the Safe (Safe
+          ``execTransaction`` gas is paid by the submitting EOA), so
+          smoothing avoids spurious POL top-up asks. USDC.e is also folded
+          into pUSD via :meth:`_merge_usdc_e_into_pusd` (v2 trading
+          collateral).
 
         :return: The adjusted fund requirements.
         """
@@ -424,6 +434,7 @@ class HttpHandler(BaseHttpHandler):
 
             native_status = safe_balances.tokens[chain_config["native_token_address"]]
 
+            adjustment_balance = 0
             if self.params.is_running_on_polymarket:
                 self._merge_usdc_e_into_pusd(safe_balances, chain_config)
                 # On Polygon: USDC balance needs to be converted to POL equivalent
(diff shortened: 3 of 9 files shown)


.safety-policy.yml
```diff
@@ -0,0 +1,14 @@
...
+# (see tomte/configs/safety-policy.yml). Add per-repo ignores here with
+# a specific PyUp ID, a reason, and an expiry so entries don't rot.
+
+security:
+  ignore-vulnerabilities:
+    # setuptools <83.0.0 (CVE-2026-59890, PyUp SFTY-20260721-58460).
+    # Pulled in transitively by tomte[safety]; not consumed by the
+    # trader's runtime image. Not pinnable from this repo — remove
+    # once tomte bumps the extra past the fix.
+    58460:
```
jmoreira-valory, line 11: 🟠 **F1 · HIGH — Safety ignore key `58460` never matches, so CI is red and every package check is skipped**

`linter_checks (ubuntu-24.04, 3.10)` fails at HEAD with exit code 64. The policy file *is* being picked up (`Scan configuration using a security policy file .safety-policy.yml`), but the scan then prints:

```
1 vulnerability reported
0 vulnerabilities ignored
-> Vulnerability found in setuptools version 79.0.1
   Vulnerability ID: SFTY-20260721-58460
```

Safety 3.7.0 matches `ignore-vulnerabilities` keys against the full vulnerability ID, and its own remediation text in that same log says so: *"To ignore this vulnerability, use PyUp vulnerability id SFTY-20260721-58460 in safety's ignore command-line argument or add the ignore to your safety policy file."* The bare numeric suffix `58460` matches nothing (and as written it is also parsed as a YAML integer key rather than a string).

This is not cosmetic. Because the Safety step fails, the four steps after it (`Check packages`, `Check hash`, `Code checks`, `AbciApp consistency checks`) are all `skipped`, so the CID cascade and the lint pipeline for this PR are entirely unverified by CI. A red CI blocks approval here regardless of the code being sound.

Minimal fix:

```suggestion
    SFTY-20260721-58460:
```

See the next finding for why I would rather see this file dropped altogether.

.github/workflows/common_checks.yaml
```diff
@@ -115,21 +115,25 @@ jobs:
...
+        # ``[testenv:safety]`` currently declares neither key, so the
+        # override is only additive if a future tomte bump adds one
+        # on that section directly.
         run: |
           tomte tox -e safety \
             -x 'testenv:safety.passenv+=NLTK_DISABLE_IMPORT_SECURITY' \
             -x 'testenv:safety.commands_pre+=python -m ensurepip --upgrade
-                python -m pip install --upgrade pip>=26.1.2'
+                python -m pip install --upgrade pip>=26.1.2' \
+            -x 'testenv:safety.commands=safety check --policy-file .safety-policy.yml'
```
jmoreira-valory, line 136: 🟡 **F2 · MEDIUM — Unrelated CI change duplicates and diverges from the working fix in open PR #1024**

Two of this PR's three commits (`044f5791`, `6a5153de`) are CI work with no connection to the funds-status fix, and they collide with open PR #1024, which touches the *same* file for the *same* advisory with a different approach:

```
-x 'testenv:safety.commands_pre+=python -m ensurepip --upgrade
    python -m pip install --upgrade pip>=26.1.2 setuptools>=83.0.0'
```

I checked #1024's run (job 97556358396, today at 19:07): it upgrades setuptools 79.0.1 -> 84.0.0 inside the safety venv, and the scan then reports **`0 vulnerabilities reported`** against tomte's stock policy file, with no repo-level policy needed. That approach is better than this one on two counts: it actually removes the vulnerable package instead of suppressing the finding, and it leaves tomte's canonical `--policy-file` path in place.

Suggested resolution: drop `.safety-policy.yml` and both CI commits from this PR and let #1024 land the CI fix, so this stays a focused one-behaviour change. If #1024 is not close to merging, copy its `setuptools>=83.0.0` bootstrap here instead of the ignore-policy machinery. Either way, whichever of the two merges first leaves the other with a conflict on this hunk.

packages/valory/skills/trader_abci/handlers.py
```diff
@@ -406,11 +406,21 @@ def _handle_get_static_file(
...
-        Adjust fund status based on chain-specific token equivalence:
-
-        - Gnosis (Omen): treat wxDAI as xDAI (1:1, same decimals)
-        - Polygon (Polymarket): treat USDC as POL by converting via exchange rate
-          and fold USDC.e balance into the pUSD bucket (v2 primary collateral)
+        Compute the fund status exposed to Pearl.
+
+        Chain-specific behaviour:
+
+        - Gnosis (Omen): no wxDAI-as-xDAI smoothing. Mech deposits (offchain
```
jmoreira-valory, line 413: 🔵 **F3 · LOW — Native bucket becomes a one-way funding pipe; confirm Pearl's top-up cadence is acceptable**

Worth an explicit sanity check on the Pearl side before this ships, because the change has a funding-flow consequence the PR body does not discuss.

On Gnosis the only funded Safe bucket is native: `FUND_REQUIREMENTS` gives wxDAI `topup: 0, threshold: 0`, so the master Safe can only ever send xDAI. Meanwhile native flows *out* of the Safe through three one-way paths (mech request price, BalanceTracker deposit, and the xDAI->wxDAI wrap in `bet_placement.py:_build_exchange_tx`), and nothing in the repo ever unwraps wxDAI back to xDAI. Winnings also come back as wxDAI.

So with the smoothing gone, native decays monotonically and every dip below the 1 xDAI threshold triggers a fresh 10 xDAI pull from the master Safe, no matter how much wxDAI has accumulated. That is the correct answer for mech deposits, which is the whole point of this PR. But it also means the accidental cap the smoothing used to impose on total funding is gone: an agent that keeps wrapping for bets will keep drawing down the master Safe.

I think this is intended and fine, since the demand is real. It is still a behaviour change beyond "report the deficit honestly", so please confirm Pearl caps or paces top-ups on its side, and consider a line in the PR body so whoever watches master-Safe balances after the rollout is not surprised.

.safety-policy.yml
```diff
@@ -0,0 +1,14 @@
+# Repo-level safety policy. Layered on top of tomte's shipped baseline
```
jmoreira-valory, line 1: 🔵 **F4 · LOW — "Layered on top of tomte's shipped baseline" is inaccurate; the override replaces it**

tomte v0.7.0's `[testenv:safety]` is:

```ini
commands =
    safety check --policy-file {envsitepackagesdir}/tomte/configs/safety-policy.yml
```

The new `-x 'testenv:safety.commands=...'` replaces that command outright, and `safety check` accepts exactly one `--policy-file`. There is no layering: this file *supersedes* tomte's, it does not extend it.

Today that costs nothing, because tomte's shipped baseline is `ignore-vulnerabilities: {}` (I checked v0.7.0). But the comment sets up a future maintainer to assume a fleet-wide ignore added in tomte would still apply here, when in fact it would be silently dropped. If the ignore-policy approach survives the finding above, reword lines 1-3 to say it replaces tomte's policy, and note that tomte's baseline entries have to be copied in by hand.

jmoreira-valory (changes requested):
### Review round 1 — ⚠️ Request changes

**What this does.** On Gnosis the /funds-status endpoint used to add the Safe's wxDAI balance into its xDAI balance before computing the native deficit Pearl reads. Because every Safe-side msg.value spend (legacy mech request price, offchain BalanceTracker deposit, and the xDAI->wxDAI wrap in bet placement) needs real native xDAI, a Safe holding wxDAI but no xDAI reported deficit=0 and never got topped up. This PR drops that smoothing so the Gnosis native deficit is computed against the native balance alone, keeps the Polygon USDC->POL smoothing, and rewrites the affected tests. It also bundles two unrelated CI commits that try to silence a setuptools advisory in the safety scan.

**The core fix is right, and I verified the premise independently rather than taking the PR body's word for it:**

- `trader_pearl/service.yaml` sets `auto_deposit_cap_per_cycle: 5000000000000000000` (5e18, 18 decimals -> native xDAI) while `polymarket_trader/service.yaml` sets `5000000` (5e6, 6 decimals -> USDC). That is direct config-level evidence that the Gnosis BalanceTracker deposit is a native `msg.value` path and the Polygon one is not, which is exactly the asymmetry the new docstring asserts.
- There is a *third* native drain the PR body does not mention, and it strengthens the case: `decision_maker_abci/behaviours/bet_placement.py:_build_exchange_tx` builds a `MultisendBatch(to=WXDAI, ..., value=self.w_xdai_deficit)` to wrap xDAI into wxDAI for a bet. That is Safe-side native spend too, and there is no reverse (unwrap) path anywhere in the repo, so wxDAI genuinely cannot substitute for xDAI.
- `FUND_REQUIREMENTS` gives the Gnosis wxDAI entry `topup: 0, threshold: 0`, so that entry existed purely to feed the smoothing. Removing the smoothing leaves it inert rather than breaking anything.

**Test coverage** is good: the three rewritten Gnosis cases plus the new 0-native / 29.32-wxDAI reproduction pin the exact production failure mode, and `test_gnosis_ignores_wxdai_balance_unknown` correctly re-purposes the old sub-call-revert case into a no-op assertion.

**Hash cascade** looks internally consistent (handlers.py + tests/test_handlers.py fingerprints -> trader_abci CID -> aea-config.yaml + packages.json -> agent CID -> both service.yaml files). Note that CI has **not** confirmed this: the `linter_checks` job aborts at the Safety step, so `Check packages`, `Check hash`, `Code checks` and `AbciApp consistency checks` all show `skipped` on this run. Nothing under `packages/` has actually been machine-verified yet.

**Findings** — 5 · 🟠 1 High · 🟡 1 Medium · 🔵 2 Low · ⚪ 1 Nit

| F1 | 🟠 HIGH | [.safety-policy.yml:11](https://github.com/valory-xyz/trader/blob/6a5153dee59d2d0818602038ebd79d7a90d29b05/.safety-policy.yml#L11) | **Safety ignore key `58460` never matches, so CI is red and every package check is skipped** |
| F2 | 🟡 MEDIUM | [.github/workflows/common_checks.yaml:136](https://github.com/valory-xyz/trader/blob/6a5153dee59d2d0818602038ebd79d7a90d29b05/.github/workflows/common_checks.yaml#L136) | **Unrelated CI change duplicates and diverges from the working fix in open PR #1024** |
| F3 | 🔵 LOW | [packages/valory/skills/trader_abci/handlers.py:413](https://github.com/valory-xyz/trader/blob/6a5153dee59d2d0818602038ebd79d7a90d29b05/packages/valory/skills/trader_abci/handlers.py#L413) | **Native bucket becomes a one-way funding pipe; confirm Pearl's top-up cadence is acceptable** |
| F4 | 🔵 LOW | [.safety-policy.yml:1](https://github.com/valory-xyz/trader/blob/6a5153dee59d2d0818602038ebd79d7a90d29b05/.safety-policy.yml#L1) | **"Layered on top of tomte's shipped baseline" is inaccurate; the override replaces it** |
| F5 | ⚪ NIT | `packages/valory/skills/trader_abci/handlers.py:191` | **`wrapped_native_address` is now written but never read** |

---
**Additional notes** (not anchored to a diff line):

- ⚪ **F5 · NIT — `wrapped_native_address` is now written but never read** The removed `else` branch was the only consumer of `chain_config["wrapped_native_address"]`. After this change both entries in `_get_chain_config` (lines 181 and 191) are write-only, and `GNOSIS_WRAPPED_NATIVE_ADDRESS` / `POLYGON_WRAPPED_NATIVE_ADDRESS` survive only to populate a dict nobody reads. I grepped the repo: the only other hits are the test fixtures.

Not worth blocking on, but dropping the two dict entries (keeping the constants, since the tests import them) would stop the next reader from assuming the Gnosis path still consults the wrapped balance.

DIvyaNautiyal07:
Closing in favour of the auto-unwrap approach discussed with Iason and team. New PR will unwrap wxDAI to xDAI before the offchain pre-deposit so Pearl doesn't need to trigger a native top-up in the wxDAI-rich edge case. Opening a fresh branch off main.

PR #1026, opened 2026-08-25
chore(trader): bump mech_interact_abci to v0.32.10

DIvyaNautiyal07:
## Summary

- Bumps `valory-xyz/mech-interact` from v0.32.9 to [v0.32.10](https://github.com/valory-xyz/mech-interact/releases/tag/v0.32.10). Picks up the wxDAI unwrap fallback: on chains where mech deposits use `msg.value` (Gnosis with `USE_OFFCHAIN=true`), a Safe that holds mostly wxDAI and little native now builds an atomic Safe multisend of `WrappedNative.withdraw` + `BalanceTracker.depositFor` instead of failing every offchain 402 with `BALANCE_SHORT`. Fixes the silent-failure loop for wxDAI-heavy Safes.
- Fixes a service.yaml wiring bug: `MECH_ANALYTICS_URL` and `USE_MECH_ANALYTICS` were being injected into the `trader_abci` skill overlay, but the code that reads them lives in `agent_performance_summary_abci`. Under `autonomy deploy`, the resolved values landed on the wrong skill's kwargs and the reader fell back to `skill.yaml` defaults (empty URL, `use_mech_analytics=False`). Result: `is_flag_enabled` always returned `False` and analytics stayed on the subgraph-only fallback even when Pearl set `USE_MECH_ANALYTICS=true`. This PR moves both params into a new `agent_performance_summary_abci` overlay in `trader_pearl/service.yaml`, `polymarket_trader/service.yaml`, and `aea-config.yaml` (for local `make run-agent`).
- CI: adds `setuptools>=83.0.0` to the safety scan venv's `commands_pre` alongside the existing pip upgrade so the scan clears `SFTY-20260721-58460` (transitive of `tomte[safety]`, not pinnable from this repo).

## Test plan

- [x] `uv run pytest packages/valory/skills/trader_abci/tests/ packages/valory/skills/agent_performance_summary_abci/tests/ -q` — 1355 pass locally.
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint` — OK locally.
- [x] `tomte tox -p -e check-hash -e check-packages -e check-abci-docstrings -e check-dependencies -e check-doc-hashes` — OK locally.
- [x] Safety scan green locally with the new setuptools upgrade.
- [x] `autonomy packages lock --check` OK; `autonomy push-all` — packages published to IPFS registry.
- [ ] On a wxDAI-heavy Gnosis Safe with `USE_OFFCHAIN=true`, verify the offchain deposit now lands via the unwrap+deposit multisend and Pearl reports the Safe as funded on the next cycle.
- [ ] With `USE_MECH_ANALYTICS=true` set via Pearl, verify HTTP calls to `mech-analytics-api.autonolas.tech` appear in the agent log (previously stayed on the subgraph fallback).


OjusWiZard (dismissed):

packages/valory/services/trader_pearl/service.yaml
```diff
@@ -344,6 +342,14 @@ models:
       url: ${MECHS_SUBGRAPH_URL:str:https://api.subgraph.autonolas.tech/api/proxy/marketplace-gnosis}
       delivery_rate_cap: ${DELIVERY_RATE_CAP:int:10000000000000000}
 ---
+public_id: valory/agent_performance_summary_abci:0.1.0
+type: skill
+models:
+  params:
+    args:
+      mech_analytics_url: ${MECH_ANALYTICS_URL:str:}
```
jmoreira-valory: 🟠 **F1 · HIGH — Analytics env vars now target an overlay no runtime code reads**

Every runtime reader of `use_mech_analytics` / `mech_analytics_url` executes in the **composed `trader_abci` context**, whose `TraderParams` is built from `trader_abci`'s `skill.yaml` args plus overrides that target `valory/trader_abci` — not from overrides targeting the abstract sub-skill. Evidence at HEAD:

- `trader_abci/behaviours.py:83` folds `*AgentPerformanceSummaryRoundBehaviour.behaviours` into the composed round behaviour, so those behaviours' `self.params` is the composed `TraderParams`.
- `agent_performance_summary_abci/behaviours.py` calls `is_flag_enabled(self.params)` (lines 317, 401) on the **same** `self.params` it reads `use_offchain` (line 578) and `mech_chain_id` (line 608) from — attributes that exist only on the composed `TraderParams` (via `MechInteractParams`), never on a standalone `AgentPerformanceSummaryParams`. So `self.params` provably is not the sub-skill's own params object.
- `agent_performance_summary_abci/graph_tooling/requests.py:107` casts `self.context.params` — again the composed context at runtime.
- `trader_abci/handlers.py` re-exports the sub-skill's handlers, so even the HTTP surface runs under `trader_abci`.
- `agent_performance_summary_abci` is `is_abstract: true`; nothing instantiates or reads its own params at runtime.

Net effect of this PR: the composed `TraderParams` now always falls back to `trader_abci/skill.yaml`'s defaults (`mech_analytics_url: ''`, `use_mech_analytics: false` — still there at lines 335-336), and `MECH_ANALYTICS_URL` / `USE_MECH_ANALYTICS` land on dead config. `is_flag_enabled` returns `False` unconditionally, in Pearl, Polymarket, and local runs alike — the exact silent-fallback bug the PR description says it fixes, now made permanent and un-enable-able.

The diff's own asymmetry is the tell: the sibling `AgentPerformanceSummaryParams` fields `is_agent_performance_summary_enabled`, `is_achievement_checker_enabled`, and `balance_tracker_address` are wired through the **same** `_ensure` mechanism and **stay** in the `trader_abci` overlay — and demonstrably work in production. If trader_abci-layer injection really landed "on the wrong skill's kwargs", those three would be equally broken.

**Fix:** restore `mech_analytics_url: ${MECH_ANALYTICS_URL:str:}` and `use_mech_analytics: ${USE_MECH_ANALYTICS:bool:false}` inside the `trader_abci` overlay of all three files (`aea-config.yaml`, `trader_pearl/service.yaml` line ~161, `polymarket_trader/service.yaml` line ~163), drop the three new `agent_performance_summary_abci` overlay blocks (dead config that will mislead the next reader), and re-run `autonomy packages lock`. If the flag genuinely didn't take effect before this PR, the root cause is elsewhere (e.g., Pearl not exporting `USE_MECH_ANALYTICS` into the deployment environment) and is worth re-diagnosing with the unchecked test-plan item — set `USE_MECH_ANALYTICS=true` and confirm `mech-analytics-api` HTTP calls appear in the agent log.

jmoreira-valory (changes requested):
### Review round 1 — ⚠️ Request changes

**What this does.** Bumps the third-party mech_interact_abci pin to mech-interact v0.32.10, which adds an atomic unwrap+deposit Safe multisend so wxDAI-heavy Gnosis Safes can fund offchain mech deposits instead of looping on BALANCE_SHORT. It also moves the MECH_ANALYTICS_URL / USE_MECH_ANALYTICS env-var overrides out of the trader_abci overlay into a new agent_performance_summary_abci overlay in aea-config.yaml and both service.yamls, and adds a setuptools upgrade to the CI safety-scan venv.

The mech-interact bump itself is solid and I verified it end-to-end:

- The new third-party CID `bafybeidufmi4…77ofm` exactly matches `skill/valory/mech_interact_abci/0.1.0` in mech-interact's `packages.json` at the `v0.32.10` tag, and the `v0.32.9...v0.32.10` delta is precisely the claimed unwrap fallback (`offchain_request.py` + tests + the same CI safety tweak; no `models.py`/params changes, so no new keys need re-declaring in the composed skill).
- The CID cascade is consistent across all six touched files, `upstream_pins` is updated, and CI is green (28/28).
- The fallback's newly load-bearing param `mech_wrapped_native_token_address` is correctly wired in all three config layers: WXDAI `0xe91D…a97d` for Gnosis and WPOL `0x0d50…1270` in `polymarket_trader/service.yaml`, so the unwrap path is actually live where it matters.

The blocker is the second bullet (the analytics env-var move) — evidence below suggests it inverts the bug it describes rather than fixing it. Notably, the one test-plan item that would falsify it ("USE_MECH_ANALYTICS=true via Pearl → analytics HTTP calls appear in the log") is unchecked.

**Findings** — 1 · 🟠 1 High

| F1 | 🟠 HIGH | [packages/valory/services/trader_pearl/service.yaml:350](https://github.com/valory-xyz/trader/blob/8447e63231ad2acc54d16c642629742f7f8d1f4f/packages/valory/services/trader_pearl/service.yaml#L350) | **Analytics env vars now target an overlay no runtime code reads** |

Tanya-atatakai (dismissed):

packages/valory/services/trader_pearl/service.yaml
DIvyaNautiyal07: You're right, I got the composition semantics wrong. Verified from the code: `TraderParams` inherits from `DecisionMakerParams` which inherits from `AgentPerformanceSummaryParams` (`decision_maker_abci/models.py:397-398`), so `self._ensure("use_mech_analytics", kwargs, bool)` at `AgentPerformanceSummaryParams.__init__` reads from the composed `TraderParams` kwargs. Overrides targeting the abstract sub-skill are dead config.

The clinching evidence you named is decisive: the sibling params (`is_agent_performance_summary_enabled`, `is_achievement_checker_enabled`, `balance_tracker_address`) live in the same params class and work in production via the `trader_abci` overlay. If my diagnosis had been right, all three would be equally broken. They aren't.

Reverted in f40e54e0d. Restored `mech_analytics_url` and `use_mech_analytics` inside the `trader_abci` overlay of `trader_pearl/service.yaml`, `polymarket_trader/service.yaml`, and `aea-config.yaml`. Dropped the three new `agent_performance_summary_abci` overlay blocks. Hashes re-locked and pushed.

The colleague-flagged symptom (analytics staying on the subgraph fallback) needs re-diagnosing — likely Pearl not exporting `USE_MECH_ANALYTICS=true` into the deployment env, or something else in the deployment layer. I'll follow up separately with the test-plan item you called out.

jmoreira-valory (approved):
### Review round 2 — ✅ Approve

**What this does.** Bumps the third-party mech_interact_abci pin to mech-interact v0.32.10, which lets wxDAI-heavy Gnosis Safes fund offchain mech deposits via an atomic unwrap+deposit Safe multisend instead of looping on BALANCE_SHORT. Since my last review, the author reverted the analytics env-var move I blocked on: MECH_ANALYTICS_URL / USE_MECH_ANALYTICS are back in the composed trader_abci overlay (the one the runtime actually reads) in aea-config.yaml and both service.yamls, with the dead abstract agent_performance_summary_abci overlays removed. Also keeps the CI safety-scan setuptools upgrade.

My round-1 blocker (F1, HIGH — analytics overrides targeting an overlay no runtime code reads) is fully addressed at `f40e54e0`. Verified at HEAD:

- `mech_analytics_url` / `use_mech_analytics` now sit inside the `valory/trader_abci:0.1.0` overlay in all three config layers (`aea-config.yaml` L285-286, `trader_pearl/service.yaml` L164-165, `polymarket_trader/service.yaml` L166-167), and the abstract `agent_performance_summary_abci` overlay is gone from all three.
- `trader_abci/skill.yaml` declares both keys with defaults, so the override is legal and feeds the composed `TraderParams` kwargs that `AgentPerformanceSummaryParams._ensure` reads.
- The agent/service CID rebump cascades consistently (packages.json + both service.yaml agent pins), and CI is green 28/28 including `check-hash` / `check-packages`.
- The mech-interact v0.32.10 bump is unchanged since round 1, where I verified the CID against the release tag and the wrapped-native-token wiring end-to-end.

Credit where due: the author's in-thread reply verified the composition semantics from source (`TraderParams` → `DecisionMakerParams` → `AgentPerformanceSummaryParams` MRO) rather than just applying the suggestion.

One non-blocking note: the two live-verification test-plan items remain unchecked — in particular "USE_MECH_ANALYTICS=true via Pearl → analytics HTTP calls appear in the log", which is the runtime check that would confirm the wiring on a deployed service. Worth ticking during the next Pearl staging cycle.

**No findings.**

Tanya-atatakai (approved):

OjusWiZard (approved):

PR #1029, opened 2026-08-27
fix(trader): guard synced_timestamp against a Tendermint period reset (OPE-1907)

OjusWiZard:
> [!NOTE]
> **#1032 is stacked on this PR** ([OPE-1914](https://linear.app/valory-xyz/issue/OPE-1914) — Polystrat restart loop and stale Profile). It targets this branch, so **this one merges first**; #1032 then retargets to `main` automatically.
>
> This branch is currently **2 commits behind `main`** (missing `13d70beef` and merge `ea841f7a3`), which is what makes it show `CONFLICTING`. That staleness propagates to #1032: it drags the stack back before the tomte 0.7.2 / analyse-service work, so `analyse-service` reports 5 findings there that `main` already fixed. Merging `main` into this branch clears both.
>
> Worth noting for reviewers of both: the `handlers.py` guard here is exactly what unblocks OPE-1914's deferred fix 6 (un-latching the HTTP 425 during Tendermint recovery), which was previously unsafe because the un-latched request fell straight into this `ValueError`.

---

Fixes the fatal half of [OPE-1907](https://linear.app/valory-xyz/issue/OPE-1907) — items 6 and 7 of the ticket's Trader work. The middleware half (items 1–5) ships separately in [olas-operate-middleware#472](https://github.com/valory-xyz/olas-operate-middleware/pull/472).

## The bug

A Tendermint period reset can wipe the round-transition history while a behaviour is mid-flight. Polystrat hit this while `PolymarketBetPlacementBehaviour` was awaiting a CLOB response — the reset landed between the request and the `matched` reply:

| time | event |
| --- | --- |
| 12:13:20.234 | `place_bet` sent |
| 12:13:20.237 | `begin_block` deadline expired, reset triggered |
| 12:13:23.278 | CLOB responds `"status": "matched"` |
| 12:13:23.280 | `Resetting tendermint node at end of period=0`, transition history wiped |
| 12:14:24.535 | behaviour resumes, reads `synced_timestamp`, process dies |

`synced_timestamp` reads `RoundSequence.last_round_transition_timestamp`, which raises `ValueError` until the next transition completes. Uncaught, it surfaced as an `AEAActException` and killed the agent.

The bet had already settled on-chain (Polygon `0x174d1225…cb19`, block 92691851), but the crash happened before `update_bet_transaction_information` could write `processed_timestamp` — so the bet still looked unprocessed and was eligible to be placed again.

## The fix

Guard the read in every dev-owned skill that defines the property, falling back to the local clock with a warning and keeping the `int` return type:

| Skill | Property owner | Notable callers |
| --- | --- | --- |
| `decision_maker_abci/behaviours/base.py:213` | `DecisionMakerBaseBehaviour` | the crash site (`:432`, `:452`), sampling, redeem, decision_receive, blacklisting |
| `staking_abci/behaviours.py:115` | `StakingInteractBaseBehaviour` | `is_checkpoint_reached` |
| `check_stop_trading_abci/rounds.py:131` | `CheckStopTradingRound` | `should_review_bets` |
| `agent_performance_summary_abci/models.py:393` | `SharedState` | ~10 in that skill's `behaviours.py` |

### Why guard the property instead of returning `Optional[int]`

`trader_abci/handlers.py:588` already guards this exact `ValueError` by falling back to `None` and skipping a cache check. Propagating that shape would push a `None` branch into ~12 call sites that all want "approximately now" — including arithmetic ones like `next_checkpoint <= synced_timestamp` and `synced_timestamp - last_review`.

Wall clock is a correct answer for all of them here: `processed_timestamp` goes to the local bets store rather than the synced db (as the comment at `base.py:445` notes), the staking and review comparisons are second-granularity, and both services are single-agent sovereign deployments (`number_of_agents: 1`), so there is no peer whose consensus clock this could diverge from. The warning keeps the substitution visible.

### Scope notes

- `mech_interact_abci/models.py:523` defines the same unguarded property, but is shadowed in the composed shared-state MRO (`trader_abci → decision_maker_abci → chatui_abci → agent_performance_summary_abci → mech_interact_abci`) by the `agent_performance_summary_abci` one. Unreachable here; left to its upstream repo.
- `abstract_round_abci/behaviours.py:225` is also unguarded, also third-party, and only reachable with slashing enabled. The other two framework call sites (`handlers.py:206`, `behaviour_utils.py:1920`) are already guarded upstream.
- Item 8 of the ticket — persisting matched orders across a period reset so they are not re-bet — is **not** in this PR. This change stops the crash; it does not close the window where a behaviour dies before `update_bet_transaction_information` runs at all.

### Known tradeoff

During the fallback window, `is_checkpoint_reached` and `should_review_bets` compare against local time, so a host clock running ahead could trigger a checkpoint early. The window is seconds — between the reset and the next transition — and the staking contract reverts a premature checkpoint.

## Tests

One fallback test per skill, plus `test_update_bet_transaction_information_after_period_reset`, which drives the actual crash site: it forces the `ValueError` and asserts `processed_timestamp` lands as an int, `update_investments` is called, and `store_bets()` runs.

3045 tests pass. 100% coverage on `decision_maker_abci/behaviours/base.py`, `staking_abci/behaviours.py`, `check_stop_trading_abci/rounds.py` and `trader_abci/handlers.py`; `agent_performance_summary_abci/models.py` holds at its pre-existing 86% with the changed lines covered. `black`, `isort`, `flake8`, `mypy`, `pylint`, copyright, ABCI docstring/spec, `check-hash` and `check-packages` all pass. Hashes re-locked — the propagation touches `chatui_abci`, `tx_settlement_multiplexer_abci`, `trader_abci`, the agent config and both service configs.


packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -212,7 +212,14 @@ def synchronized_data(self) -> SynchronizedData:
     @property
     def synced_timestamp(self) -> int:
         """Return the synchronized timestamp across the agents."""
```
DIvyaNautiyal07, line 214: The docstring still reads `"""Return the synchronized timestamp across the agents."""` — after the fix, the fallback path returns the **local** clock, not a synchronized value. This matters for consumers doing agent-vs-agent comparisons (the PR body's own "known tradeoff" note on `is_checkpoint_reached`). A caller reading only the docstring gets no signal that the return may be non-consensus.

Applies identically to the other three sites:
- `staking_abci/behaviours.py:116`
- `check_stop_trading_abci/rounds.py:133`
- `agent_performance_summary_abci/models.py:395`

```suggestion
    def synced_timestamp(self) -> int:
        """Return the synchronized timestamp; falls back to local clock on a Tendermint period reset (see warning log)."""
```

DIvyaNautiyal07 (commented):
### Cross-file findings

**P1 — `market_manager_abci/graph_tooling/requests.py:117-120` has the same unguarded read.** The fix covers the four `synced_timestamp` properties, but a structurally identical `synced_time` property in `MarketManagerBaseBehaviour` does `int(self.shared_state.round_sequence.last_round_transition_timestamp.timestamp())` with no guard. Live consumer: `_fetch_bets` uses it (`self.synced_time + self.params.opening_margin`). Same failure class as the crash this PR fixes — a Tendermint reset during `FetchBetsBehaviour` still raises `ValueError` → `AEAActException` and kills the agent.

Suggested fix in that file:
```python
@property
def synced_time(self) -> int:
    """Get the synchronized time among agents."""
    try:
        return int(self.shared_state.round_sequence.last_round_transition_timestamp.timestamp())
    except ValueError as e:
        self.context.logger.warning(
            "No round transition has completed yet, e.g. after a Tendermint "
            f"period reset: {e}. Falling back to the local clock."
        )
        return int(datetime.now(timezone.utc).timestamp())
```

**Nit — `decision_maker_abci/behaviours/blacklisting.py:46-50` defines a dead `synced_time` property.** It's never called in that file (`_blacklist` at line 93 uses `self.synced_timestamp` from `base.py`, which this PR guarded). Either remove it or apply the same guard — leaving it defined-but-unguarded is a trap for a future caller.

**Nit — PR body cites wrong file:line numbers.** `reedem.py:956-964` should be `:969` (the actual `update_bet_transaction_information()` call); `polymarket_bet_placement.py:148,156` should be `:181` and `:189`. Consider citing function names instead of line numbers to reduce rot.

**Nit — PR body says `handlers.py:588` "already guards this exact `ValueError`".** That guard is `except Exception` (broader) and it protects a different call path — the framework's synced_timestamp, not any of the four patched sites. Not a matched precedent; more precise would be "handles this class of failure via `except Exception`".

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -212,7 +212,14 @@ def synchronized_data(self) -> SynchronizedData:
...
-        return int(self.round_sequence.last_round_transition_timestamp.timestamp())
+        try:
+            return int(self.round_sequence.last_round_transition_timestamp.timestamp())
+        except ValueError as e:
+            self.context.logger.warning(
+                "No round transition has completed yet, e.g. after a Tendermint "
```
DIvyaNautiyal07, line 219: The warning message string is duplicated verbatim across all four patched sites (`decision_maker_abci/behaviours/base.py:219`, `staking_abci/behaviours.py:121`, `check_stop_trading_abci/rounds.py:140`, `agent_performance_summary_abci/models.py:402`).

The string embeds three independent facts: the reason ("no round transition"), the example trigger ("Tendermint period reset"), and the fallback strategy ("local clock"). Any one changing without updating all four leaves the code and the log message disagreeing — grep is the only safety net.

Either extract to a module-level constant per skill, or accept the duplication and pin it with a test that greps for the exact phrase across the four sites. A shared cross-skill utility would violate the skill-boundary discipline the repo maintains, so a shared home isn't the right fix here.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_base.py
```diff
@@ -540,6 +540,31 @@ def test_synced_timestamp(self) -> None:
         result = behaviour.synced_timestamp
         assert result == 1700000000
 
+    def test_synced_timestamp_falls_back_to_local_clock(self) -> None:
```
DIvyaNautiyal07, line 543: The four fallback tests all patch `datetime` and pin `mock_datetime.now.return_value.timestamp.return_value = 1700000000.5`, then assert `result == 1700000000`. A mutation that returns any hardcoded value close to 1.7e9 (or that drops `timezone.utc` and gets a naive datetime whose `.timestamp()` still computes) can slip through.

One of the four tests would be stronger with a real-clock proximity check — no `datetime` mock, seed the ValueError, assert the returned value is within a few seconds of `time.time()`:

```python
import time

type(round_sequence).last_round_transition_timestamp = PropertyMock(
    side_effect=ValueError("no transition yet")
)
before = int(time.time())
result = behaviour.synced_timestamp
after = int(time.time())
assert before <= result <= after + 1
```

This is the only assertion that would catch a mutation like `return 0` or one that swaps `datetime.now(timezone.utc)` for a hardcoded constant.

Adamantios (dismissed):
LGTM though it would be better to completely remove this synced timestamp logic instead and simply use the `time` module as it does not make sense for a single-agent service. It unnecessarily complicates the code.

DIvyaNautiyal07 (dismissed):

packages/valory/skills/decision_maker_abci/handlers.py
```diff
@@ -388,9 +388,16 @@ def _handle_get_health(
         is_tm_unhealthy = round_sequence.block_stall_deadline_expired
 
         current_time = datetime.now().timestamp()
-        seconds_since_last_transition = current_time - datetime.timestamp(
-            round_sequence.last_round_transition_timestamp
-        )
+        try:
```
jmoreira-valory, line 391: 🔵 **F1 · LOW — New `_handle_get_health` guard is unreachable; its test asserts a state the FSM cannot produce**

`_handle_get_health` already bails out at line 375 via `_has_transitioned()`, which catches `ValueError` from `last_round_transition_height`. In open-autonomy v0.21.26 (the pinned version) those two properties move in lockstep — `RoundSequence._reset_to_default_params()` clears both (`abstract_round_abci/base.py:3814-3815`) and `_update_round` sets both (`:3801-3802`), and those are the only assignments. So `last_round_transition_height` raises **iff** `last_round_transition_timestamp` raises. `_handle_get_health` is a plain handler with no yield points, and ABCI messages are dispatched on the same agent message loop, so nothing can interleave between line 375 and line 393.

Net effect: after a real period reset `/healthcheck` returns **425 Too Early** from `_handle_too_early`, never a 200 carrying `seconds_since_last_transition: 0.0`. The new `except ValueError` branch cannot be reached in production.

That matters mainly because of the test. `test_handle_get_health_after_period_reset` only gets there by patching `_has_transitioned` to `True`, and then asserts `is_healthy is True` under the docstring "stays healthy when the transition history was wiped". It documents a post-reset contract the agent does not actually honour — and `/healthcheck` is precisely what the middleware half of OPE-1907 (olas-operate-middleware#472) consumes, so a reader coordinating the two halves is being told the wrong thing.

Keeping the guard as belt-and-braces is fine and costs nothing. The fix is to stop the test claiming otherwise: either rename it and its docstring to say it exercises the defensive branch in isolation, or assert what the endpoint really does after a reset (`_has_transitioned()` False, `status_code == 425`).

packages/valory/skills/market_manager_abci/graph_tooling/requests.py
```diff
@@ -116,7 +117,16 @@ def synchronized_data(self) -> SynchronizedData:
     @property
     def synced_time(self) -> int:
         """Get the synchronized time among agents."""
-        synced_time = self.shared_state.round_sequence.last_round_transition_timestamp
+        try:
+            synced_time = (
+                self.shared_state.round_sequence.last_round_transition_timestamp
+            )
+        except ValueError as e:
+            self.context.logger.warning(
```
jmoreira-valory, line 125: ⚪ **F2 · NIT — Fallback warns once per read and makes `synced_time` non-constant within a round**

The warning fires on every property access, and several call sites read the property inside loops: `_requeue_bets_for_selling` reads it **twice per bet** (`behaviours/update_bets.py:62,65`), `_blacklist_expired_bets` once per bet (`:81`), `behaviours/polymarket_fetch_market.py:129,132,152` repeats the shape, and `reedem.py:150` evaluates `self.synced_timestamp` once per trade inside a generator expression. The window is not short: this PR's own incident timeline has the reset at 12:13:20 and the read still raising at 12:14:24, so it lasted over a minute. With a few hundred tracked bets that is O(bets) identical WARNING lines per iteration in logs Pearl users ship back — enough noise to bury the one line that actually explains the substitution.

Second-order: before this change `synced_time` was a fixed value for the whole period, so repeated reads were interchangeable. On the fallback path each read returns a fresh clock, so `time_since_last_sell_check` and `is_ready_to_sell(self.synced_time, ...)` in the same loop iteration now compare against two slightly different "now"s. Sub-second drift, harmless here, but it quietly removes an invariant a future caller might lean on.

Both go away cheaply: hoist the read to a single local per method (`now = self.synced_time` before the loop), and log the fallback once per episode — e.g. a flag on the shared state cleared when a transition completes — rather than once per access.

jmoreira-valory (dismissed):

**What this does.** Reading `RoundSequence.last_round_transition_timestamp` raises `ValueError` for as long as Tendermint's transition history has just been wiped, and that killed a live Polymarket agent that resumed mid-bet after a period reset. This PR wraps every dev-owned `synced_timestamp` / `synced_time` property (decision_maker, staking, check_stop_trading, agent_performance_summary, market_manager) and the `/healthcheck` handler in a try/except that falls back to the local wall clock, logs a warning, and keeps the `int` return type so no caller has to grow a `None` branch. It also deletes a dead duplicate `synced_time` property from the blacklisting behaviour.

Verified at HEAD rather than taken from the PR body:

- **Coverage is complete.** Grepping every non-test `.py` in the eight dev skills at `b5e66da` leaves exactly six reads of `last_round_transition_timestamp`, and all six are inside the new guards. Nothing dev-owned is left unguarded.
- **The fallback is epoch-correct.** `Header.timestamp` is built with `datetime.fromtimestamp(seconds + nanos)` (no `tz`), i.e. a naive *local* datetime, so `int(ts.timestamp())` round-trips to true epoch seconds — the same quantity `int(datetime.now(timezone.utc).timestamp())` produces. Swapping between the two paths introduces no UTC-offset jump on a non-UTC host, which is the obvious way this kind of fix goes wrong.
- **The single-agent premise holds.** `number_of_agents: 1` confirmed in both `trader_pearl` and `polymarket_trader` service.yaml at HEAD, so the local-clock substitution cannot desync a peer's payload hash.
- **Dead-code removal is safe.** `BlacklistingBehaviour._blacklist` and `async_act` use `synced_timestamp`, never the removed `synced_time`; no other reference to it survives in the skill.
- CI is green (30/30 checks). All three prior inline threads are resolved and the cross-file P1 landed in `865b463a`.

**Merge state:** GitHub reports `mergeable_state: dirty` — the branch conflicts with `main` (base `196aea9`, `main` now at `ea841f7`) despite the merge commit at HEAD. The conflicts will land in `packages.json` and the `skill.yaml` hash blocks; please re-run `autonomy packages lock` after resolving rather than hand-picking CIDs, or `check-hash` will fail on the drift.

**Findings** — 2 · 🔵 1 Low · ⚪ 1 Nit

| F1 | 🔵 LOW | [packages/valory/skills/decision_maker_abci/handlers.py:391](https://github.com/valory-xyz/trader/blob/b5e66da8b6fca2016d1bff73e046353fe7dbc447/packages/valory/skills/decision_maker_abci/handlers.py#L391) | **New `_handle_get_health` guard is unreachable; its test asserts a state the FSM cannot produce** |
| F2 | ⚪ NIT | [packages/valory/skills/market_manager_abci/graph_tooling/requests.py:125](https://github.com/valory-xyz/trader/blob/b5e66da8b6fca2016d1bff73e046353fe7dbc447/packages/valory/skills/market_manager_abci/graph_tooling/requests.py#L125) | **Fallback warns once per read and makes `synced_time` non-constant within a round** |

iasonrovis (approved):

jmoreira-valory (approved):

**What this does.** Guards every dev-owned `synced_timestamp` / `synced_time` property (decision_maker, staking, check_stop_trading, agent_performance_summary, market_manager) plus the `/healthcheck` handler against the `ValueError` that `RoundSequence.last_round_transition_timestamp` raises after a Tendermint period reset wipes the transition history, falling back to the local UTC clock with a warning. This is the crash that killed a live Polymarket agent mid-bet (OPE-1907) and left a settled bet looking unprocessed and re-bettable.

Re-approval at `dcfff333` — my prior approval at `b5e66da8` was auto-dismissed by the push. The delta is exclusively a merge of `main`: `13d70bee` (tomte v0.7.2 + open-autonomy 0.21.28 + analyse-service re-enable, landed on `main` via #1031) and the merge commits, plus the resulting CID re-lock. Verified via the compare API that no PR-owned source file changed since my last review, the re-locked hashes are consistent across `packages.json` / `aea-config.yaml` / skill.yamls / service.yamls (and mechanically confirmed by the green `check-hash` / `check-packages` jobs), and the merge cleared the previous `CONFLICTING` state (`mergeable: true`).

CI is fully green at HEAD (28/28 check runs). All five review threads are resolved, @DIvyaNautiyal07's P1 (`market_manager` `synced_time`) remains covered by the guard in `graph_tooling/requests.py`, and my two prior non-blocking notes (unreachable `_handle_get_health` guard given `_has_transitioned()`, per-read warning noise in loops) stand as resolved threads and are not re-raised.

**No findings.**

PR #1031, opened 2026-08-28
chore: bump tomte to v0.7.2 and open-autonomy to 0.21.28, enable analyse-service

OjusWiZard:
## Summary

Re-enables the `analyse-service` check in CI, together with the two upstream bumps that make it able to run and run deterministically. These only work as a set: uncommenting on its own would reproduce the old failures, because the fixes arrive with the pins.

The comment in `common_checks.yaml` said the check "does not work properly as it is checking funds manager skill which is not an ABCI skill". That described a symptom. The cause was `_get_chained_abci_skill` iterating an unordered set, so `funds_manager` and `trader_abci` won at random — whoever wrote the comment saw `funds_manager` win twice.

## The two bumps

**tomte v0.7.0 → v0.7.2.** v0.7.0 rendered a single `autonomy analyse service --public-id $SERVICE_PUBLIC_ID`. That variable is empty in a repo with more than one service, so tox ran a bare `--public-id --skip-warnings` and the check could not execute at all. v0.7.2 renders one command per service.

**open-autonomy 0.21.26 → 0.21.28.** `_get_chained_abci_skill` now iterates `sorted(agent_config.skills, key=str)` and selects the skill no other declares as a dependency. It resolves `valory/trader_abci` every run.

Both move in *every* place they appear, or the driver and the rendered tox config disagree — tomte on 9 lines (2 in `pyproject.toml`, 7 CI install lines), open-autonomy on 7 (`pyproject.toml`, `tox.ini` Authorized Packages, and 5 in `release.yaml`).

## The findings: seven, not five

`validate_agent_override_env_vars` raises on the first failing component and aborts before the service-level validator runs, so fixing the first five revealed two more.

| Finding | Was | Now |
|---|---|---|
| 4 × Polymarket / Polygon subgraph `url` | bare literal | `${<NAME>:str:<same literal>}` |
| `polymarket_client.host` / `.chain_id` | bare literal | `${POLYMARKET_CLOB_HOST:str:…}` / `${POLYMARKET_CHAIN_ID:int:137}` |
| `params.mech_analytics_url` | `${MECH_ANALYTICS_URL:str:}` | `${MECH_ANALYTICS_URL:str:null}` + `Optional[str]` |

**Why named, and why it is safe.** All four subgraph names already exist in `service.yaml`, so this completes the "one name end-to-end" alignment from #982 rather than inventing knobs. Service overrides *require* a name (`validate_env_var_name=True`); agent overrides do not.

This is the mirror image of the over-naming bug in `525f924c2`. There, the value was anonymous, the middleware was feeding it through the path key, and naming it killed that channel. Here the values were **bare literals** — no channel was live, and the new defaults are byte-identical, so nothing that resolves today changes. Verified by resolving the real config with open-aea's resolver:

| form | no env | path key set | bare name set |
|---|---|---|---|
| named | old literal | old literal | override wins |
| anonymous | old literal | override wins | ignored |

The three runtime-computed `setup.*` params reverted in `8a33cca95` are untouched and still anonymous; a deploy build confirms they still resolve via their path keys.

**`mech_analytics_url` was a live bug, not just a lint finding.** The empty tail in `${MECH_ANALYTICS_URL:str:}` makes aea's `ENV_VARIABLE_RE` backtrack and produce the literal string `'str:'`. A deployment build emits `MECH_ANALYTICS_URL=str:` into `agent_0.env`. Being truthy, it silently defeated the deliberate guard from the feature's own commit ("Flag on with URL unset throws rather than silently returning zero, which would zero the cost leg and inflate ROI"). The grammar has no empty-string default, so the config passes `null` and the model normalises `None` to `""`.

## Also included

- **`release.yaml`** — `AUTHOR` came from `grep 'service'`, which matches the `service_staking_token` *contract* before any service and returned `valory` by luck. Now reads the first `dev` key from `packages.json`, fails loudly when empty, and quotes the expansion. Verified inside `valory/open-autonomy-user:0.21.28` with the workspace mounted as the action mounts it.
- **`CLAUDE.md`** — adds a Repo Gotchas section and corrects two check descriptions that were wrong. The `make` targets it documented cannot run at all: they invoke bare `tox`, but this repo's `tox.ini` defines no environments — tomte renders them.

## Test plan

- [x] `analyse-service` clean on **both** services — "Service is ready to be deployed."
- [x] `check-hash`, `check-packages`, `check-doc-hashes`, `check-dependencies`
- [x] `check-abci-docstrings`, `check-abciapp-specs`, `check-handlers`
- [x] `tomte check-copyright --author valory`
- [x] `black-check`, `isort-check`, `flake8`, `mypy`, `pylint`
- [x] 1230 tests pass; `models.py` coverage identical to baseline (254 stmts / 34 miss, same missing-line set, confirmed by stashing)
- [x] `test_agent_config_resolution.py` — the regression test guarding named-vs-anonymous resolution
- [x] Deployment build of `polymarket_trader` to confirm resolved env values end-to-end
- [x] **End-to-end config resolution against the real Pearl service configs**, for both flavors, under **open-autonomy 0.21.24** — the version the installed middleware actually runs, not the 0.21.28 pinned here (see below)

## End-to-end validation against Pearl

Built the agent config through the middleware's own `ServiceBuilder`
(`try_update_abci_connection_params` → `try_update_runtime_params` → `generate_agent`) against the real `~/.operate` service configs, on this branch and on `13d70beef~1`, then diffed. Run in **`host` deployment mode**, which is what Pearl uses on Linux — the service dirs contain `agent.json` and `tendermint.json`, not compose output — and under **open-autonomy 0.21.24**, the version the installed middleware runs rather than the 0.21.28 pinned here. Read-only; throwaway key; nothing under `~/.operate/services` was modified.

**The consensus assertion (the PR #330 / `03d2cd64` regression class) passes on both.** Naming a field the middleware rewrites at runtime would silently leave the agent on its YAML default, which no static check catches:

```
trader_pearl        SETUP_ALL_PARTICIPANTS      = ["0x3767E7…bbb77"]   real, not 0x0000…
                    SETUP_SAFE_CONTRACT_ADDRESS = 0x3F06E0…1bfa2
polymarket_trader   SETUP_ALL_PARTICIPANTS      = ["0x4C7139…2b06"]    real, not 0x0000…
                    SETUP_SAFE_CONTRACT_ADDRESS = 0x28C51c…97E6
```

All six fields the middleware rewrites — the three `setup.*`, the two tendermint URLs, and `on_chain_service_id` (which the middleware adds on top of open-autonomy's five) — remain anonymous in the agent config and are untouched here.

**Full diff of the generated agent env,** with the per-build random container-name suffix normalised:

| | trader_pearl | polymarket_trader |
|---|---|---|
| keys | 559 → 559 | 616 → **618** |
| added | none | `POLYMARKET_CLOB_HOST`, `POLYMARKET_CHAIN_ID` |
| changed | `AEA_AGENT`, 2 × `MECH_ANALYTICS_URL` | same 3 |

**The two new keys are inert.** The path keys they sit beside are byte-identical before and after — `CONNECTION_POLYMARKET_CLIENT_CONFIG_HOST` stays `https://clob.polymarket.com` and `…_CHAIN_ID` stays `137` — so the named form carries exactly the literals that were already resolving. The four subgraph URLs do not appear as added, because `service.yaml` already defined those names.

**`'str:'` reaching production is confirmed, not inferred:** the baseline build emits `MECH_ANALYTICS_URL=str:` for both live services.

**The restored fail-loud guard cannot break an existing deployment.** `USE_MECH_ANALYTICS` is `false` in both built envs, and no service in `~/.operate` declares either variable, so the guard's `flag on + empty URL` condition is unreachable today. This is a distinct question from "did any value change", and worth asking whenever a fix restores a guard a bug had been suppressing.

These results were independently reproduced by the parallel optimus work against the same two Trader services.

## Notes for the reviewer

- **`darglint` fails on Python 3.14** with `'FunctionDescription' object has no attribute 'docstring'` (exit 129). It reproduces identically on an untouched tree and is unrelated to this PR. CI runs lint on 3.10 and does not see it.
- **`tox -p` across `autonomy`-based environments is flaky** — concurrent `autonomy` processes race on the global `~/.aea/cli_config.yaml` and abort with `ValidationError: 'author' is a required property`, about one run in three. **CI is unaffected**: its only `-p` group is the pure linters, and the four checks in the step enabled here run sequentially. The exposure is `make common-checks-1`, which is already dead in this repo for the bare-`tox` reason. Worth a separate ticket; note that fixing the bare-`tox` lines without also dropping `-p` would turn a reliably-dead target into a flaky one.
- The lockfile drops `open-aea-ledger-solana` and friends, which 0.21.28 no longer pulls. Nothing here imports them and `check-dependencies` passes.


Adamantios (approved):

packages/valory/skills/agent_performance_summary_abci/models.py
```diff
@@ -341,7 +341,10 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         # than silently returning zero rows and inflating ROI). Both default
         # to the safe pre-migration behaviour: flag off, subgraph read
         # unchanged. See docs (consumer migration §7 in mech-analytics repo).
-        self.mech_analytics_url: str = self._ensure("mech_analytics_url", kwargs, str)
+        # The env-var template grammar has no empty default, so unset arrives as None.
```
DIvyaNautiyal07, line 344: **Nit** — the comment attributes the switch to `${MECH_ANALYTICS_URL:str:null}` to the template grammar, but `${VAR:str:}` (empty default) parses to an empty string in aea just fine. The actual reason is that `validate_service_override_env_vars` (autonomy 0.21.28) rejects the empty-default form on service overrides, which is exactly the strictness rule this PR's body calls out.

A reader following the current comment would look for a template-grammar limitation that isn't there. Small reword so future maintainers reach for the right invariant:

```suggestion
        # analyse-service's validate_service_override_env_vars rejects the empty-default
        # form ${VAR:str:}, so we use ${VAR:str:null} in the service config and normalise
        # the None back to "" here.
```

DIvyaNautiyal07 (approved):
One inline nit on the comment wording; not blocking.

packages/valory/skills/agent_performance_summary_abci/models.py
OjusWiZard: deferring for speed for now

PR #1032, opened 2026-08-31
fix(polymarket): stop the bet store stalling the agent, and unbreak the squid headers

OjusWiZard:
> [!IMPORTANT]
> **Stacked on #1029** — base is `ojuswirastogi/ope-1907-guard-synced-timestamp`, not `main`. Review only the 4 commits listed here; #1029's own changes are part of the base. **Merge #1029 first**, then this retargets to `main` automatically.
>
> Two things the stack currently inherits from #1029, neither caused by this PR:
> - **#1029 is 2 commits behind `main`** (missing `13d70beef` + merge `ea841f7a3`), which is why it shows `CONFLICTING`. Until it catches up, this branch also lacks the tomte 0.7.2 / analyse-service work — so `analyse-service` reports 5 pre-existing findings here (`mech_analytics_url` empty default plus 4 bare Polymarket subgraph URLs), all of which `13d70beef` already fixed on `main`. Merging `main` into #1029 clears them.
> - Because #1029 predates that commit, CI on this branch runs the older lint config.
>
> **#1029 also unblocks OPE-1914 fix 6.** It guards the `ValueError` on `last_round_transition_timestamp` in `decision_maker_abci/handlers.py` — the exact crash that made un-latching the 425 unsafe (see *Not in this PR* below). With #1029 in, that fix becomes viable as a follow-up.

---

Fixes the Polystrat restart loop, the minutes-long Profile load, and the stale Profile data. Closes [OPE-1914](https://linear.app/valory-xyz/issue/OPE-1914).

> [!NOTE]
> Updated after review round 1. Three claims in the first commit message no longer hold and are corrected below — see **Corrections after review**. The headers fix in particular changed shape, so the approving review was for a different resolution.

## The restart loop and the slow Profile are one bug

`multi_bets.json` had grown to 138,697 bets (256 MB) with nothing that ever removes an entry. `get_bet_idx` linear-scanned that list once per incoming market — 5,289 × 138,697 — so `_process_chunk` blocked the single AEA loop for 61–118 s. That starves the agent's HTTP server, which serves the AgentUI, *and* trips Tendermint's `BLOCKS_STALL_TOLERANCE` of 60. Recovery calls `reset_state()`, which zeroes `_last_round_transition_height`, so `_has_transitioned()` stays false and `/healthcheck` latches on 425 until the middleware kills the agent.

| measured on the live 138,697-bet store | before | after |
| --- | --- | --- |
| lookups for one 5,289-market pass | 10.31 s | — |
| whole `_process_chunk` | — | 0.44 s (10 yields) |
| `read_bets` at every agent start | 14.64 s | 3.14 s |
| store on disk | 256.5 MB | 72.6 MB |
| bets retained | 138,697 | 28,382 |

**Index the store once per pass.** `get_bet_idx` becomes `build_bet_index()` on the shared `BetsManagerBehaviour`. The Omen path in `update_bets.py` had the identical scan against a smaller `n`.

**Prune the store.** `prune_bets(now)` drops bets that are expired, have no amount recorded, and ended more than `expired_bet_retention` (default 30 days) ago.

**Yield while walking a chunk.** `_process_chunk` and the per-market parse loop in `_fetch_markets_from_polymarket` both yield every `YIELD_EVERY_N_MARKETS`.

## Stale Profile data: every squid query was returning HTTP 400

All 17 `headers.Content-Type` leaves in `aea-config.yaml` resolved to the literal string `{"Content-Type":"application/json"}`, so every squid query failed with `POST body missing, invalid Content-Type, or JSON object has no keys.` and the UI fell back to an `agent_performance.json` observed three days stale.

`Content-Type` is not a bash-safe key, so open-aea JSON-encodes the whole `headers` dict into one `..._ARGS_HEADERS` env var, and `restrict_model_args` truncates the leaf's own path to that same name — an anonymous `${str:...}` leaf therefore receives the entire dict. Every other endpoint tolerates the malformed header; only the squid's graphql-yoga rejects it.

**The fix is to delete the override, not to re-template it.** `trader_abci/skill.yaml` already carries the same value as 18 plain literals, so the agent-level override only ever restated it. Removing it leaves **no placeholder anywhere in the header path**, so this bug class cannot recur here whatever open-aea does.

Both `service.yaml` overrides (`${HEADERS:dict:...}`, 13 + 14 leaves) are gone too — measured inert: setting all their path-keyed vars, and `HEADERS` itself, moved nothing.

Verified on a **built agent** via `aea config get vendor.valory.skills.trader_abci.models.<m>.args.headers` — six models, all still `{"Content-Type": "application/json"}`, identical to the pre-change build.

## Corrections after review

1. **The headers fix is no longer a named `${API_CONTENT_TYPE:str:...}` placeholder.** Checking the reviewer's premise surfaced that `generate_agent` (`autonomy/deploy/base.py:740`) builds the agent env only from service-override-derived vars, with no host-env passthrough. `API_CONTENT_TYPE` was declared solely in `aea-config.yaml`, so it could never be set under `autonomy deploy` — a knob that silently no-ops, the same criticism F3 made one layer down. Hence removal.
2. **Retention is a parameter, not a module constant.** The first commit argued for a constant to avoid the config surface that caused the header bug. The counter-argument is better: the deletion is one-way, `store_bets` truncates in place with no backup, and changing a constant costs a release plus the CID cascade plus a `PROPEL_SERVICE_HASH_ID` bump. `expired_bet_retention` keeps the 2592000 default and makes "set it very large" the off switch.
3. **The investment guard does not "protect exactly the rows that matter".** `has_investments` means *amounts recorded now*, not *ever traded* — `update_bets_investments` skips expired bets, so a settled market never has its amounts repopulated. A market this agent **won** (bet 2.04, payout 7.286, settled 2026-05-25) carries `investments: {"Yes": [], "No": []}` and would be pruned. Consequence is cosmetic — `_find_market_entry` returns `None` and the row still renders from the subgraph dict. Measured: pruning moves exactly **one** row per agent onto that fallback, which 42% of rows already use on one of the two live agents.

## A latent bug the review caught

`prune_bets` read `self.synced_time`, which is on `QueryingBehaviour` — **not** on `BetsManagerBehaviour` or `BaseBehaviour`. It worked only because both callers happen to inherit it; `DecisionMakerBaseBehaviour` does not, so sharing the helper as intended would have raised `AttributeError`. It now takes `now` explicitly. The tests could not have caught it — injecting `synced_time` via `setattr` succeeds *precisely because* no such property exists on the class.

## Two review claims that did not survive checking

- **"Single-agent Pearl deployments don't involve Tendermint."** `trader_pearl` is `number_of_agents: 1` *and* configures `tendermint_url` / `reset_tendermint_after`; `BLOCKS_STALL_TOLERANCE` is live and read by the health handler. The comment naming Tendermint as the reason for the yield interval is restored.
- **"The `% (N + 1)` mutant survives both yield tests."** It does not: with a 1000-element chunk it fires at 501 and 1002, and 1002 is past the end, so the existing `== 2` assertion already failed it. The exact-boundary test is kept on its own merits.

## Not in this PR

- **`ApiSpecs.process_response` 4xx logging.** `abstract_round_abci` is `third_party` and gitignored here; belongs in open-autonomy.
- **Un-latching the 425 during Tendermint recovery.** `_reset_to_default_params` clears `_last_round_transition_timestamp` and `_last_round_transition_height` together, so latching `_has_transitioned` drops the request into a `ValueError` twelve lines below the gate, which the broad handler turns into a 500 — worse than the 425. Needs the middleware's health contract changed alongside.
- **The open-aea `restrict_model_args` regression.** Still optional, and now more so: there is no placeholder left in the header path to depend on it.

## Verification

- Built-agent check via `aea config get` (above), plus: with the collapsed `..._ARGS_HEADERS` vars present, resolution corrupts 17/17 before and 0/17 after.
- Live squid: corrupted header → **HTTP 400**, corrected → **HTTP 200** with data.
- Benchmarks ran the patched code against a read-only copy of the live store.
- Every new test confirmed non-vacuous by reintroducing the corresponding bug shape; the duplicate-id and yield-boundary tests were mutation-checked.
- black, isort, flake8, mypy, pylint, copyright, check-hash, check-packages, check-doc-hashes, analyse-service (two consecutive clean runs), check-abci-docstrings, check-abciapp-specs, check-handlers, check-dependencies all pass.
- 2,241 tests pass, 100% coverage on all five changed modules.


packages/valory/skills/market_manager_abci/behaviours/base.py
```diff
@@ -189,6 +194,38 @@ def _normalize_polymarket_collateral(self) -> None:
...
+
+        :return: a mapping from bet id to its index in ``self.bets``.
+        """
+        return {bet.id: index for index, bet in enumerate(self.bets)}
+
+    def prune_bets(self) -> None:
+        """Drop long-settled bets that the agent never took a position in."""
+        # `openingTimestamp` carries the market's `endDate` on Polymarket, so
+        # this window means "the market closed more than a retention period ago".
+        cutoff = self.synced_time - EXPIRED_BET_RETENTION
```
jmoreira-valory: 🔵 **F1 · LOW — prune_bets lives on a base class that has no synced_time**

`prune_bets()` is defined on `BetsManagerBehaviour`, but `self.synced_time` is **not** an attribute of that class, nor of `BaseBehaviour`. It comes from `QueryingBehaviour` (`graph_tooling/requests.py:117`), and the method only works because both current callers happen to be `class X(BetsManagerBehaviour, QueryingBehaviour)`.

That matters because `BetsManagerBehaviour` is also the base of the entire decision-maker tree — `DecisionMakerBaseBehaviour(BetsManagerBehaviour, ABC)` in `decision_maker_abci/behaviours/base.py:114` — with no `QueryingBehaviour` in the MRO. `BlacklistingBehaviour` had to hand-roll its own `synced_time` property (`decision_maker_abci/behaviours/blacklisting.py:47`) for exactly this reason. So calling `prune_bets()` from any decision-maker behaviour — the natural next step, given the stated rationale for putting the helper on the base rather than in either behaviour — raises `AttributeError: 'X' object has no attribute 'synced_time'` at runtime.

The new tests can't catch this: `_ConcreteBetsManager` in `test_behaviours_base.py` subclasses `BetsManagerBehaviour` directly and `_make_behaviour(..., synced_time=NOW)` injects the attribute with `setattr`. That `setattr` succeeds *only because no such property exists on the class* — the tests are green on precisely the class shape that would fail in production. (mypy stays quiet because `BaseBehaviour` comes from the gitignored third-party `abstract_round_abci`, so it degrades to `Any` and every attribute access is allowed.)

Cheapest fix is to make the dependency explicit rather than inherited:

```suggestion
    def prune_bets(self, now: float) -> None:
```

and pass `self.synced_time` from the two call sites. Alternatively hoist a single `synced_time` onto `BetsManagerBehaviour` and delete the two divergent copies — note they disagree on type today (`QueryingBehaviour` returns `int`, `BlacklistingBehaviour` returns `float`).

packages/valory/skills/market_manager_abci/behaviours/base.py
```diff
@@ -50,6 +51,10 @@
 MULTI_BETS_FILENAME = "multi_bets.json"
 READ_MODE = "r"
 WRITE_MODE = "w"
+EXPIRED_BET_RETENTION = 30 * DAY_IN_SECONDS
```
jmoreira-valory: 🔵 **F2 · LOW — 30-day retention is hardcoded, so the one-way deletion has no deploy-time kill switch**

On first run against the reporter's store this drops ~110k of 138,697 entries, and it is irreversible: `store_bets()` opens `multi_bets.json` in `"w"` and truncates in place, with no backup and no `.bak` rotation. If the `has_investments` guard turns out to under-protect on some user's store (an older store written before investments were tracked, a bet whose amounts were zeroed by `update_investments(0)` → `set_investment_amount(vote, 0)`), the only recovery is "there is none".

Because `EXPIRED_BET_RETENTION` is a module constant, widening or disabling it needs a package release plus the full CID cascade and a `PROPEL_SERVICE_HASH_ID` bump in `agent-deployments` for both `trader_pearl` and `polymarket_trader` — days, for a number. Promoting it to a `MarketManagerParams` value with an `${int:...}` leaf in `aea-config.yaml` makes it a one-line env override, and setting it very large becomes the off switch:

```suggestion
EXPIRED_BET_RETENTION = 30 * DAY_IN_SECONDS  # default; prefer a param so it is tunable at deploy time
```

Worth considering even as a follow-up: the first release that ships this is the one where you most want the dial.

packages/valory/skills/market_manager_abci/behaviours/base.py
```diff
@@ -189,6 +194,38 @@ def _normalize_polymarket_collateral(self) -> None:
...
                 bet.collateralToken = ""
 
+    def build_bet_index(self) -> Dict[str, int]:
+        """Map each stored bet's id to its index in ``self.bets``.
+
+        :return: a mapping from bet id to its index in ``self.bets``.
+        """
+        return {bet.id: index for index, bet in enumerate(self.bets)}
+
+    def prune_bets(self) -> None:
```
jmoreira-valory: 🔵 **F5 · LOW — prune_bets is the first mutation that shifts positions, which sampled_bet_index depends on**

The sampled bet is carried between rounds as a **positional index** into the bets list, not by id: `SamplingBehaviour` computes `self.bets.index(sorted_bets[0])` (`sampling.py:238`, `:263`), stores it as `synchronized_data.sampled_bet_index`, and downstream behaviours re-read the file and do `self.bets[bet_index]` (`decision_maker_abci/behaviours/base.py:281`, `blacklisting.py:55`).

Until now nothing ever removed an element from `self.bets` — `_process_chunk` only appends and updates in place, so positions were stable across the whole period. `prune_bets()` is the first operation that shifts every position after the first dropped entry.

I traced this and it is **safe today**: `sampled_bet_index` is not in `cross_period_persisted_keys`, and prune runs in `setup()` of the period's first market-manager behaviour, strictly before `SamplingBehaviour` recomputes it. But that ordering is an unwritten contract, and the failure mode if it is ever broken (a second `prune_bets()` call site, a mid-period re-entry into `UpdateBetsRound`) is that the agent bets on or blacklists a *different market than the one it evaluated* — silent, and financial. The inline `# must follow the blacklist pass above` comment at the call sites is good; please add the other half to the method docstring:

```suggestion
    def prune_bets(self) -> None:
        """Drop long-settled bets that the agent never took a position in.

        Shifts positions in ``self.bets``; MUST run before ``SamplingBehaviour``
        recomputes ``sampled_bet_index``, which is a positional index.
        """
```

packages/valory/skills/market_manager_abci/behaviours/base.py
```diff
@@ -189,6 +194,38 @@ def _normalize_polymarket_collateral(self) -> None:
             if getattr(bet, "collateralToken", ""):
                 bet.collateralToken = ""
 
+    def build_bet_index(self) -> Dict[str, int]:
```
jmoreira-valory, line 196: ⚪ **F6 · NIT — build_bet_index resolves duplicate ids to the last entry, get_bet_idx resolved to the first**

`get_bet_idx` used `next((i for i, bet in enumerate(self.bets) if bet.id == bet_id), None)` — first match. The dict comprehension keeps the **last** entry for a repeated id. I could not find a path that appends a duplicate into the store (`_process_chunk` guards on the index on both paths), so this looks inert, but it is a silent behavioural delta in a helper that is now shared by two market paths and reachable from the whole decision-maker tree. One line in the docstring noting last-wins on duplicate ids would keep the next reader from assuming first-wins.

jmoreira-valory (dismissed):

**What this does.** Two unrelated production fixes for the Polymarket trader. First, the bet store (multi_bets.json) had grown to ~139k entries with nothing ever removing them, and the per-market lookup was a linear scan over that list, so a single market-fetch pass blocked the AEA loop long enough to trip Tendermint's stall tolerance and wedge the agent in a restart loop; this replaces the scan with a one-pass dict index, adds a 30-day prune for settled bets the agent never took a position in, and makes the chunk walk yield back to the agent loop periodically. Second, every `Content-Type` header leaf in the agent's `aea-config.yaml` was an anonymous `${str:...}` placeholder whose path-derived env var name collides with the collapsed `..._ARGS_HEADERS` var a deployment exports, so all 17 headers resolved to the whole JSON dict as a string and every squid query came back HTTP 400; naming the placeholder `API_CONTENT_TYPE` stops that fallback ever being consulted.

Unusually well-evidenced PR. I verified the load-bearing claims against HEAD rather than taking the body's word:

- **Both `get_bet_idx` call sites were migrated** and there is no orphan `self._process_chunk(...)` left un-`yield from`-ed — the classic generator-conversion footgun where a chunk is silently never processed. Both callers (`update_bets.py` `_update_bets`, `polymarket_fetch_market.py` `_update_bets`) use `yield from`, and both are themselves driven by `yield from` in `async_act`.
- **`bet_index[bet.id] = len(self.bets)` before `self.bets.append(bet)` preserves the old within-chunk duplicate semantics** — a second occurrence of the same id in one chunk still hits the update path, as it did when the scan walked the growing list.
- **A bare `yield` is safe here.** `AsyncBehaviour.act()` only enters `WAITING_MESSAGE` when `wait_for_message` sets it explicitly; from `RUNNING` it does `send(None)` on the next tick, so an unadorned `yield` costs exactly one agent-loop iteration.
- **The prune predicate is unit-correct.** Polymarket's `openingTimestamp` is `int(isoparse(endDate).timestamp())` (seconds), and `synced_time` is seconds, so the 30-day window means what the comment says.
- **Consumers of a dropped bet degrade rather than break.** `predictions_helper._find_market_entry` returning `None` falls back to the subgraph's market dict and re-fetches `prediction_response` from the mech, so a false-negative on `has_investments` costs Profile detail and a round-trip, not a rendered row. The `has_investments` guard itself is well chosen — `_replace_with_existing_investments_if_empty` on both the Omen and Polymarket refresh paths keeps the field from being transiently emptied by an API failure, and `_calculate_outcome_investment` sums BUY trades only, so a fully-sold position still reports a position.
- **Nothing re-adds a pruned Omen bet:** `_fetch_bets` filters on `opening_threshold = synced_time + opening_margin`, so long-settled markets are never returned.
- **The hash cascade is complete.** `market_manager_abci` → `{decision_maker_abci, trader_abci}`, `decision_maker_abci` → `{trader_abci, tx_settlement_multiplexer_abci}`, then agent + both services. No dependent was missed; `agent_performance_summary_abci`, `chatui_abci`, `staking_abci` and `check_stop_trading_abci` genuinely do not depend on either.
- No FSM surface is touched (no `Event` / `rounds.py` / `fsm_specification.yaml` / composition change), so the 4-place sync and unmapped-final-state checks are not in play.
- CI is 28/28 green at `35ec974`.

The `Bet.has_investments` / `invested_amount`-raises-after-`blacklist_forever` trap is a good catch, and the test that asserts `invested_amount` raises while `has_investments` still answers is exactly the right shape. Everything below is non-blocking.

**Findings** — 6 · 🔵 5 Low · ⚪ 1 Nit

| F1 | 🔵 LOW | [packages/valory/skills/market_manager_abci/behaviours/base.py:208](https://github.com/valory-xyz/trader/blob/35ec974699c72a922cb3ac4194bf953afb9daf46/packages/valory/skills/market_manager_abci/behaviours/base.py#L208) | **prune_bets lives on a base class that has no synced_time** |
| F2 | 🔵 LOW | [packages/valory/skills/market_manager_abci/behaviours/base.py:54](https://github.com/valory-xyz/trader/blob/35ec974699c72a922cb3ac4194bf953afb9daf46/packages/valory/skills/market_manager_abci/behaviours/base.py#L54) | **30-day retention is hardcoded, so the one-way deletion has no deploy-time kill switch** |
| F3 | 🔵 LOW | `packages/valory/services/trader_pearl/service.yaml:205` | **service.yaml still exports the colliding HEADERS var, which is now inert** |
| F4 | 🔵 LOW | `packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py:504` | **The yield safeguard covers _process_chunk but not the other full-market pass** |
| F5 | 🔵 LOW | [packages/valory/skills/market_manager_abci/behaviours/base.py:204](https://github.com/valory-xyz/trader/blob/35ec974699c72a922cb3ac4194bf953afb9daf46/packages/valory/skills/market_manager_abci/behaviours/base.py#L204) | **prune_bets is the first mutation that shifts positions, which sampled_bet_index depends on** |
| F6 | ⚪ NIT | [packages/valory/skills/market_manager_abci/behaviours/base.py:197](https://github.com/valory-xyz/trader/blob/35ec974699c72a922cb3ac4194bf953afb9daf46/packages/valory/skills/market_manager_abci/behaviours/base.py#L197) | **build_bet_index resolves duplicate ids to the last entry, get_bet_idx resolved to the first** |

---
**Additional notes** (not anchored to a diff line):

- 🔵 **F3 · LOW — service.yaml still exports the colliding HEADERS var, which is now inert** Both service files still override `models.*.args.headers` with a whole-dict placeholder — `${HEADERS:dict:{"Content-Type":"application/json"}}`, 13 leaves in `trader_pearl` and 14 in `polymarket_trader`. That override reaches the agent only via the generated `SKILL_TRADER_ABCI_MODELS_<MODEL>_ARGS_HEADERS` env var, and after this PR nothing in `aea-config.yaml` consults that name any more (that was the whole point of naming the leaf). Two consequences:

1. The operator-facing `HEADERS` knob now silently no-ops. Someone setting `HEADERS` in a deployment gets no error and no effect — the worst shape of config failure, and the same class of failure this PR is fixing.
2. The var that *caused* the corruption is still being exported. The trap stays armed for any future nested-dict model arg overridden at the service layer whose leaf key is not bash-safe; the fix here is per-leaf, not structural.

Suggest either dropping the `headers` override from both service files (nothing consumes it), or re-pointing it at the new knob so the three layers agree on which var owns the header. Whichever you pick, the `> [!IMPORTANT]` note from the PR body about never declaring `API_CONTENT_TYPE` without a non-empty default belongs as a comment next to the placeholders in `aea-config.yaml` — a PR body is not where the next person edits this.
- 🔵 **F4 · LOW — The yield safeguard covers _process_chunk but not the other full-market pass** The stated goal is that "a slow pass cannot stall Tendermint whatever the store size", but `_process_chunk` is not the only O(markets) synchronous stretch in this behaviour. `_fetch_markets_from_polymarket` walks the same ~5,289 markets in `_validate_markets_by_category`, `_deduplicate_markets`, and this per-market loop, which does three `json.loads`, a `dateutil.isoparse`, price/liquidity parsing and outcome-token construction per market — all without giving the loop back. `read_bets()` in `setup()` is another 3.14 s of uninterruptible work by the PR's own measurements, and `setup()` cannot yield at all.

So the index fix is what actually removes the stall; the yield is defence in depth for one of at least three passes. Worth either extending the same `YIELD_EVERY_N_MARKETS` treatment to the parse loop (it is already inside a generator, so it is a mechanical change), or trimming the claim so a future reader does not assume this behaviour is now yield-safe end to end.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
```diff
@@ -2682,8 +2710,51 @@ def test_fetch_markets_no_categories(self) -> None:
...
-        """Test get_bet_idx returns 0 for first element."""
+    def test_build_bet_index_first_element(self) -> None:
+        """Test build_bet_index maps the first element to 0."""
         bet = _make_bet(id="b1")
         behaviour = _make_behaviour(bets=[bet])
-        assert behaviour.get_bet_idx("b1") == 0
+        assert behaviour.build_bet_index()["b1"] == 0
+
+
+class TestProcessChunkYields:
```
DIvyaNautiyal07, line 2750: **P1 — missing test for duplicate ID within a single chunk.**

`bet_index[bet.id] = len(self.bets)` at `polymarket_fetch_market.py:415` and `update_bets.py:185` is load-bearing: it prevents a second occurrence of the same id in the same chunk from double-appending. `_deduplicate_markets` guards raw markets at a different layer and would not catch this.

Removing that single line silently double-appends and no test in either `TestProcessChunk` or `TestProcessChunkYields` (here or in `test_behaviours_update_bets.py`) would fail.

Suggested test: build a chunk with two entries sharing the same `id`, call `list(behaviour._process_chunk(chunk))`, assert `len(behaviour.bets) == 1` and that the second entry's fields overwrote the first.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
```diff
@@ -2682,8 +2710,51 @@ def test_fetch_markets_no_categories(self) -> None:
...
+            fee=0,
+            openingTimestamp=9999999999,
+            outcomeSlotCount=2,
+            outcomeTokenAmounts=[100, 200],
+            outcomeTokenMarginalPrices=[0.5, 0.5],
+            outcomes=["Yes", "No"],
+            scaledLiquidityMeasure=10.0,
+        )
+
+    def test_yields_once_per_interval(self) -> None:
```
DIvyaNautiyal07, line 2753: **P2 — yield boundary not exercised.**

`test_yields_once_per_interval` uses `2 * YIELD_EVERY_N_MARKETS` (1000, expects 2) and `test_does_not_yield_below_the_interval` uses `YIELD_EVERY_N_MARKETS - 1` (499, expects 0). The mutation `if processed % (YIELD_EVERY_N_MARKETS + 1) == 0` survives both — it yields at 501, 1002, which is 1 and 2 respectively in the two ranges tested, passing the exact-count assertions for the wrong reason.

Add a case with exactly `YIELD_EVERY_N_MARKETS` items → assert exactly 1 yield. Same in `test_behaviours_update_bets.py` `TestProcessChunkYields`.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -242,6 +242,13 @@ def invested_amount(self) -> int:
         """Get the amount invested in bets."""
         return self.invested_amount_yes + self.invested_amount_no
 
+    @property
+    def has_investments(self) -> bool:
```
DIvyaNautiyal07, line 246: **P2 — `has_investments` returns False on legacy stores where `investments={}`, which `prune_bets` now interprets as 'safe to drop.'**

`__post_init__` seeds `investments` with the two outcome keys on newly constructed Bets, but a Bet loaded from an on-disk multi_bets.json predating that field would have `investments={}`. `any(sum(amounts) > 0 for amounts in {}.values())` returns False → `prune_bets` classifies it as no-position and drops it once `openingTimestamp < cutoff`.

Before this PR that was a display glitch. With `prune_bets` writing back to disk it becomes permanent loss of trade history that `agent_performance_summary` depends on. Consider a defensive default in `BetsDecoder` (fill missing outcome keys with `[]`), or make `has_investments` treat an empty dict as 'unknown, keep' rather than 'confirmed no position.'

---

**Nit (same line) — body can drop `sum > 0` since amounts are non-negative wei:**
```suggestion
    def has_investments(self) -> bool:
        """Whether any investment is recorded, readable on blacklisted bets."""
        # Reads the `investments` dict directly; `invested_amount` reads the `yes`/`no`
        # properties which raise once `blacklist_forever` has set `outcomes = None`.
        return any(a for amounts in self.investments.values() for a in amounts)
```
This also fixes the comment: the crash isn't from a name-resolution step, it's from `yes`/`no` dispatching to `_get_binary_outcome` → `get_outcome`.

packages/valory/skills/market_manager_abci/behaviours/base.py
```diff
@@ -50,6 +51,10 @@
 MULTI_BETS_FILENAME = "multi_bets.json"
 READ_MODE = "r"
 WRITE_MODE = "w"
+EXPIRED_BET_RETENTION = 30 * DAY_IN_SECONDS
+# Bounds how long a chunk pass can hold the AEA loop, which the agent's
```
DIvyaNautiyal07: **Nit — comment couples to an external framework detail that will rot.**

'Tendermint's stall tolerance' is a specific threshold from a specific consensus stack. Pearl deployments are sovereign single-agent, so the stall gate is Pearl middleware's healthcheck, not Tendermint. The durable WHY here is just that the AEA loop is cooperative and long chunks starve the HTTP server.

```suggestion
EXPIRED_BET_RETENTION = 30 * DAY_IN_SECONDS
# The AEA loop is cooperative; yielding while walking a chunk keeps the HTTP server
# responsive so healthchecks don't time out on large stores.
YIELD_EVERY_N_MARKETS = 500
```

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -386,26 +385,34 @@ def _is_null_or_mismatch_violation(raw_bet: Dict[str, Any]) -> bool:
                 return True
         return False
 
-    def _process_chunk(self, chunk: Optional[List[Dict[str, Any]]]) -> None:
-        """Process a chunk of bets."""
+    def _process_chunk(self, chunk: Optional[List[Dict[str, Any]]]) -> Generator:
```
DIvyaNautiyal07: **Nit — return type is under-parameterized.**

This file already uses `Generator[None, None, Optional[List]]` and `Generator[None, None, Optional[List[Dict[str, Any]]]]` where the shape is knowable. `_process_chunk` yields None, sends nothing, returns nothing.

```suggestion
    def _process_chunk(self, chunk: Optional[List[Dict[str, Any]]]) -> Generator[None, None, None]:
```
Same nit applies to `update_bets.py:169`.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -386,26 +385,34 @@ def _is_null_or_mismatch_violation(raw_bet: Dict[str, Any]) -> bool:
...
-            index = self.get_bet_idx(bet.id)
+            index = bet_index.get(bet.id)
             if index is None:
                 new_count += 1
                 if bet.queue_status.is_expired():
                     if self._is_null_or_mismatch_violation(raw_bet):
                         null_or_mismatch_drops += 1
                     else:
                         zero_liquidity_drops += 1
+                bet_index[bet.id] = len(self.bets)
```
DIvyaNautiyal07, line 418: **Nit — the update branch never refreshes `bet_index`; worth making the invariant explicit.**

Correct today: existing positions in `self.bets` are stable for the lifetime of a single chunk, so the pre-computed index stays valid for updates. But the asymmetry between the two branches (new-bet writes to the index, update-bet does not) is a future-refactor trap — any change that reorders or removes from `self.bets` inside `update_market_info` would silently serve wrong indices for the rest of the chunk.

One-line comment above this write is enough:
```suggestion
                # existing positions are stable for the lifetime of this chunk; only the
                # new-bet branch needs to keep the index in sync
                bet_index[bet.id] = len(self.bets)
```

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -344,10 +345,8 @@ def setup(self) -> None:
         # helps in resetting the queue number to 0
         if self.bets:
             self._blacklist_expired_bets()
-
-    def get_bet_idx(self, bet_id: str) -> Optional[int]:
-        """Get the index of the bet with the given id, if it exists, otherwise `None`."""
-        return next((i for i, bet in enumerate(self.bets) if bet.id == bet_id), None)
+            # must follow the blacklist pass above, which is what sets `EXPIRED`
```
DIvyaNautiyal07: **Nit — call-site change-narrative comment; delete.**

The ordering invariant is already expressed by `prune_bets` itself checking `bet.queue_status.is_expired()` — a caller that flipped the order would just no-op the prune, not produce a wrong result. This comment describes why the code was written this way rather than an invariant a future reader needs to act on, so it belongs in the commit message. Same at `update_bets.py:166`.

```suggestion
        if self.bets:
            self._blacklist_expired_bets()
            self.prune_bets()
```

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
```diff
@@ -2682,8 +2710,51 @@ def test_fetch_markets_no_categories(self) -> None:
...
         behaviour = _make_behaviour(bets=[bet])
-        assert behaviour.get_bet_idx("b1") == 0
+        assert behaviour.build_bet_index()["b1"] == 0
+
+
+class TestProcessChunkYields:
+    """Tests that _process_chunk hands the AEA loop back while walking a chunk."""
+
+    @staticmethod
+    def _raw(index: int) -> Dict[str, Any]:
```
DIvyaNautiyal07: **Nit — `_raw(index)` is duplicated verbatim in `test_behaviours_update_bets.py:1155`.**

Same 12 fields, same defaults, same purpose (build a synthetic raw-bet payload for chunk yield tests). A shared factory in a test-utility module or conftest keeps the two files from drifting.

OjusWiZard:
@jmoreira-valory — all six findings addressed, but two of them took me somewhere different from what you suggested, and one of your premises turned out to matter more than either of us realised. Pushed as `fe84d565`.

**F3 went further than you asked, because checking your premise found a second dead knob.** You were right that the `${HEADERS:dict:...}` overrides were inert — I verified it rather than taking it: they emit 13 (and 14) path-keyed `..._ARGS_HEADERS` vars, and setting all of them *plus* `HEADERS` itself moves nothing; only the named leaf did. But then `generate_agent` (`autonomy/deploy/base.py:740`) builds the agent env from service-override-derived vars plus four common ones, with **no host-env passthrough** — and `API_CONTENT_TYPE` was declared only in `aea-config.yaml`, never in a `service.yaml`. So the named placeholder could never be set under `autonomy deploy` either. It was the same silent no-op knob you objected to, one layer up.

Since `trader_abci/skill.yaml` already carries the value as 18 plain literals, the agent override only ever restated it. So the headers block is gone from `aea-config.yaml` too. There is now no placeholder anywhere in the header path. Confirmed on a **built agent** with `aea config get vendor.valory.skills.trader_abci.models.<m>.args.headers`: six models, all still `{"Content-Type": "application/json"}`, identical to the pre-change build.

Your suggestion to put the caveat as a comment beside the placeholders can't work, by the way — I probed it: inserted a comment, ran `autonomy packages lock`, and it was gone. Both files are tool-managed and carry zero comments.

**F1 was a real latent bug and my tests were structurally incapable of catching it.** Confirmed exactly as you described. Worth restating the part that made it invisible: `_make_behaviour(..., synced_time=NOW)` succeeds *precisely because* no such property exists on the class, so the suite was green on the one class shape that would fail in production. `prune_bets(now)` now takes it explicitly.

**F2 taken.** Your irreversibility argument beats the one I shipped. `expired_bet_retention`, same 2592000 default, plumbed through `models.py`, three `skill.yaml`s, `aea-config.yaml` and both `service.yaml`s.

While verifying your `has_investments` reasoning I found the guard is weaker than **I** claimed in the first commit message. It means *amounts recorded now*, not *ever traded* — `update_bets_investments` skips expired bets, so a settled market never has its amounts repopulated. A market this agent **won** (bet 2.04, payout 7.286, settled 2026-05-25) carries `investments: {"Yes": [], "No": []}` and gets pruned. Your read that consumers degrade rather than break holds and is what keeps this cosmetic: measured across both live stores, pruning moves exactly one Profile row per agent onto the subgraph fallback — which 42% of rows already use on one of them.

**F4, F5, F6** done: the parse loop yields on the same interval (you were right that it's a second O(markets) stretch, and that `setup`'s `read_bets` is a third that can't yield at all); `prune_bets` documents that it shifts positions and must precede `SamplingBehaviour` recomputing the positional `sampled_bet_index`; `build_bet_index` documents last-wins.

I re-verified your other claims independently rather than trusting them — the hash cascade, `openingTimestamp_gt` keeping settled Omen markets from being re-added, `_calculate_outcome_investment` summing BUY only, both call sites using `yield from`, no FSM surface touched, and CI 28/28. All held.

**Re-requesting review** because the central config change is now a different shape from the one you approved.

OjusWiZard:
@DIvyaNautiyal07 — thanks, the two test findings were the most useful part of this round. Pushed as `fe84d565`. I checked each claim against the code rather than taking it, and two didn't hold up — details below so you can push back if I've got it wrong.

**P1 — taken.** Added `test_duplicate_id_in_one_chunk_updates_instead_of_appending` to both files. Mutation-checked: deleting `bet_index[bet.id] = len(self.bets)` makes it fail with `assert 2 == 1`. One nuance — `_deduplicate_markets` keys by `market_id` across categories, so a duplicate can't actually reach `_process_chunk` from the live path today. The test pins the index-write invariant rather than covering a reachable bug, which I think is still worth having.

**P2 (yield boundary) — test added, but the stated reason doesn't hold.** I ran the `% (N + 1)` mutant against the original pair at `HEAD`: it **failed** `test_yields_once_per_interval` with `assert 1 == 2`. With a 1000-element chunk the mutant fires at 501 and 1002 — but 1002 is past the end, so it's one yield, not two, and the existing `== 2` assertion already caught it. I kept the exact-boundary case on its own merits (an explicit assertion at exactly `N` is the right place to pin an off-by-one) but not on the basis that the pair was inadequate.

**P2 (`has_investments` on legacy stores) — premise is unreachable, but you were onto something real.** `BetsDecoder` routes both branches through `Bet(**data)`, so `__post_init__` seeds the outcome keys and a stored bet with no `investments` key decodes to `{'Yes': [], 'No': []}`, never `{}`. I ran it to be sure. Empirically it's moot too: 0 bets with `investments == {}` across 228k bets on two live stores.

But the underlying worry was right in a different form. `has_investments` means *amounts recorded now*, not *ever traded* — `update_bets_investments` skips expired bets, so a settled market never gets its amounts repopulated. A market this agent **won** (bet 2.04, payout 7.286) carries empty lists and gets pruned. The property's docstring now says exactly that, and retention became a tunable param partly because of it. Your "permanent loss of history that `agent_performance_summary` depends on" was the right instinct aimed at the wrong mechanism.

**Tendermint comment — I applied your reword, then reverted it.** `trader_pearl/service.yaml` is `number_of_agents: 1` *and* configures `tendermint_url`, `tendermint_com_url` and `reset_tendermint_after`. `BLOCKS_STALL_TOLERANCE = 60` is live at `abstract_round_abci/base.py:91` and `block_stall_deadline_expired` is read by the health handler at `decision_maker_abci/handlers.py:388`. Tendermint is exactly the mechanism in this ticket's root cause — the stall trips it, recovery calls `reset_state()`, and that's what latches the 425. Sovereign deployments run Tendermint as framework plumbing; it isn't absent. Your version was accurate but dropped the causal fact that explains why the agent gets *killed* rather than merely slowed, so I restored a precise version naming it.

**Taken as-is:** the `has_investments` body simplification and your corrected comment (you were right that the crash is `yes`/`no` dispatching to `get_outcome`, not name resolution); `Generator[None, None, None]` on both; the index-asymmetry comment; deleting both call-site ordering comments — I verified your reasoning that flipping the order only matches fewer bets and can't drop one it shouldn't, so it was change-narrative, not an invariant; and `_raw` moved to a shared `tests/conftest.py` as `raw_bet`.

The agent-config tests were rebuilt around a more general invariant while I was in there: no anonymous placeholder nested under any non-exempt model arg, which catches the next one anyone adds rather than just headers. Each new test was confirmed to fail on the corresponding reintroduction.

packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py
```diff
@@ -494,7 +504,9 @@ def _fetch_markets_from_polymarket(self) -> Generator[None, None, Optional[List]
             unexpected_error_in_category = 0
             closed_in_category = 0
 
-            for market in markets:
+            for processed, market in enumerate(markets, start=1):
```
jmoreira-valory, line 507: 🔵 **F1 · LOW — Parse-loop yield counter resets per category, so the 500-market bound is per-category not per-pass**

`enumerate(markets, start=1)` at line 507 sits inside the `for category, markets in deduplicated_by_category.items()` loop at line 495, so `processed` restarts at 1 for every category. The yield therefore only fires for a category that on its own holds 500 or more markets.

On the measured workload — ~5,289 markets spread across the category map — a pass where no single category clears 500 hands the loop back **zero** times, and the loop is held for the whole parse. That is exactly the shape the constant's comment says it bounds, so the invariant the code documents is not quite the one it delivers.

Worth noting the asymmetry: the sibling counters in `_process_chunk` here (line 404) and in `update_bets.py` (line 180) are both flat loops over the whole chunk, so they do deliver the documented bound. This is the only one of the three that does not.

LOW rather than blocking because the dominant cost — the 61-118 s `get_bet_idx` scan — is genuinely fixed, and three `json.loads` plus a date parse over ~5k markets is nowhere near `BLOCKS_STALL_TOLERANCE`. But the bound is cheap to make real, and it is what keeps this from regressing as the market count grows.

Hoist the counter above the category loop:

```python
        processed = 0
        for category, markets in deduplicated_by_category.items():
            ...
            for market in markets:
                processed += 1
                if processed % YIELD_EVERY_N_MARKETS == 0:
                    yield
```

A test for it: two categories of `YIELD_EVERY_N_MARKETS // 2 + 1` markets each, asserting one parse-loop yield — which fails on the current code.

packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py
```diff
@@ -1709,6 +1739,34 @@ def test_successful_fetch_with_valid_markets(self) -> None:  # type: ignore[arg-
...
+        try:
+            next(gen)
+            while True:
+                gen.send(None)
+                yields += 1
+        except StopIteration:
+            pass
+
+        # one yield from the connection request, one from the parse loop boundary
+        assert yields >= 1
```
jmoreira-valory, line 1768: ⚪ **F2 · NIT — Yield-count assertion is looser than the value it knows, and the comment above it counts differently**

The test is not vacuous — `next(gen)` consumes the connection-request yield without incrementing `yields`, so removing the parse-loop yield drops the count to 0 and `>= 1` fails. But the expected value here is exactly 1, and `>= 1` leaves a family of mutants alive: `% 1`, `% 250`, or a yield moved inside the try-block all still pass.

That reads as an oversight rather than a choice, given `TestProcessChunkYields` in the same file was deliberately tightened this round — `test_yields_exactly_once_at_the_boundary` was added precisely to pin an exact count at the boundary. This test covers the same construct in the other loop and should get the same treatment.

The inline comment is also counting something different from the assertion — it describes two yields (connection request plus parse loop) while the assertion only ever sees the one the loop counts, which makes the intent harder to reconstruct later.

```suggestion
        # the connection-request yield is consumed by `next` above and not counted
        assert yields == 1
```

jmoreira-valory (dismissed):

**What this does.** Two production fixes for the Polymarket trader. The bet store had grown to ~139k entries and every incoming market did a linear scan over it, so one fetch pass could hold the single-threaded AEA loop for a minute or more, starving the agent's HTTP server and tripping Tendermint's stall tolerance into a restart loop; this replaces the scan with a one-pass dict index, adds a configurable retention window that drops long-settled bets nobody has a position in, and hands the loop back every 500 markets. Separately, every squid GraphQL query was failing with HTTP 400 because an anonymous placeholder under a `headers` dict resolved to the whole dict stringified; the fix deletes the redundant override outright so the skill's plain literal is what ships. Later rounds turned the retention window into a real `expired_bet_retention` param, made `prune_bets` take `now` as an argument instead of reaching for an attribute its class does not own, and added the duplicate-id and yield-boundary tests the second reviewer asked for.

**Re-affirming at the current HEAD.** My round-1 approval was against `35ec9746` and has since been dismissed; this covers `436a32cf`. Two things changed since I last looked, and both resolve open items:

**CI is now fully green — 28/28.** When I drafted the previous round the six Windows/macOS `test` jobs were still running and my approval was conditional on them. They finished green: all 15 `test` jobs (ubuntu-24.04 / macos-15 / windows-2025 x Python 3.10-3.14), `linter_checks`, `lock_check` on all three OSes, `coverage`, `copyright_and_dependencies_check`, CodeQL, Snyk, gitleaks. The condition is discharged.

**The newest commit is a two-line docstring fix plus its CID cascade.** `436a32cf` adds the `:return:` line to `has_investments` that `darglint` DAR201 demanded once the docstring went multi-line — nothing of substance. It did force seven files to move (the `bets.py` fingerprint, the `market_manager_abci` CID, then `decision_maker_abci` / `trader_abci` / `tx_settlement_multiplexer_abci` skill.yaml, `aea-config.yaml`, and both `service.yaml` files); I checked the chain is consistent end to end, and `lock_check` passing on all three OSes is the machine-verified confirmation.

**Everything I raised in round 1 is addressed, and I re-verified it at this HEAD.** `prune_bets` now takes `now: int` explicitly with both call sites passing `self.synced_time` (F1); the retention window is `self.params.expired_bet_retention` with the override chain present in `market_manager_abci` / `decision_maker_abci` / `trader_abci` skill.yaml, `aea-config.yaml`, and both `service.yaml` files (F2); the position-shift ordering and the last-wins-on-duplicate-ids behaviour are stated in docstrings (F5, F6). The `0 < bet.openingTimestamp` guard also correctly exempts rows with no timestamp.

**On @DIvyaNautiyal07's open comment at `bets.py:246` — the mechanism does not hold, and the suggested fix would be a no-op.** The claim is that a legacy store with `investments={}` reaches `has_investments` as an empty dict and gets classified as safe to drop. It cannot: `Bet.__post_init__` seeds both outcome keys, and `BetsDecoder.hook` constructs every on-disk row through `Bet(**data)` in *both* branches — the exact-field-match one and the `"id" in data_attributes` compatibility one — so `{}` never survives load. That makes the proposed defensive default in `BetsDecoder` already the current behaviour, and "treat an empty dict as unknown, keep" a branch that would never fire.

The underlying worry — a market the agent actually traded gets pruned because its amounts were never repopulated — is real, but it is the case the PR body already documents as Correction #3, and the stated consequence does not survive checking. `agent_performance_summary` does read `multi_bets.json`, but only as metadata enrichment: the row and all its financials come from `_fetch_bet_from_subgraph`, and a miss falls through `if not market_info: market_info = bet.get("market", {}) or {}`. `prediction_response` is re-fetched from the mech when absent, `potential_net_profit` is only read for *pending* bets (which a market settled 30+ days ago is not), and `openingTimestamp` only feeds `remaining_seconds`, already 0 for a settled market. Title and id survive on the fallback dict, so the tool lookup and slug fetch still work. So this stays cosmetic, matching the author's measurement of one row per agent. Not a blocker in my read — but it is @DIvyaNautiyal07's thread and their call whether the reply satisfies them.

Their other two points landed: the duplicate-id test exists, `test_yields_exactly_once_at_the_boundary` pins the boundary, and the duplicated `_raw` helper is now a shared `tests/conftest.py`.

The two findings below are unchanged from my last round — both still present at this HEAD, both non-blocking.

**Findings** — 2 · 🔵 1 Low · ⚪ 1 Nit

| F1 | 🔵 LOW | [packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py:507](https://github.com/valory-xyz/trader/blob/436a32cf9f688f3c34dd29335819da862a6d1d62/packages/valory/skills/market_manager_abci/behaviours/polymarket_fetch_market.py#L507) | **Parse-loop yield counter resets per category, so the 500-market bound is per-category not per-pass** |
| F2 | ⚪ NIT | [packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py:1768](https://github.com/valory-xyz/trader/blob/436a32cf9f688f3c34dd29335819da862a6d1d62/packages/valory/skills/market_manager_abci/tests/test_behaviours_polymarket.py#L1768) | **Yield-count assertion is looser than the value it knows, and the comment above it counts differently** |

DIvyaNautiyal07 (dismissed):

iasonrovis (approved):

jmoreira-valory (approved):

**What this does.** Two production fixes for the Polymarket trader. The bet store had grown to ~139k entries with a linear scan per incoming market, blocking the single-threaded AEA loop long enough to starve the agent's HTTP server and trip Tendermint's stall tolerance into a restart loop; this replaces the scan with a per-pass dict index, prunes long-settled bets with no recorded position behind a new `expired_bet_retention` parameter, and yields control back to the loop every 500 markets. Separately, it deletes the `headers.Content-Type` overrides from aea-config and both service.yamls, whose placeholder resolution corrupted the header into a JSON-encoded dict string and made every squid query fail with HTTP 400 -- the skill.yaml literals the agent actually uses are locked in by three new regression tests.

Re-affirming my prior approval at the new HEAD `ee69f05f`. The branch moved only by a rebase onto the updated OPE-1907 base plus three benign commits: a package-hash relock onto that base, a behavior-equivalent one-line restyle of the release.yaml AUTHOR guard, and a darglint docstring for `has_investments`. The `synced_timestamp` ValueError guards appearing in the compare view belong to #1029 (the base), not this PR. Verified at HEAD: CI is fully green (38/38 check runs), the compare between my last-reviewed commit and HEAD contains nothing MEDIUM+, and every round-1 review thread (mine and @DIvyaNautiyal07's) is resolved -- `prune_bets(now)` takes the timestamp explicitly, the retention window is a deploy-time parameter wired through all three config layers, and the duplicate-id and yield-boundary tests landed in both test modules.

Two of my round-2 comments remain open and are explicitly non-blocking: the parse-loop yield counter in `_fetch_markets_from_polymarket` still resets per category (LOW -- the bound is per-category, not per-pass, but each category's parse work is well under the stall tolerance), and `test_parse_loop_yields_on_a_large_category` still asserts `yields >= 1` where the exact count is known (NIT). Both threads are already on the PR; fold them into a follow-up at leisure.

**No findings.**

Issue #450, opened 2025-06-23
Unnecessary storage of on-chain service id in synchronized data


Adamantios:
The `on_chain_service_id` does not need to be stored in the synced data, as it is already in the context of the agent and can be accessed via the `params` attribute, e.g. here:

https://github.com/valory-xyz/trader/blob/801821576cd5c01cee5888e8b4c6a450c8ad9bb8/packages/valory/skills/trader_abci/handlers.py#L183

_Originally posted by @Adamantios in https://github.com/valory-xyz/trader/pull/430#discussion_r2161565725_

PR #1035, opened 2026-09-02
feat: plumb tool-reported researchability through to the strategy layer

jmoreira-valory:
# feat: plumb tool-reported researchability through to the strategy layer

## Summary

The mech's optional `researchability` signal (0-1, "can research even answer
this question?") travels to the betting strategy and nowhere else:

- `_get_decision` coerces the raw response's optional `researchability`
  (absent / boolean / non-numeric / out-of-range all degrade to `None`, and a
  stale value can never leak across markets) onto a behaviour attribute. A
  value that is **present but rejected** logs a warning, so "tool never
  implemented the field" and "implemented and broken" are distinguishable;
  an absent key stays quiet.
- `get_bet_amount` forwards it into the strategy `run(**kwargs)` next to
  `confidence` (already plumbed, unread). Shipped strategies ignore it; a
  future sizing strategy (kelly_shrink -- continuous stake shrinkage toward
  the price, the market-aware study's forward-test candidate) can read it
  with **zero further trader changes**.

Deliberately absent, per the study evidence (mech-predict #450) and review
discussion below:

- **No gate**: hard researchability gates were pre-registered, tested, and
  failed out-of-sample (the discarded bets were still profitable; the tool
  over-reports the field). The "skip hopeless markets" effect already emerges
  from the tool's price-anchored probability meeting the existing min-edge
  rule.
- **No local recording**: `PredictionResponse` and the stored-bets
  serialization are untouched -- every delivery already records the signal
  permanently on IPFS/subgraph, so offline analysis needs nothing from the
  trader. The only thing that cannot be recovered after the fact is the
  signal at bet-sizing time -- which is exactly and only what this PR plumbs.
- No new config parameter, no new decision path, tool-selection policy
  untouched.

## Request side

Unchanged: the trader already attaches `request_context` (market price, close
time, resolution rules, ...) to every mech request via
`sampled_bet.to_request_context()`.

## Tests

- Parametrized `_get_decision` capture matrix (0.35 kept; absent / bool /
  out-of-range / non-numeric -> None; stale-value reset pinned; rejected
  values warn, absent stays quiet).
- Parametrized forwarding test (0.35 and None reach the strategy kwargs).
- Both mutation-verified: dropping the stash line or the forwarding kwarg
  fails a test. 2080 tests green; isort/black/pylint/mypy/flake8/darglint
  green; `autonomy packages lock` clean, third-party CIDs untouched.


jmoreira-valory:
**Review verdict: DO NOT MERGE as a gate — this implements the exact design the market-aware study falsified, and the gate's exit path has a policy side effect.** (Posted as a comment because GitHub blocks request-changes on one's own PR.)

## 1. The hard researchability gate is empirically contraindicated (blocking)

The A/B study that shipped `superforcaster-market-aware` (mech-predict #450; full report in trader-analysis `2026-08-24/market-aware-tool/2026-08-31-refined-study/`) pre-registered and tested exactly this rule — "only bet when the tool says the question is researchable" — and it failed on the only sample with enough bets:

- On 163 bets, the researchability rule KEPT bets earning +38.08% and THREW AWAY bets earning **+19.79%** — the discarded bets were still profitable. Every other candidate threshold was neutral or backwards (evidence_quality >= 0.6 discarded the single best slice, +54.36%).
- The tool over-reports researchability (154/163 of its bets labelled researchable, including 23 of the 32 an independent classifier calls not-researchable) — the field is not calibrated enough to be a binary gate input.
- The shipped recommendation, verbatim: *"keep the existing betting rules exactly as they are; record the four extra outputs without acting on them; and test on live markets before turning any of them into a filter."*

The validated forward candidate is **continuous stake shrinkage inside the strategy** (kelly_shrink: `p' = m + lam*(p - m)`, `lam = lam_min + (1 - lam_min) * researchability`) — low researchability shrinks the edge and the *existing* min-edge gate drops the bet naturally. A binary skill-level gate is both the wrong shape and the wrong layer. Default-0.0 makes this PR inert today, but it institutionalizes a knob whose only activation is the measured-harmful move — a dead-config hazard.

## 2. Gate-fire aliases with tool failure in the selection policy (bug)

The gate exits via `return None` from `_get_decision()`. Downstream, `policy.tool_responded(...)` runs only under `if prediction_response is not None:` — so every gated market **skips the policy update entirely**, exactly like a parse failure. An epistemic "researchable but do-not-trade" becomes indistinguishable from "tool returned garbage" to EGreedyPolicy: enable the gate and the reporting tool's response accounting is suppressed on precisely the markets where it was most candid. If any gate survives (see 1: it should not), it must fire *after* the policy update, on a dedicated code path.

## 3. Nits

- `models.py`: the "no-effect value / startup" comment paragraph is duplicated (copy-paste artifact).
- `isinstance(researchability, (int, float))` accepts `bool` (`True < threshold` comparisons); the mech-side coercion deliberately rejects bools — mirror that.
- The threshold is unvalidated (nothing stops `1.5` = gate everything from a gating tool).

## Suggested resolution

Repurpose the PR to what the study actually calls for: **pass the raw `researchability` (and siblings) through to the strategy layer** (e.g. into the strategy `run(**kwargs)` dict next to `confidence`, which is already plumbed and unread) **and log it** — no gate, no new decision path. That gives kelly_shrink a forward-test path as a customs strategy package with zero skill-level behavior change. Alternatively close, and keep the signal logging-only until live data justifies more.

jmoreira-valory:
Reworked per the review above: the branch is reset onto main and force-pushed as `36330fc63` — gate removed, signal now advisory plumbing to the strategy kwargs (details in the updated PR description). Diff shrank to ~60 substantive lines + CID cascade; no new config. The kelly_shrink sizing strategy can now be forward-tested as a customs package with no further trader change.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -112,17 +113,34 @@ class PredictionResponse:
     p_no: float
     confidence: float
     info_utility: float
+    researchability: Optional[float]
```
OjusWiZard: **Critical — this field addition breaks decoding of every bets store written before the upgrade.**

`BetsDecoder.hook` (`bets.py:551-557`) decides whether a nested dict is a `PredictionResponse` by *exact* sorted key-set equality against `dataclasses.fields(PredictionResponse)`. This line grows that set from 4 names to 5, so a `prediction_response` blob persisted by any prior agent version no longer matches. It then also fails the `Bet` branch and the `"id" in data_attributes` fallback (a nested prediction dict has no `id`) and falls through to the terminal `return data` — decoding as a **plain dict**.

Reproduced against this branch:

```
type of prediction_response: <class 'dict'>
ERROR accessing .vote -> AttributeError 'dict' object has no attribute 'vote'
```

`read_bets()` (`market_manager_abci/behaviours/base.py:171`) reports success and logs nothing — decoding genuinely *succeeds*, and it only catches `(JSONDecodeError, TypeError)`. The crash surfaces later, in `decision_receive.py:396`:

```python
previous_response = deepcopy(bet.prediction_response)   # the stale dict
...
bet.prediction_response = prediction_response           # fresh, typed
rebet_allowed = bet.rebet_allowed(previous_response, ...)
```

which reaches `bets.py:456`:

```python
more_confident = (
    self.prediction_response.win_probability      # fresh — fine
    >= prediction_response.win_probability        # the dict -> AttributeError
)
```

`rebet_allowed` short-circuits only when `n_bets == 0`, so this hits **exactly the markets the agent has already traded**. Nothing catches it and `AsyncBehaviour.act()` does not either, so it stalls FSM processing for that market on every tick, unrecoverable without hand-editing the store. `Bet.update_investments` (`bets.py:397`) is a second reachable site.

Not hypothetical: 5c5be27b2 documents a live Polystrat store of 28,392 bets (pruned from 138,697), read at every agent start and surviving upgrades.

Worth noting the repo already has the right pattern for this and it wasn't followed here. `Bet`-level optional fields are protected by the `"id" in data_attributes` fallback and covered by `test_old_json_without_strategy_vote_deserializes` / `..._poly_tags_...` / `..._description_...` (`tests/test_bets.py:1019/1040/1059`). `PredictionResponse` has neither. That's why 254 tests pass — the existing round-trip tests always encode-then-decode fresh state, where `asdict` emits all 5 keys.

**Fix** (prototyped and verified) — match on mandatory-subset instead of exact equality, consistent with the constructor's existing "ignoring extra keys" contract:

```python
mandatory = {P_YES_FIELD, P_NO_FIELD, CONFIDENCE_FIELD, INFO_UTILITY_FIELD}
prediction_fields = {f.name for f in dataclasses.fields(PredictionResponse)}
if mandatory <= set(data_attributes) <= prediction_fields:
    return PredictionResponse(**data)
```

Verified across all three shapes:

```
old 4-key       -> PredictionResponse  win_prob=0.6  research=None
new 5-key null  -> PredictionResponse  win_prob=0.6  research=None
new 5-key value -> PredictionResponse  win_prob=0.6  research=0.8
```

Please add `test_old_json_without_researchability_deserializes` mirroring the three existing `Bet`-level compat tests.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -112,17 +113,34 @@ class PredictionResponse:
...
 
-        # all the fields are probabilities; run checks on whether the current prediction response is valid or not.
+        # all the mandatory fields are probabilities; run checks on whether
+        # the current prediction response is valid or not.
         probabilities = (
-            getattr(self, field_.name) for field_ in dataclasses.fields(self)
+            value
+            for field_ in dataclasses.fields(self)
+            if (value := getattr(self, field_.name)) is not None
         )
```
OjusWiZard: **This validator iterates *all* dataclass fields, so the next non-numeric field crashes every parse.**

The filter is `is not None`, not a type check. It works today only because every field happens to be numeric. Add e.g. `tool_name: Optional[str]` and `0 <= "foo" <= 1` raises `TypeError` on **every** mech response, not just malformed ones. This PR just added the first non-original field to this class, so that is a foreseeable next step rather than a contrived one.

The generic range check is also already dead work for `researchability` — it is pre-validated to `[0,1]`-or-`None` a few lines above, so the loop does nothing for it and only carries the latent trap.

Suggest replacing the whole block (comment included) with the thing the comment is claiming:

```python
probabilities = (self.p_yes, self.p_no, self.confidence, self.info_utility)
```

Behaviourally identical, drops the walrus, and states "the mandatory fields" directly instead of asking the reader to reconstruct that from `dataclasses.fields` introspection plus prose.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -112,17 +113,34 @@ class PredictionResponse:
...
         """Initialize the mech's prediction ignoring extra keys."""
         self.p_yes = float(kwargs.pop(P_YES_FIELD))
         self.p_no = float(kwargs.pop(P_NO_FIELD))
         self.confidence = float(kwargs.pop(CONFIDENCE_FIELD))
         self.info_utility = float(kwargs.pop(INFO_UTILITY_FIELD))
+        # Optional 0-1 signal emitted by market-aware tools ("can research
+        # inform this question?"). Absent, malformed, boolean, or
+        # out-of-range values all degrade to None so that a response from
+        # ANY tool parses exactly as before -- the signal is advisory
+        # (strategies may read it) and never load-bearing here.
```
OjusWiZard: Five lines of design rationale that the PR description already covers at length — `git blame` -> commit -> PR answers all of it, so per our comment bar this should go.

The one thing here that *would* clear the bar isn't actually said: `not isinstance(raw_researchability, bool)` is genuinely non-obvious (`bool` subclasses `int`). The comment says *that* booleans are excluded, not *why* they need excluding. If anything survives, it is that single line.

packages/valory/skills/market_manager_abci/bets.py
```diff
@@ -112,17 +113,34 @@ class PredictionResponse:
...
+        raw_researchability = kwargs.pop(RESEARCHABILITY_FIELD, None)
+        self.researchability = (
+            float(raw_researchability)
+            if isinstance(raw_researchability, (int, float))
+            and not isinstance(raw_researchability, bool)
+            and 0 <= raw_researchability <= 1
+            else None
+        )
```
OjusWiZard: **A misbehaving tool is indistinguishable from a tool that doesn't emit the field.**

Every other field here raises `ValueError("Invalid prediction response initialization.")` when out of range, so a broken tool is loud. `researchability` alone degrades silently, and the log at `decision_receive.py:499` fires only when the value is non-`None` — so a tool that starts emitting `"high"` or `1.5` produces byte-identical observable behaviour to one that never implemented the field. A tool regression would be absorbed forever with zero operator-visible signal.

Suggest logging a warning at the call site that has logger access when the key is *present but rejected*, distinguishing "not implemented" from "implemented and broken".

Separately, minor: the three chained conditions gating a `float()` inside a ternary are dense. Extracting them makes the bool-before-range ordering legible without needing the comment above:

```python
def _coerce_researchability(raw: Any) -> Optional[float]:
    """Coerce a raw mech-tool value into a valid researchability score, or None."""
    if isinstance(raw, bool) or not isinstance(raw, (int, float)):
        return None
    if not 0 <= raw <= 1:
        return None
    return float(raw)
```

Judgment call — the current code is correct either way.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
```diff
@@ -496,6 +496,13 @@ def _is_profitable(
             if self.benchmarking_mode.enabled
             else self.collateral_token
         )
+        if prediction_response.researchability is not None:
+            # Log-only: the signal is recorded for offline analysis and
+            # handed to the strategy; no decision path branches on it here.
```
OjusWiZard: The two lines below this comment only call `logger.info` — there is no other statement for a reader to wonder about, so "Log-only ... no decision path branches on it here" restates what the code plainly shows. Per our comment bar, delete.

packages/valory/skills/decision_maker_abci/behaviours/base.py
```diff
@@ -604,6 +605,10 @@ def get_bet_amount(  # pylint: disable=too-many-arguments,too-many-locals
                     "bankroll": bankroll,
                     "p_yes": p_yes,
                     "confidence": confidence,
+                    # Optional 0-1 signal from market-aware mech tools; None
+                    # for every other tool. Advisory: shipped strategies
+                    # ignore it, a future strategy may read it for sizing.
```
OjusWiZard: Design rationale that the PR description and commit message already carry. The parameter's own `researchability: Optional[float] = None` signature at line 576 already says it is optional, and "a future strategy may read it for sizing" is exactly the kind of forward-looking note that rots when that strategy lands (or doesn't). Delete.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
```diff
@@ -863,6 +865,36 @@ def mock_get_bet_amount(*_args: Any, **_kwargs: Any) -> Generator:
...
+    @pytest.mark.parametrize("signal", [0.35, None])
+    def test_researchability_is_forwarded_to_the_strategy(self, signal: Any) -> None:
+        """The optional mech signal reaches get_bet_amount as a kwarg.
+
+        Advisory plumbing only: shipped strategies ignore it, but a sizing
+        strategy (e.g. kelly_shrink) must be able to read it, and a tool
+        that does not emit it must forward None unchanged.
+
+        :param signal: the researchability value carried by the response.
+        """
```
OjusWiZard, line 922: Multi-paragraph rationale in a docstring — the design story, the advisory framing, and a named future strategy. The test name and the assertion on `captured` already show what is being verified.

Reduce to the one-line summary, keeping `:param signal:` if darglint wants it:

```python
"""The optional mech signal reaches get_bet_amount as a kwarg.

:param signal: the researchability value carried by the response.
"""
```

OjusWiZard (commented):
Multi-perspective review (correctness, tests, silent failures, type design, comments, simplification).

**One blocking issue**, inline on `bets.py:116`: adding a field to `PredictionResponse` breaks `BetsDecoder.hook`'s exact-key-set type discrimination, so every bets store written before this deploy decodes `prediction_response` as a plain `dict`. It fails silently at read time and crashes later in `rebet_allowed` for any market with an existing position. Reproduced, and a fix is prototyped and verified in that comment.

The rest are non-blocking: a latent `TypeError` in the reworked validator, an unobservable tool-misbehaviour path, and four comments/docstrings that restate what the PR description and history already record.

Verified alongside this: 254 tests pass, `check-hash` clean, and both shipped strategies take `**kwargs` with `.get()` so the extra kwarg is harmless. The parsing matrix itself is correct — bool, NaN, out-of-range, non-numeric and absent all degrade to `None` as designed. The decision path really is unchanged; the problem is purely the decoder that was only ever safe for `Bet`.

packages/valory/skills/market_manager_abci/bets.py
DIvyaNautiyal07: **Blocking** (impact lands in `BetsDecoder.hook` at line 556, outside this diff, but this new field is the trigger) — adding `researchability` to `PredictionResponse` breaks `BetsDecoder.hook`'s exact-equality field-set match for any bet persisted before this PR. This hook is called for every JSON object during `read_bets` (`behaviours/base.py:171`).

Reproduced:

```
prediction_attributes = ['confidence','info_utility','p_no','p_yes','researchability']   # 5 keys (post-PR dataclass)
old data_attributes    = ['confidence','info_utility','p_no','p_yes']                     # 4 keys (pre-PR persisted state)
→ equality fails → hook falls through → `return data` (raw dict)
```

The outer bet then constructs via `Bet(**data)` with `data['prediction_response']` being a raw dict — dataclass doesn't validate types at runtime, so it's accepted. Every downstream attribute read (`self.prediction_response.vote` at bets.py:397 / :462, `.win_probability` at :456) raises `AttributeError` on a legacy bet until a fresh mech response overwrites it via `bet.prediction_response = prediction_response` in `decision_receive.py:399`. Any existing agent with a populated `multi_bets_filepath` from before this PR hits this on the first read after upgrade.

Cheapest fix — relax the discriminator to a mandatory-fields subset check so both pre-PR (4-key) and post-PR (5-key) blobs decode as `PredictionResponse`:

```suggestion
        prediction_attributes = {
            f.name for f in dataclasses.fields(PredictionResponse)
        }
        mandatory_attributes = {
            P_YES_FIELD, P_NO_FIELD, CONFIDENCE_FIELD, INFO_UTILITY_FIELD
        }
        data_attributes = set(data.keys())
        if (
            mandatory_attributes.issubset(data_attributes)
            and data_attributes.issubset(prediction_attributes)
        ):
```

Add a regression test that seeds a pre-PR-shape JSON:

```python
def test_decode_pre_pr_prediction_response_without_researchability(self):
    legacy = '{"p_yes":0.6,"p_no":0.4,"confidence":0.8,"info_utility":0.7}'
    decoded = json.loads(legacy, cls=BetsDecoder)
    assert isinstance(decoded, PredictionResponse)
    assert decoded.researchability is None
```

and a Bet-level round-trip that constructs a bet JSON with the old 4-key `prediction_response`, decodes it, and asserts `isinstance(bet.prediction_response, PredictionResponse)` — the existing decoder tests all use the new 5-key shape so they miss this.

jmoreira-valory:
Slimmed further per discussion: local recording removed (`629a5a0fa`). `PredictionResponse` and stored-bets serialization now revert to main untouched — the delivery layer already records the signal permanently on IPFS/subgraph. What remains is only the piece that cannot be reconstructed later: the signal reaching the strategy kwargs at bet-sizing time, for a future kelly_shrink forward test. No gate, no config, no recording — pure plumbing.

packages/valory/skills/market_manager_abci/bets.py
jmoreira-valory: Excellent catch — this was a real pre-upgrade-store breaker in the version you reviewed. Resolved by scope reduction rather than a patch: in 629a5a0fa the `PredictionResponse` field was removed entirely (per the PR discussion, the delivery layer already records the signal permanently on IPFS/subgraph, so the trader stores nothing). `bets.py` and its serialization are byte-identical to main again, so `BetsDecoder.hook`'s key-set match is untouched. The signal now travels `_get_decision` -> behaviour attribute -> strategy kwargs only.

jmoreira-valory: Same resolution as the sibling thread: the field addition was removed in 629a5a0fa (bets.py reverted to main), so the `BetsDecoder.hook` exact-equality match never sees a changed key set. Nothing persisted changes shape.

packages/valory/skills/market_manager_abci/bets.py
jmoreira-valory: Gone with the field: 629a5a0fa reverted `PredictionResponse` to main, so the walrus filter and the latent all-fields range-check trap this thread describes are no longer part of the diff. (Agreed on the substance — if a non-numeric field ever lands in that class, the explicit four-tuple you suggest is the right shape.)

packages/valory/skills/market_manager_abci/bets.py
jmoreira-valory: The five-line rationale went away with the bets.py revert in 629a5a0fa; the one line you said should survive now does, in the coercion helper's guard in ffb47d5f0: `# bool subclasses int, so True/False would otherwise pass as 1.0/0.0.`

packages/valory/skills/market_manager_abci/bets.py
jmoreira-valory: Fixed in ffb47d5f0, at the call site with logger access as you suggested: a present-but-rejected value now logs `Mech response carried an invalid researchability value ...; ignoring it`, while an absent key stays quiet — "not implemented" and "implemented and broken" are now distinguishable. Asserted across the parametrized capture matrix (bool / out-of-range / non-numeric all warn; absent doesn't) and mutation-verified — removing the warning fails 6 cases. The dense ternary is also gone: coercion lives in a small module helper with early returns.

packages/valory/skills/decision_maker_abci/behaviours/decision_receive.py
jmoreira-valory: The log line (and its comment) were removed entirely in the 629a5a0fa scope reduction — recording was dropped from the PR, so there is nothing left at this site.

packages/valory/skills/decision_maker_abci/behaviours/base.py
jmoreira-valory: Deleted in ffb47d5f0 — the kwargs entry now carries no comment; the `Optional[float] = None` signature and the PR description carry the context.

packages/valory/skills/decision_maker_abci/tests/behaviours/test_decision_receive.py
jmoreira-valory: Trimmed in ffb47d5f0 to exactly your suggested form (one-line summary + the darglint `:param:`).

jmoreira-valory:
All 8 review threads addressed and resolved (details in-thread). Five targeted the pre-slim-down scope and were resolved by the `629a5a0fa` reduction itself — including both blocking BetsDecoder findings (bets.py is byte-identical to main again, so no persisted-store compatibility risk exists). The three still-applicable ones are fixed in `ffb47d5f0`: present-but-rejected researchability values now warn (absent stays quiet, mutation-verified), the bool-subclasses-int note survives in the coercion guard, and the rationale comments/docstrings are trimmed per the comment bar. 2080 tests, lint sextet, lock clean.

Adamantios (approved):

